PR #11, opened 2026-06-07
Add Voxra web app

ashish993:
## Summary
- Add the Voxra Next.js web dashboard under apps/web/voxra
- Document setup, credential handling, side effects, preview behavior, and production checklist
- Add Voxra entries to the root and apps README app tables
- Add a PostCSS override so npm audit reports zero vulnerabilities for the app lockfile

## Verification
- npm run lint
- npm test
- npm audit --omit=dev
- npm run build
- python3 scripts/validate_repository.py

## Notes
- The app can initiate real CALL-E outbound calls after authenticated user action, so the README calls out authorization and side-effect boundaries.
- Build verification was run from apps/web/voxra. Node 23 emitted an experimental type-stripping warning; the build completed successfully.

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,6 +238,7 @@ Runnable demo apps live under [`apps/`](apps/). They are not a CALL-E SDK and do
 | [`apps/typescript/broker-login-client-standalone`](apps/typescript/broker-login-client-standalone/) | TypeScript | CALL-E brokered login client without a shared package dependency. |
 | [`apps/python/oauth-login-client`](apps/python/oauth-login-client/) | Python | CALL-E OAuth login client for MCP Streamable HTTP. |
 | [`apps/typescript/oauth-login-client`](apps/typescript/oauth-login-client/) | TypeScript | CALL-E OAuth login client for MCP Streamable HTTP. |
+| [`apps/web/voxra`](apps/web/voxra/) | TypeScript / Next.js | Web dashboard for CALL-E brokered login, one-off outbound calls, transcripts, analytics summaries, and webhook settings. |
 
 The default e2e tests use a local fake broker/OAuth/MCP server or dry-run paths, so they do not require real CALL-E credentials or browser login. Live verification is opt-in in each app README.
 
--- a/apps/README.md
+++ b/apps/README.md
@@ -16,6 +16,7 @@ Current apps:
 | [`typescript/broker-login-client-standalone`](typescript/broker-login-client-standalone/) | TypeScript | CALL-E brokered login client without a shared package dependency. |
 | [`python/oauth-login-client`](python/oauth-login-client/) | Python | CALL-E OAuth login client for MCP Streamable HTTP. |
 | [`typescript/oauth-login-client`](typescript/oauth-login-client/) | TypeScript | CALL-E OAuth login client for MCP Streamable HTTP. |
+| [`web/voxra`](web/voxra/) | TypeScript / Next.js | Web dashboard for CALL-E brokered login, one-off outbound calls, transcripts, analytics summaries, and webhook settings. |
 
 Suggested grouping:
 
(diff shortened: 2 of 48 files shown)


Ray-56 (changes requested):
Thanks for the contribution. The repo validator and app checks pass under Node 24:

- `npm ci`
- `npm test`
- `npm run lint`
- `npm run build`

Before merge, please address the following items.

1. Add webhook SSRF protection

`app/api/settings/webhook/route.ts` currently only checks that webhook URLs use `https:`, and `lib/webhooks.ts` later fetches the configured URL from the server.

Please add server-side validation so webhook targets are public HTTPS endpoints only:

- reject localhost and loopback addresses
- reject private IP ranges
- reject link-local, reserved, multicast, and cloud metadata addresses
- if the target is a hostname, resolve DNS server-side and reject it if any resolved address is not public
- disable redirects, or re-validate every redirect target before following it
- do not log full webhook URLs, because they may contain path/query tokens

Expected behavior:

- `https://localhost/...` is rejected
- `https://127.0.0.1/...` is rejected
- `https://[::1]/...` is rejected
- `https://169.254.169.254/...` is rejected
- a hostname resolving to a private IP is rejected
- failed webhook logs do not include full paths, query strings, or tokens

2. Make call history and completion webhooks server-authoritative

`app/api/calls/history/route.ts` currently accepts terminal call records from the browser and can trigger `call.completed` / `call.failed` webhooks from that client-submitted data.

Please do not trust browser-submitted terminal records as the source of truth. Suggested direction:

- when `/api/calls/run` succeeds, store a server-side record keyed by `run_id` and user identity/email
- include phone, goal, created time, and current phase in that server-side record
- when `/api/calls/status/[runId]` sees an upstream terminal status, verify that the `run_id` belongs to the current user
- persist the final history record from the verified server-side status result
- trigger webhook delivery only from that verified terminal-status path
- make terminal persistence and webhook delivery idempotent so repeated polling does not create duplicate history entries or duplicate webhooks
- if `/api/calls/history` POST is kept, it must not accept `completed` / `failed` records and must not trigger webhooks

Expected behavior:

- manually POSTing a fake completed record to `/api/calls/history` does not create a completion record and does not trigger a webhook
- unknown or cross-user `run_id` values are rejected
- a webhook is sent only after upstream status confirms a terminal call result
- the same terminal `run_id` cannot send duplicate completion webhooks

3. Remove the Redis-unavailable inline auth-state fallback

`lib/auth.ts` says auth state is stored in Redis so the browser never receives `session_secret`, but when Redis is unavailable it falls back to returning encrypted inline auth state containing the broker session secret. Session storage later requires Redis anyway, so this fallback is inconsistent.

Please make Redis a hard requirement for auth state and session storage:

- remove the inline auth-state fallback
- if Redis is unavailable, `/api/auth/login` should return a clear 503-style response
- keep `/api/ready` and actual login behavior consistent
- update README wording so Redis is clearly required for this app

Expected behavior:

- Redis unavailable means login fails clearly before the broker exchange flow proceeds
- browser responses never contain an auth-state blob carrying broker `session_secret`
- README documents Redis as required

4. Add supported Node version guidance

Node 21 failed Vitest startup locally, while Node 24 passed. Please add either a `package.json` `engines` field or README guidance for supported Node versions.

5. Minor cleanup

`git diff --check` reports an extra blank line at EOF in:

```text
apps/web/voxra/app/globals.css
```

Please remove it.

Once these are updated, I’m happy to re-review.

Ray-56:

Remaining Must Fix only:

- Restrict CALLE_MCP_URL/CALLE_BROKER_BASE_URL and outbound webhook destinations to approved secure origins; never log credential-bearing URLs.
- Validate authorized ASCII E.164 recipients, provide a documented credential-free preview, and mask phone/provider error output.
- Do not accept browser-supplied terminal records as trusted call completion or trigger completion webhooks from them; bind history/results to the authenticated owner and provider evidence.

PR #14, opened 2026-06-15
refactor: refined outbound-call-skill-creator

Derekwang2002:
## Summary

This PR refines `outbound-call-skill-creator` into a more explicit, safer, and more auditable generator for outbound phone-call business skills.

Key changes:

- Added a three-level binding model:
  - `fully-bound`
  - `parameterized-bound` as the default
  - `unbound-generic` as dry-run-only by default
- Clarified execution modes:
  - `dry-run-then-batch-approval`
  - `per-call-approval`
  - `approved-direct-execution`
- Distinguished best-effort creation-time preflight from the mandatory runtime gate before real calls.
- Clarified writeback behavior:
  - writeback policy is chosen at creation time
  - writeback target can be fixed or parameterized depending on binding level
  - sensitive values must not be written or surfaced
- Added creation prompt guidance for skill naming, output target selection, binding level selection, and execution mode selection.
- Added a standardized creation summary contract.
- Added structured runtime contracts for:
  - concrete runtime requests
  - runtime gate reports
  - writeback field mapping
  - direct execution guardrails
- Tightened generated skill validation:
  - requires a selected binding level
  - requires a selected execution mode
  - requires runtime gate requirements
  - rejects `unbound-generic` with `approved-direct-execution`
- Split detailed guidance into focused references:
  - `binding-contract.md`
  - `execution-modes.md`
  - `creation-summary.md`
- Updated README and design docs to reflect the current behavior and example flow.

## Why

Outbound phone-call workflows have real-world side effects, so the generated skills need a stronger contract than a generic source-to-call template.

The main design goal is to balance reuse and automation:

- Fully binding everything improves automation but makes skills too narrow.
- Leaving source/writeback unbound improves flexibility but forces users to reconfirm critical safety details every run.
- `parameterized-bound` gives the best default: schema, consent, dedupe, goal, and writeback policy are fixed, while approved runtime parameters can still vary.

The runtime gate also makes the safety model clearer: creation-time checks are useful but cannot be the final authority. Real calls must only happen after the concrete runtime request verifies source access, required fields, consent, dedupe, writeback/session fallback, and provider readiness.

## Validation

Validated with:

```bash
Result:
Repository validation passed.
The repository validation also exercises the generated-skill checker smoke tests, including the new rejection case for unbound-generic plus approved-direct-execution.
```


Ray-56:
Thanks for tightening the outbound skill creator contract. I think the PR is mergeable overall; these are small follow-up fixes to keep the checker aligned with the new contract:

1. The generated skill contract now lists `preflight and creation summary` as a required `SKILL.md` section, but `skills/outbound-call-skill-creator/scripts/check-generated-skill.mjs` does not currently enforce it. Please add a required marker for that section, update the valid fixture in `scripts/validate_repository.py`, and add a negative test that removing the section fails with something like `Generated skill SKILL.md must include preflight and creation summary`.

2. The checker can currently treat `Maximum execution mode: approved-direct-execution` as the selected execution mode, because `EXECUTION_MODE_RE` matches any `execution mode:` phrase. Please tighten the regex to accept only an explicit selected field, for example line-anchored `Execution mode:` or `Selected execution mode:`, and add a negative test where a generated skill only says `Maximum execution mode: approved-direct-execution` but does not declare a selected execution mode. That should fail with `Generated skill SKILL.md must declare a selected execution mode`.

After those changes, `python3 scripts/validate_repository.py` should still pass.

Derekwang2002:
Addressed in `4e5a0f7`.

  - Added the required `preflight and creation summary` marker to the generated skill checker.
  - Updated the valid fixture and added a negative test for the missing section.
  - Tightened selected execution mode matching to only accept line-anchored `Execution mode:` /
  `Selected execution mode:`.
  - Added a negative test to ensure `Maximum execution mode:` alone does not satisfy selected
  execution mode.

  Validation: `python3 scripts/validate_repository.py` passes.

PR #18, opened 2026-06-18
refactor: prune outbound call skill creator

Derekwang2002:
## Summary
This PR prunes `outbound-call-skill-creator` around the **batch-call automation path** so generated skills are less template-like and more directly executable:

- Removes the _generic/unbound(binding level)_ and _per-call approval_ paths, making `parameterized-bound` the minimum contract and keeping `fully-bound` for fixed production or scheduled workflows. 
- Tightens generated-skill validation so legacy binding or execution modes are rejected instead of quietly accepted.

The result-output model is now more explicit: prefer verified source writeback, use source-adjacent result artifacts when results should stay beside the source without mutating it, fall back to a new local result CSV, and keep session-table output only as a last-resort non-persistent fallback.

## Validation
- `node --check skills/outbound-call-skill-creator/scripts/check-generated-skill.mjs`
- Smoke-tested the generated skill checker with valid `parameterized-bound` and `fully-bound` examples.
- Confirmed the checker rejects legacy `unbound-generic`, `per-call-approval`, and missing result-output behavior cases.

### DIFF
--- a/docs/superpowers/specs/2026-06-12-outbound-call-skill-creator-design.md
+++ b/docs/superpowers/specs/2026-06-12-outbound-call-skill-creator-design.md
@@ -4,19 +4,19 @@
 
 Create an `outbound-call-skill-creator` Agent Skill that helps an agent generate a directly usable outbound phone-call workflow skill.
 
-The generated business skill should behave like `google-form-callback`: after it is created, a user can invoke it with a concrete request such as "process all June 20 submissions" and the generated skill can handle the selected data source, compile outbound call goals, run the approved execution path through the configured MCP provider route, deduplicate work, and write results back or display a tabular session summary.
+The generated business skill should behave like `google-form-callback`: after it is created, a user can invoke it with a concrete request such as "process all June 20 submissions" and the generated skill can handle the selected data source, compile outbound call goals, run the approved execution path through the configured MCP provider route, deduplicate work, and write durable results through source writeback, a source-adjacent result artifact, or a new local result CSV.
 
 `outbound-call-skill-creator` is not a generic outbound runtime platform. It is a skill-generation workflow for creating focused, installable business skills in a selected output target.
 
 ## Scope
 
-The first version supports three built-in source and writeback families:
+The first version supports three built-in source and durable result-output families:
 
 - `google-form`
 - `tiktok-ads`
 - `local-csv`
 
-If the user chooses another source or destination, the creator enters a requirements-gathering workflow and generates a custom adapter contract for the resulting business skill. The custom path should not guess API details, credentials, identifiers, or writeback behavior.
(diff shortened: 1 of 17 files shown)


skills/outbound-call-skill-creator/references/generated-skill-contract.md
```diff
@@ -272,36 +269,44 @@ Generated skills that support `approved-direct-execution` must include this chec
...
+
+Use this result-output mapping shape when durable output or last-resort session-table output is configured:
 
 ```yaml
-writeback:
-  policy: source-writeback | local-csv | session-table
-  target_mode: source-csv-in-place | result-csv-file | source-writeback | session-table | runtime-parameter-name
+result_output:
+  policy: source-writeback | source-adjacent-result-artifact | local-result-csv | source-csv-in-place | session-table-fallback
+  target_mode: source-csv-in-place | result-csv-file | source-writeback | source-adjacent-artifact | session-table | runtime-parameter-name
```
Ray-56: The enum naming here is a little hard to distinguish: `policy` uses `source-adjacent-result-artifact`, while `target_mode` uses `source-adjacent-artifact`, and the checker accepts both. If the shorter target mode is intentionally distinct from the policy value, could we add a short note explaining that distinction? Otherwise I think we should standardize the spelling to avoid generated skills using two apparently equivalent values.

docs/superpowers/specs/2026-06-12-outbound-call-skill-creator-design.md
```diff
@@ -173,25 +172,26 @@ Host scheduler handles recurrence.
...
-Generated skills support three writeback outcomes:
+Generated skills support durable result-output outcomes:
 
-- source writeback, such as Google Sheets or an approved TikTok Ads writeback action
-- local file writeback, such as CSV output
-- session table output when writeback is not configured
+- source writeback to the bound source instance or canonical source record store
+- source-adjacent result artifact in the same provider, account, workspace, folder, or campaign context
+- new local result CSV output
+- session table output only as a last-resort non-persistent fallback
```
Ray-56, line 188: This updated result-output policy looks right, but the Local CSV section above still describes the old fallback model: when writeback is not configured, it says the generated skill outputs a session table. That now conflicts with this policy, which prefers a source-adjacent artifact or new local result CSV and reserves session-table output as a last-resort non-persistent fallback. Since this design doc is linked from the README as developer documentation, can we align the stale Local CSV section with the new result-output model?

Derekwang2002: These lines are correct, but I've already fixed the issue in another part of the file.

skills/outbound-call-skill-creator/references/generated-skill-contract.md
Derekwang2002: Fixed

PR #25, opened 2026-07-08
feat(hubspot-calle): support direct CALL-E workflow calls

Derekwang2002:
## Summary

- Switch the HubSpot CALL-E scaffold to a low-permission direct-call workflow mode that creates CALL-E call tasks without HubSpot CRM writeback.
- Add bundled HubSpot serverless functions, Contact App Cards, a direct-call installer helper, and administrator and user documentation.
- Harden the workflow endpoint with server-side HubSpot request-signature validation, retry-safe CALL-E idempotency, updated safety guidance, secret handling, and links to the official [CALL-E API Reference](https://docs.heycall-e.com/#/api-reference).

This belongs in the repository as a low-code workflow plugin and reference implementation for approval-gated AI-agent phone calls from HubSpot.

## Type

- [ ] New skill
- [ ] New runnable app
- [x] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe
- [x] Validation or tooling update

## Checklist

- [x] Recurring workflows include cancellation behavior (not applicable: the plugin does not create recurring schedules; administrators control HubSpot workflow recurrence and can disable new calls).
- [x] Runnable code has a dry-run, fake-server, or no-call path by default (the installer supports dry runs, tests use mocks, and App Cards require explicit confirmation).

## Validation

- `node --test plugins/hubspot-calle/scripts/test/*.test.mjs`
- `npm test` in `plugins/hubspot-calle/hubspot-project/src/app/functions`
- `npm test` in `plugins/hubspot-calle/hubspot-project/src/app/cards`
- `hs project validate -a 246704939`
- `git diff --check`

## HubSpot Verification

- Deployed successfully in the HubSpot developer test account.
- Verified the static app installation, workflow action, Contact App Cards, serverless secret flow, and low-permission Contact read scope.
- Default automated tests do not place a live phone call.


plugins/hubspot-calle/hubspot-project/src/app/workflow-actions/create-call-candidate-hsmeta.json
```diff
@@ -0,0 +1,215 @@
...
+        "supportedValueTypes": [
+          "STATIC_VALUE"
+        ],
+        "isRequired": true
+      },
+      {
+        "typeDefinition": {
+          "name": "endpoint_token",
+          "type": "string",
+          "fieldType": "text"
```
Ray-56: Please move this credential out of ordinary workflow input data. `endpoint_token` is a normal `STATIC_VALUE` text field, and the public handler authorizes calls solely by comparing it with a server secret. HubSpot custom-action requests provide `X-HubSpot-Signature-v3` and a timestamp through `context.headers`; validate those before any provider call and fail closed. If signature validation is not available in this deployment model, use another server-side-only authentication channel rather than a reusable text input.

plugins/hubspot-calle/hubspot-project/src/app/functions/calle-shared.js
```diff
@@ -0,0 +1,273 @@
...
+    throw ambiguousCallEError("a response-read error");
+  }
+
+  let responseBody;
+  try {
+    responseBody = text ? JSON.parse(text) : {};
+  } catch {
+    throw ambiguousCallEError("a response-parse error");
+  }
+  if (!response.ok) {
```
Ray-56, line 372: Please preserve provider error classification and the current idempotency key for retryable outcomes. This branch turns every parsed non-2xx response into a deterministic error; the Card then returns `retry_same_intent: false`, clears the request ID, and the next confirmation generates a new idempotency key. CALL-E documents `rate_limit_exceeded`, `provider_unavailable`, and `internal_error` as retryable cases, so that can create duplicate calls. Carry the provider status/code from here, keep the same intent for 429/ambiguous/5xx failures, and map deterministic 4xx responses separately. The workflow handler should also avoid converting every provider rejection into a 502, because HubSpot retries 5xx responses.

plugins/hubspot-calle/hubspot-project/src/app/cards/card-utils.mjs
```diff
@@ -0,0 +1,116 @@
...
+    phone_property: String(phoneProperty || "phone").trim() || "phone",
+    request_id: String(requestId || "").trim(),
+    confirmed: true,
+  };
+}
+
+export function buildCardServerlessRequest(input) {
+  return {
+    parameters: buildCardCallParameters(input),
+    propertiesToSend: PHONE_PROPERTY_OPTIONS.map(({ value }) => value),
```
Ray-56, line 37: These properties make the Deal Card non-portable. Both Card manifests advertise `DEAL`, but this request only asks for `phone` and `mobilephone`, which are default Contact properties, not default Deal properties. The setup guidance also does not define a required Deal phone property. In a normal account the Deal Card therefore returns `missing_phone`. Either limit the Card to Contact records, or implement and document a Deal phone source, such as a configured Deal property or an associated Contact lookup.

Ray-56 (commented):
I found three issues that should be addressed before merge. Each inline comment explains the impact and a concrete fix direction.

plugins/hubspot-calle/hubspot-project/src/app/workflow-actions/create-call-candidate-hsmeta.json
Derekwang2002: Removed the visible workflow token. The public endpoint now validates HubSpot v3 request signatures with server-side secrets.

plugins/hubspot-calle/hubspot-project/src/app/functions/calle-shared.js
Derekwang2002: Retryable CALL-E failures now preserve the same request intent and idempotency key. Deterministic errors remain separate.

plugins/hubspot-calle/hubspot-project/src/app/cards/card-utils.mjs
Derekwang2002: Limited both App Cards to Contact records, where phone and mobilephone are portable default properties.

Derekwang2002:
Thanks for the review; all issues fixed

plugins/hubspot-calle/hubspot-project/src/app/cards/call-candidate-card-hsmeta.json
```diff
@@ -0,0 +1,13 @@
+{
+  "uid": "calle_call_candidate_card",
+  "type": "card",
+  "config": {
+    "name": "CALL-E",
+    "description": "Displays low-permission CALL-E direct workflow mode on CRM records.",
+    "entrypoint": "/app/cards/CallCandidateCard.jsx",
+    "location": "crm.record.tab",
+    "objectTypes": [
+      "CONTACT"
```
Ray-56, line 10: The implementation is now Contact-only, but the PR Summary still promises “Contact and Deal App Cards,” and the HubSpot Verification section still says both Deal cards and Deal read scopes were verified. That no longer matches these manifests or the Contact-only scope in app-hsmeta.json. Keeping the cards Contact-only is reasonable because Deals have no portable default phone property; please update the PR description and verification claims accordingly, or restore Deal support with a documented phone-source contract and the required scope.

Derekwang2002: Updated PR #25 description to consistently state:
  - Contact-only App Cards
  - Contact-only read scope
  - Contact-only HubSpot verification

PR #28, opened 2026-07-23
Add consent-first callback window coordinator

estona815:
## What changed

- add a focused Python callback-window coordination app using the public CALL-E server SDK;
- make masked, credential-free preview mode the default;
- require both a recorded recipient callback request and a separate live-execution confirmation;
- return a strict structured result for right-person check, AI-disclosure consent, selected window, and voicemail permission;
- add stable idempotency, phone-like text redaction, no-overwrite private outputs, tests, and operator documentation; and
- list the app in both app indexes.

## Why

Service teams often receive an explicit callback request without a usable time window. The app turns one authorized coordination call into a small, inspectable result without booking, purchasing, changing a service, or collecting sensitive information.

## Safety and operator impact

Preview mode never calls CALL-E. Live mode is one recipient and one call task per stable workflow id. The call discloses that it is AI, checks the intended recipient and ongoing consent, ends on wrong-person or opt-out responses, creates no recurring schedule, masks preview numbers, and omits transcripts from returned output.

## Validation

- `ruff check apps/python/callback-window-coordinator/client.py apps/python/callback-window-coordinator/test_client.py`
- `ruff format --check apps/python/callback-window-coordinator/client.py apps/python/callback-window-coordinator/test_client.py`
- `pytest -q apps/python/callback-window-coordinator` (5 passed)
- masked preview JSON parse check

Default tests use an injected fake client and place no real phone call. Live verification remains explicitly opt-in with an authorized E.164 recipient and a server-side API key.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,7 @@ Runnable demo apps live under [`apps/`](apps/). They are not a CALL-E SDK and do
 
 | App | Language | Purpose |
 | --- | --- | --- |
+| [`apps/python/callback-window-coordinator`](apps/python/callback-window-coordinator/) | Python | Consent-first callback-window coordinator with masked preview, stable idempotency, and structured CALL-E results. |
 | [`apps/python/batch-runner`](apps/python/batch-runner/) | Python | JSONL batch runner using CALL-E CLI auth state, FastMCP, Rich output, and MCP tool-call metadata. |
 | [`apps/python/broker-login-client`](apps/python/broker-login-client/) | Python | CALL-E brokered login client with local token cache and MCP HTTP calls. |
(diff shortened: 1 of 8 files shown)


Ray-56:
Thanks for the contribution. Repository validation and the five default tests pass, but I cannot merge this yet because two privacy guarantees are incomplete:

1. `PHONE_LIKE_PATTERN` only redacts contiguous digits. Common formatted numbers such as `+1 (202) 555-0123` or `202 555 0123` can survive in `structured_result.evidence_summary`, despite the README promising that phone numbers are removed from returned evidence. Please broaden the redaction and add regression cases for formatted numbers.
2. `write_output()` creates and writes the result with the process default permissions, then changes it to `0600`. On a permissive umask the private result is briefly readable by other users. Please create the file atomically with mode `0600` and add a permission assertion.

PR #29, opened 2026-07-23
feat: add crisis-lifeline-bridge skill (verify-before-refer crisis routing)

indicaindependent:
## crisis-lifeline-bridge — verify-before-refer for phone-call agents

**Contribution area:** Agent Skills (`skills/crisis-lifeline-bridge/`)

### What it does
When an AI agent detects that a person is in **acute crisis**, this skill finds a **real local human service** for their specific need and location, then uses CALL-E to place **one verification call** to that agency to confirm the line is real, in service, and can actually help — **before** the number is ever handed to the person.

It turns "here's a hotline, good luck" into "I called them, they're real, they're open, here's who to ask for."

### Why it matters
People in crisis get handed generic numbers constantly, and agent-surfaced numbers are often out of date, disconnected, wrong department, or a national line that can't help locally. A dead or wrong number given to someone in crisis is worse than none — it burns the one moment they were ready to reach out. **No resource is given to the person until a call has confirmed it is real and reachable.**

### Safety by design (matches this repo's emergency/medical/financial boundary rules)
- **Never replaces emergency services.** Imminent danger / active self-harm → direct to 911 + 988 immediately; this skill does not run in that case.
- **Consent-gated.** The verification call goes to the *agency*, never to the person in crisis. The caller represents no one and shares no one's identity.
- **No guessing.** Phone/region/language are never inferred; E.164 only; missing fields surface as plan questions.
- **Honesty rail.** Only names a resource verified *this run* or a known national line — never invents a hotline.
- **Dry-run by default.** Both scripts run with **no real call** and no credentials; live mode is opt-in and refuses to dial fictional sample numbers (`+1 555 01xx`).

### Contents
- `SKILL.md` + `references/{safety,examples,calle-cli-bootstrap,crisis-classification}.md`
- `scripts/find-local-agency.mjs` — candidate research, dry-run samples + real national fallbacks
- `scripts/verify-agency-call.mjs` — CALL-E `plan_call → run_call → get_call_run`, dry-run default, sample-number guard
- README resource-list entry

### Verification
- `python3 scripts/validate_repository.py` → **Repository validation passed.**
- Branch name passes `scripts/check_branch_name.py`.
- Scripts tested in dry-run; `plan_call` exercised live against the CALL-E MCP (planning only, no call placed).

Built for the CALL-E "Your Code Is Calling" hackathon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,7 @@ This project is an awesome list for AI-agent phone-call workflows. Add resources
 - [`call-reminder`](skills/call-reminder/) - Scheduler wrapper skill for recurring CALL-E phone-call reminders.
 - [`google-form-callback`](skills/google-form-callback/) - Google Form response workflow for safe one-off callback calls with dry-runs, scheduling plans, and Sheets writeback. See the [workflow guide](docs/google-form-callback/).
 - [`outbound-call-skill-creator`](skills/outbound-call-skill-creator/) - Creator skill for generating focused outbound phone-call workflow skills from Google Forms, TikTok Ads, Notion, Airtable, local CSV files, or custom sources.
+- [`crisis-lifeline-bridge`](skills/crisis-lifeline-bridge/) - Verify-before-refer safety skill: find a real local crisis service and use CALL-E to phone-verify it is real and in service before referring a person in crisis, with a dry-run no-call default.
 
 ### Apps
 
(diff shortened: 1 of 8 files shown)


Ray-56:
Thanks for the contribution. Repository validation and the documented dry runs pass, but this high-stakes workflow is not safe to merge yet:

1. The trigger says `severity >= 4`, which includes severity 5, while the same skill says severity 5 must never run the verification workflow. Please make the emergency boundary unambiguous and fail closed.
2. The scripts silently default the country/region/language/timezone to US values. Live calls must require or safely omit these values rather than guess them.
3. The national short-code fallbacks are labeled as E.164, but the live verifier rejects them under its own E.164 regex, so the advertised fallback cannot enter the verification flow.
4. Dry-run returns `verified: true` and `recommend_refer: true` for an unverified agency. A machine consumer could treat simulated data as a real referral; simulated results must fail closed and never recommend referral.
5. `calle()` treats a non-zero CLI exit as success whenever stdout is non-empty, and the plan/run responses are then ignored. A structured CLI error can therefore end with `ok: true`. Reject every non-zero exit and validate each step before reporting success.
6. `references/crisis-classification.md` links to nonexistent `safety-crisis-boundaries.md`; the actual file is `safety.md`.

PR #31, opened 2026-07-25
feat(web-oauth-login-client): add User Interface for web oauth login client

daviddprtma:
## Summary
Add user interface for web oauth login client to make user have an ability to check the status that have been connected with calle api integrations and all are corrected to use.

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

## Video Demo 
[Video Demo Web Interface Oauth Login Client](https://www.youtube.com/watch?v=mbyBJNqQU50)


Ray-56:
Thanks for the contribution. The app builds, but I cannot merge it yet:

1. Remote MCP tool names/descriptions, resource URIs/MIME types, and error text are interpolated into `innerHTML`. That enables DOM XSS from server-controlled metadata and is especially serious because OAuth tokens are stored in `sessionStorage`. Please render untrusted values with `textContent`/DOM APIs (or robust escaping) and add a malicious-metadata regression test.
2. The E2E test expects a title matching `/Vite/`, while the page title is `CALL-E OAuth Login`, so the test fails once Chromium runs. The test should verify the actual login/status UI and the core flow, not only body visibility.
3. The UI marks the integration “Connected” before `listTools()` and `listResources()` succeed; failures leave a misleading connected status. Please distinguish transport authentication from usable integrations.
4. The MCP endpoint is hard-coded to one remote service and there is no local fake OAuth/MCP path for the core demo. Please make the endpoint configurable and provide an offline/default verification path consistent with the other OAuth clients in this repository.

`git diff --check` also reports trailing whitespace, and the unused Vite starter files can be removed.

daviddprtma:
> Thanks for the contribution. The app builds, but I cannot merge it yet:
> 
> 1. Remote MCP tool names/descriptions, resource URIs/MIME types, and error text are interpolated into `innerHTML`. That enables DOM XSS from server-controlled metadata and is especially serious because OAuth tokens are stored in `sessionStorage`. Please render untrusted values with `textContent`/DOM APIs (or robust escaping) and add a malicious-metadata regression test.
> 2. The E2E test expects a title matching `/Vite/`, while the page title is `CALL-E OAuth Login`, so the test fails once Chromium runs. The test should verify the actual login/status UI and the core flow, not only body visibility.
> 3. The UI marks the integration “Connected” before `listTools()` and `listResources()` succeed; failures leave a misleading connected status. Please distinguish transport authentication from usable integrations.
> 4. The MCP endpoint is hard-coded to one remote service and there is no local fake OAuth/MCP path for the core demo. Please make the endpoint configurable and provide an offline/default verification path consistent with the other OAuth clients in this repository.
> 
> `git diff --check` also reports trailing whitespace, and the unused Vite starter files can be removed.


@Ray-56 thank you for the feedback. here's the breakdown changes for what it's been updated recently

1 — DOM XSS via innerHTML (Security)
Problem: tool.name, tool.description, content.uri, content.mime_type, err.message and other server-controlled strings are interpolated directly into innerHTML template literals. A malicious MCP server can inject arbitrary HTML/JS and steal the OAuth token from sessionStorage.

Fix:
- Replace every innerHTML template literal that contains untrusted data with document.createElement + textContent assignments.
- Safe static HTML (empty-state cards, SVG icons) may remain as innerHTML.
- Add a sanitizeText(s) helper for the one remaining case (error-card <p>).
- Add regression test e2e/xss.spec.ts that mounts the app with mocked MCP responses containing  payloads in tool names, descriptions, resource URIs, and MIME types, then asserts the script never ran.

2 — E2E test: wrong title assertion + trivial coverage
Problem: Test checks toHaveTitle(/Vite/) but the page <title> is CALL-E OAuth Login. Second test only checks body visibility — no real UI is tested.

Fix: Replace example.spec.ts with tests that:

- Assert toHaveTitle(/CALL-E OAuth Login/).
- Locate #login-btn and verify it is visible + enabled by default.
- Verify #status-text reads "Disconnected" on initial load.
- Verify #error-container is initially hidden.

3 — Premature "Connected" status before listTools / listResources succeed
Problem: onConnected() calls updateStatus(true) before the tool/resource fetches complete. If they fail, the UI still says "Connected".

Fix: Introduce a two-phase status:

- After client.connect(): show "Authenticating…" (amber dot).
- After both listTools() and listResources() succeed (or one partial success): promote to "Connected" (green).
- If both fail: set status to "Error" (red dot) and keep login button visible.
- Add .authenticating CSS class for amber dot (reuses existing dot pattern).

4 — Hard-coded MCP endpoint / no offline path
Problem: SERVER_URL is hard-coded to one private remote service. There is no way to run the demo without live credentials, which violates the repository's app design rules.

Fix:

- Read the endpoint from import.meta.env.VITE_MCP_SERVER_URL with the existing value as the default fallback, so the app still works without any config.
- Create apps/web/oauth-login-client/mock-server/server.ts — a tiny node:http-based fake MCP server that:
- Returns minimal MCP handshake / tools/list / resources/list JSON.
- Has no OAuth, to model the "offline / no credentials" demo path.
- Add "dev:mock" npm script that starts both the fake server and Vite with VITE_MCP_SERVER_URL=http://localhost:3001/mcp.
- Update README to document the two startup modes.

5 — Trailing whitespace + unused Vite starter files
Problem: git diff --check reports trailing whitespace; counter.ts is a leftover from the Vite scaffold and is never imported.

Fix:
- Delete src/counter.ts (unused Vite starter file).
- Strip trailing whitespace from all modified files.

your suggestion is welcome about my PR. thanks.

daviddprtma:
also I have submitted my project for Call-E Hackathon in devpost & you can check it out in here https://devpost.com/software/web-interface-oauth-login-client?ref_content=my-projects-tab&ref_feature=my_projects

Ray-56:
Thanks for the update. The DOM rendering fix, title assertion, and endpoint configurability are good improvements, but this head still has merge blockers:

1. The XSS regression tests can pass without exercising the MCP rendering path. `waitForSelector(...).catch(() => {})` swallows the only evidence that tool/resource data rendered, and the final tool-name assertion expects the resource-URI payload instead of `XSS_TOOL_NAME` (`e2e/xss.spec.ts:125-149`). Please make the waits fail normally and assert each injected value in the element that actually renders it.
2. `npm run dev:mock` runs `vite --mode mock`, but the PR contains no `.env.mock`, so `VITE_MCP_SERVER_URL` is unset and the app falls back to the remote OAuth endpoint. Please add the minimal mock-mode configuration and verify the documented offline flow end to end.
3. If either tools or resources loads, the UI reports the fully healthy `Connected` state (`src/main.ts:345-385`). Please expose a degraded/partial state when one capability request fails, or require both before showing full readiness.

The repository validator and production build pass. `git diff --check` still reports trailing whitespace in `index.html` and `src/style.css`.

Ray-56:
I re-reviewed the current head specifically as an OAuth/login surface. The earlier test, mock-mode, and readiness blockers still apply, and there are two additional authentication blockers:

1. OAuth state is not correlated. `BrowserOAuthClientProvider` implements PKCE persistence but no `state()` generation/storage, while `checkOAuthCallback()` accepts any `?code=` and calls `finishAuth(code)` (`src/oauth-provider.ts:36-83`, `src/main.ts:412-445`). Please generate a cryptographically random state for each login, bind it to the current flow/server, verify and consume it before exchanging the code, and add rejection tests for missing/mismatched/replayed state.
2. OAuth credentials are not partitioned by MCP server. Client information, tokens, verifier, and discovery state all use the same `calle_oauth_*` keys, but `SERVER_URL` is configurable (`src/oauth-provider.ts:8,36-83`, `src/main.ts:15-17`). A same-origin deployment that changes endpoints can reuse or send one server’s bearer token/client state to another server. Namespace stored credentials by the canonical MCP resource/server (and authorization issuer where available), invalidate incompatible entries, and provide an explicit logout/clear-credentials path.

Also remove the callback query parameters even when token exchange fails; the current catch path leaves the authorization code in the address bar. These need login-focused regression tests in addition to repairing the existing XSS suite.

daviddprtma:
hello @Ray-56 thank you for the feedback. here's the fix for what I've changed from your suggestion:

1.) XSS test regression (e2e/xss.spec.ts:125-149)
Fix:
- Remove all .catch(() => {}) from waitForSelector calls; let them fail loudly.
- The tools card renders XSS_TOOL_NAME () in .card-title; resources render XSS_URI (javascript:window.__xss=1). Split into two focused assertions that select the correct container.
- Add data-testid attributes on the tools and resources grids to make selectors precise.

2.) .env.mock is absent → npm run dev:mock falls back to remote endpoint
Fix:
- Ensure .env.mock is committed (not in .gitignore). Current .gitignore likely has .env* but should allow .env.mock.
- Verify the mock server path matches (/mcp endpoint).
- Confirm dev:mock concurrently starts both server and Vite in mock mode.

3.) Partial-load shows "Connected" instead of degraded state (src/main.ts:345-385)
Fix:
- Add a 'degraded' ConnectionStatus variant.
- When exactly one of toolsOk / resourcesOk is true, show 'degraded' with text "Partial — one capability unavailable".
- Only promote to 'connected' when both succeed.
- Add CSS .dot.degraded styling (amber / orange).

4.) OAuth state not correlated (CSRF / replay attack)
Fix:
- Add state(): string to BrowserOAuthClientProvider:
On first call, generate a 128-bit CSRF token via crypto.getRandomValues, base64url-encode it, save it under a server-namespaced storage key, and return it.
On subsequent calls (within same flow) return the stored value.
- Add validateState(state: string): void:
Read the stored state, compare, throw on mismatch or absence.
Clear the state from storage after consuming it (one-time use).
- In checkOAuthCallback():
Read ?state= from the callback URL.
Call provider.validateState(returnedState) before finishAuth(code).
If validation fails: clear the code from the address bar, show an error, abort.
- Add E2E tests for: missing state, mismatched state, replayed state (second use of same state token).

5.) OAuth credentials not partitioned by MCP server
Fix:
- Compute a namespace from the canonical server origin + pathname (e.g. btoa(new URL(serverUrl).origin), truncated).
- Change STORAGE_PREFIX from a module constant to an instance property set from serverUrl in the constructor.
- Storage keys become: calle_oauth_<namespace>_clientInfo, calle_oauth_<namespace>_tokens, etc.
- On construction, detect if stored keys exist for a different namespace (stale server) and purge them.
- Add an explicit clearCredentials() method (used for logout + incompatibility).
- Wire clearCredentials() to a Logout button / link in the UI.
- Update checkOAuthCallback() and connect() in main.ts to pass SERVER_URL to the provider constructor.
- Remove callback query params even in the catch path (currently missing).
- Add login-focused regression tests.

6.) Trailing whitespace (index.html, src/style.css)
Fix: Strip trailing whitespace from both files.

Ray-56:
Thanks for the update. The earlier rendering, mock-mode, readiness, and success-callback state fixes are present, but this head still has current OAuth blockers:

- Credential partitioning is not safe: deriveNamespace uses only the origin and truncates base64 to 32 characters. Same-origin MCP paths share tokens, state, PKCE verifier, client registration, and discovery data, and distinct long origins can collide. Namespace the full canonical MCP server URL with a collision-resistant digest; do not purge unrelated namespaces.
- The provider does not implement the SDK invalidateCredentials(scope) hook, so invalid_grant or invalid_client recovery retries with stale stored credentials. The manual clearCredentials method does not cover this SDK path.
- OAuth error callbacks display the error without validating and consuming state or clearing the pending verifier/state. Bind error responses to the same flow just like success responses.
- The partition tests mostly reimplement storage logic instead of exercising BrowserOAuthClientProvider, so they miss the collision; the logout test also does not assert credential deletion. Add provider-level collision, invalidation, logout, and cross-tab regression coverage.

Also enforce HTTPS for non-loopback MCP endpoints and remove the remaining trailing whitespace reported by git diff --check.

daviddprtma:
> * The partition tests mostly reimplement storage logic instead of exercising BrowserOAuthClientProvider, so they miss the collision; the logout test also does not assert credential deletion

thank you for the feedback. I want to ask about Is it acceptable to expose BrowserOAuthClientProvider globally in the browser environment to facilitate the Playwright E2E tests? This avoids reimplementing the logic in tests.

Ray-56:
The current head only merges upstream `main`; the OAuth implementation is unchanged, so the current blockers remain:

- Storage is still keyed by truncated `URL.origin`, allowing same-origin MCP paths and some long origins to share tokens, state, PKCE verifier, registration, and discovery data. Use a collision-resistant binding to the full canonical resource server plus issuer, and do not purge unrelated namespaces.
- The provider still lacks the SDK `invalidateCredentials(scope)` hook, so invalid token/client recovery can reuse stale credentials.
- OAuth error callbacks do not validate and consume `state` or clear the pending verifier/state, and configured non-loopback MCP endpoints do not require HTTPS.
- Provider-level collision, invalidation, logout-deletion, and error-callback tests are still missing; `git diff --check` also fails in `e2e/xss.spec.ts`.

Please do not expose `BrowserOAuthClientProvider` on `window` in production for testing. Export/import the module in a unit test, or use a test-only build hook that cannot ship in the production bundle. Repository validation/build pass; Playwright could not bind its local server in this sandbox.

daviddprtma:
hi @Ray-56 
Here's a summary of the changes:

1.) Collision-Resistant Storage: Updated deriveNamespace() to asynchronously hash the full canonical serverUrl using crypto.subtle.digest("SHA-256") and base64url-encode the output. This ensures namespaces derived from identical origins but different long paths will not collide. I also removed the logic that purged unrelated namespaces.

2.) Invalidate Credentials: Implemented the invalidateCredentials(scope?) hook in BrowserOAuthClientProvider to clear the storage. This ensures stale credentials are not reused during invalid grant recoveries.

3.) OAuth Error Callbacks & HTTPS Enforcement:
- Updated BrowserOAuthClientProvider.create(...) to strictly enforce that non-loopback server URLs must use https:.
- Refactored checkOAuthCallback to handle OAuth error responses gracefully. It now extracts and validates the state CSRF token even on errors, and clears the active pending states/verifiers correctly before showing the error to the user.

4.) Test Coverage & Validation:
- Completely replaced the e2e/oauth-partition.spec.ts script to exercise BrowserOAuthClientProvider directly using Playwright's page.evaluate. It covers collision, invalidateCredentials(), logout flow, error callbacks, and includes a cross-tab test to assert sessionStorage boundaries.
- Used Vite's import.meta.env.MODE build hook to expose BrowserOAuthClientProvider on the window object only during tests or development environments, guaranteeing it never ships to the production bundle.
- Removed the trailing whitespace on lines 40–41 in e2e/xss.spec.ts resolving the git diff --check failure.

Here's the test is passed too

Screenshot 2026-07-30 094136


any feedback is welcome. thank you.

Ray-56:

- npm run build fails with TS6133 because scope is unused. More importantly, invalidateCredentials ignores the SDK scope and clears all state. MCP SDK 1.29 calls invalidateCredentials("tokens") after invalid_grant and retries the code exchange; deleting clientInfo and the verifier makes that retry fail. Implement the all/client/tokens/verifier/discovery scopes separately.
- deriveNamespace catches any hashing failure and uses the shared namespace default, so different servers can reuse tokens, PKCE verifier, state, and discovery data. Fail closed instead.
- oauth-state.spec.ts still plants the old base64/origin storage key while production hashes the full URL. The mismatch and replay cases can pass through "no pending state" without testing a real stored token.
- The endpoint guard also permits non-HTTP schemes on loopback, for example ftp://localhost.

Repository validation and diff check pass. No credentials or real authentication were used.

Ray-56:
Thanks for the update. The current head is still blocked by these OAuth issues:
- `npm run build` fails with TS6133 because `invalidateCredentials(scope)` ignores `scope`; it also clears every credential class instead of honoring scoped invalidation.
- The OAuth state tests still seed legacy base64(origin) keys instead of the SHA-256 full-URL namespace, so the valid, mismatch, and replay cases do not exercise the production state binding.
- Namespace hashing falls back to a shared `default` bucket, and loopback validation permits non-HTTP(S) schemes. Both should fail closed to preserve token isolation and redirect trust.

daviddprtma:
@Ray-56 thank you for the feedback. I have updated the codes like as what you want. any feedback again is welcome. thank you.

Ray-56:
Current OAuth blockers at `fd54fa5`:

- The MCP server URL is scheme-checked, but `redirectUri` is not. Serving this app from a non-loopback HTTP origin can register a plaintext callback that receives the authorization code and stores tokens.
- Discovery is cached under a resource-only namespace. The provider neither binds credentials to the discovered authorization issuer nor validates issuer equality and HTTPS authorization/token/registration endpoints before redirecting or exchanging the code. An issuer change can therefore reuse stale discovery/client/tokens or send the login/code verifier to an untrusted plaintext endpoint.
- All 19 E2E tests pass, but the supposed successful state callback logs `Existing OAuth client information is required`; it never exercises a successful state+PKCE code exchange. Add a real local success-flow regression plus issuer-change and scoped-invalidation coverage.

daviddprtma:
@Ray-56 thank you for the feedback. here's what's changed for what you've mention and I've updated it: 

1. redirectUri Scheme Validation
- Added assertSecureUrl and isLoopbackHost helpers in oauth-provider.ts to share scheme validation logic.
- Updated BrowserOAuthClientProvider.create() to validate the redirectUri before initializing the provider. If the app is served from a non-loopback plaintext origin (e.g., http://evil.example.com), it will instantly throw and prevent an insecure callback from being registered with the authorization server.

2. Discovery Issuer Binding & HTTPS Validation
- Added a validateDiscoveryState() helper to enforce that all critical AS endpoints (authorization_endpoint, token_endpoint, registration_endpoint) and the authorizationServerUrl use https: (unless on loopback).
- Overrode saveDiscoveryState to inspect the OAuthDiscoveryState and write the canonical issuer to a separate boundIssuer key.
- Bound issuer matching is now enforced on every subsequent save or load. If the discovered issuer ever diverges from the bound issuer, or if any endpoint is downgraded to plaintext HTTP, it immediately throws and clears all credentials to prevent silent reuse of unsafe data.

3. E2E Test Coverage for Security Scenarios
- Created a new test suite at e2e/oauth-security.spec.ts
- Success-flow Regression: Simulates a complete PKCE code exchange by hooking the token exchange endpoint in the MCP stub and pre-seeding the required client info and discovery state. Validates the app reaches connected (or degraded) status and correctly hides the code/state params.
- Issuer-change Invalidation: Specifically exercises the new issuer binding logic to confirm that a changing authorization server clears all credentials and throws.
- Scoped Invalidation Coverage: Explicitly tests all invalidateCredentials(scope) paths (tokens, verifier, discovery, client, all) to ensure only the specified storage keys are wiped.

PR #32, opened 2026-07-25
feat(apps): add MedRoute pharmacy availability workbench

MukoyaKuya:
## Summary
- add MedRoute, a consent-first pharmacy availability workbench under `apps/typescript/medroute`
- use the official `@call-e/calle` SDK to create explicit-recipient calls and collect structured, comparable availability results
- provide a safe, deterministic no-call demo by default plus an opt-in, server-side live-call path
- include result comparison, purchase-quantity totals, transcript PDFs, idempotency, cooldowns, rate limiting, and optional PostgreSQL/OIDC production safeguards

## Safety and side effects
- live calling requires an explicit operator action, two consent acknowledgements, authorized E.164 recipients, and a stable idempotency key
- credentials remain server-side; examples use fictional recipients and the app masks numbers in saved responses
- the workflow does not collect PHI, give medical advice, place orders, reserve medication, or accept payment
- after dispatch, calls are intentionally not auto-retried or cancelled; the recorded provider outcome must be reviewed before a new authorized check

## Validation
- `npm run check`
- `npm run lint`
- `npm test` (133 passing tests)
- `python scripts/validate_repository.py` was attempted. It currently fails identically on unmodified upstream `main` in the generated outbound-skill checker smoke test, before evaluating this app.


Ray-56:
Thanks for the contribution. Repository validation and the documented smoke test pass after installing the Python requirements, but the live path has several merge blockers:

1. The authorization acknowledgement checkbox is enforced only in browser JavaScript and is not sent to the server. A direct `POST /api/check` with `confirmLive: true` bypasses it and can place calls. Enforce both consent gates at the server trust boundary.
2. Live calls have no stable idempotency key. Disabling the UI button does not protect retries, duplicate HTTP requests, or multiple clients from calling the same pharmacy twice.
3. If any live call rejects, its fallback result has no `result`; `score()` then dereferences `item.result.stock_status`, turning a partial failure into HTTP 500 instead of returning the promised ranked partial results.
4. Full phone numbers are returned, stored in history, displayed in result summaries, and written into transcript PDFs. Mask them in summaries and generated artifacts.
5. Every recipient is forced to `locale: "en-KE"` and `region: "KE"` even though the app accepts arbitrary E.164 numbers and does not state a Kenya-only boundary. Require these values or explicitly restrict and validate the supported region.

MukoyaKuya:
Thanks Ray.

Ray-56:
Thanks for the new hardening commits. Server-side consent checks, partial-result handling, and phone masking are improved, but the current head still has live-call and authentication blockers:

1. Production OIDC performs authentication but no authorization: any token with the configured issuer/audience receives access to all history/transcripts and can place live calls, and a token without `sub` is accepted as actor `unknown` (`server.js:88-99,270-315`). Require a non-empty subject and an explicit role/scope for read access and live-call execution.
2. Transcript downloads are normal `<a href>` navigations (`public/app.js:42-44,68-71`), so they cannot attach the required bearer header and will receive 401. Fetch the PDF with the authenticated API helper and download a blob, or use another short-lived authenticated download mechanism.
3. The pharmacy cooldown is check-then-act, not atomic. Two requests with different idempotency keys can both reserve, both observe no recent call, and both start calls before either result writes the cooldown (`server.js:340-358`, `production-store.js:34-47`). Reserve each recipient transactionally before calling.
4. After a live call may have succeeded, any persistence/audit failure releases the idempotency reservation (`server.js:327-362`). Retrying the same key can then place the calls again. Once side effects begin, retain an unknown/pending record and require reconciliation rather than deleting the reservation.

The repository validator and syntax check pass. I could not execute the full smoke suite because the new `pg` package was unavailable in the restricted offline review environment, so the production PostgreSQL/OIDC path still needs runnable coverage.

Ray-56:

Remaining Must Fix only:

- server.js:88-99,270-315 — Require nonempty authenticated subject and actual bounded operator authorization before exposing all transcripts or allowing calls.
- public/app.js:42-44,68-71 — Authenticated transcript downloads use bare anchors and fail with 401; use authenticated fetch/blob.
- server.js:327-362 — Do not release an idempotency reservation after calls may have occurred when persistence/audit fails, which makes the same intent dialable again.

Ray-56:
Superseding review for head 31dd3e7e: the frozen blockers are closed. I applied one bounded maintainer cleanup: Restrict non-production service binding to loopback; production authentication remains unchanged. Focused credential-free checks and clean-source repository validation passed; no real calls were placed. No current Must Fix remains.

PR #34, opened 2026-07-25
feat(apps): add waitlist-backfill web app

LiamLacey95:
## What this is

A cancellation just opened an appointment slot. This app phones the waiting list **one person at a time**, stops at the first person who says yes, and never calls anyone behind them.

Reminder and confirmation calls go out to someone who already has an appointment. Backfill is the inverse: an appointment nobody has yet, and a short window to give it away. It is the phone task with an obvious cost attached — an unfilled slot is revenue that simply does not happen — and it is tedious enough that it often does not get done at all.

Six people on the sample list, two calls placed, one slot filled, and a written reason for each of the four phones that never rang:

| # | Contact | Outcome | Reason |
| --- | --- | --- | --- |
| 01 | Tomas Lindqvist | not called | `timezone_missing` |
| 02 | Grace Abara | not called | `consent_revoked` |
| 03 | Ken Watanabe | not called | `frequency_cap` |
| 04 | Dana Whitfield | called | said no |
| 05 | Marcus Oyelaran | called | said yes → **slot filled** |
| 06 | Priya Raman | never called | `slot_already_filled` |

## Why sequential, and not multi-recipient fan-out

The API can ring a list of recipients together. For this task that is the wrong behaviour, so this app deliberately does not use it.

There is exactly one slot. Fan-out means most people hear about an appointment that is already gone by the time they answer, and if two say yes you have promised the same slot twice and have to ring one of them back to take it away. Sequential calling with a hard stop at the first acceptance is slower in wall-clock terms and correct in every other way.

The suppression is the part worth watching. Priya Raman passes every guardrail and would have been called; she is not, because the slot went two positions earlier. **Calls not placed are the feature**, and the log records why for each one.

## Safety

Preview is the default and places no calls. `CALLE_MODE=simulate` runs the full loop — acceptance, suppression and all — against a fake transport, so the complete behaviour is reviewable without a CALL-E account. Live mode needs a key **and** the operator to type the specific slot id; a general "yes" does not authorise a call. The CLI equivalent is `--execute --confirm-slot <id>`, which exits non-zero rather than guessing.

Guardrails run before each call and emit machine-readable reason codes: E.164, consent scope and revocation, quiet hours in the contact's own IANA timezone, frequency caps, content boundaries for medical/legal/financial/emergency material, and the explicit-intent gate. Numbers are masked to `+1********78` wherever they are logged or displayed, and a test asserts no raw number reaches the audit trail.

Calls carry a deterministic idempotency key, `backfill:<slot>:<contact>:<attempt>`, so re-running a slot that looks hung will not ring the same person twice. There are no recurring jobs and no scheduling: one run is one slot, so nothing can fire unattended tomorrow. `POST /api/cancel` stops the run before the next call, and closing the tab does the same.

### P4, implemented rather than described

Tomas Lindqvist is first on the waiting list and is never called, because nobody recorded his timezone at intake.

Guessing here is easy and wrong. `+1` spans six zones. `EST` is ambiguous between North America and Australia. A stored UTC offset is correct until the next daylight-saving change and silently wrong after it. Any of those guesses puts an automated call into somebody's evening, which is precisely the failure the quiet-hours check exists to prevent — so a guessed timezone is worse than no call at all.

`resolveTimeZone()` accepts an explicit IANA identifier and nothing else: no inference from phone number, country code, locale, language, IP address, abbreviation, or offset. The fix for Tomas is to ask him and record it, which is a data-entry problem rather than a call problem.

## Checks

- `node --test test/` — 19 tests, no credentials, no network, no installed dependencies
- `python3 scripts/validate_repository.py` — passes
- Branch name passes `scripts/check_branch_name.py`
- Every sample number is in the `+1 555 01xx` reserved fictional range

The assertions that matter most are the negative ones: preview places zero calls; a confirmation naming the wrong slot places zero calls; the contact behind the acceptance is never called even though the fixture scripts her to say yes; and a fake transport never announces that real calls will be placed.

## Notes for maintainers

No build step and no framework — `node server.mjs` and open the page. The only dependency is `@call-e/calle`, dynamically imported so the default fake path needs no install. The UI follows the CALL-E dashboard's visual language.

While building this I ran into five things that may be worth your time; they are listed at the end of the app README. The one with a patch attached is a CRLF bug in `scripts/check-generated-skill.mjs` that makes `validate_repository.py` fail on a clean Windows checkout — submitted separately as #33 so this PR stays scoped.

A demo video will follow on this PR.


LiamLacey95:
## Verified against a real call

Ran end to end in live mode against a consented UK number. One real call placed, structured result returned, slot filled, and the person behind the acceptance never dialled.

```
- Tomas Lindqvist (+1********13)     timezone_missing
  No IANA timezone supplied.
- Grace Abara (+1********47)         consent_revoked
- Ken Watanabe (+1********64)        frequency_cap
> Liam (+4*********56)
< Liam (+4*********56)               yes
# Liam (+4*********56)
  Liam accepted. Every remaining person will NOT be called.
. Priya Raman (+1********05)         slot_already_filled
  Slot filled by Liam. This person was never called.

FILLED by Liam | 1 call(s) placed, 4 avoided | 3 blocked by a guardrail, 1 suppressed after the fill
```

`recipientResultSchema` came back as `{"can_take_slot":"yes"}`, which is what drives the stop.

### The quiet-hours check refused first, and that was the right answer

The first live attempt placed **zero** calls:

```
- Liam (+4*********56)               quiet_hours
  Local time 21:07 in Europe/London is outside 09:00-20:00.
```

The shipped sample policy is 09:00–20:00, and it declined to dial a real number at 21:07 local. The run above used a window widened to 23:00 for the account holder's own line, with their consent — an operator widening their own policy, not a bypass of somebody else's protection.

Worth noting the timezone was supplied explicitly as `Europe/London`. It was **not** derived from the `+44` prefix, which is exactly the inference P4 rules out.

### Two things this exercise found

1. **Suppression only protects people behind an *acceptance*.** When the consented line was blocked by quiet hours, the slot never filled, so the run correctly walked on to the next entry — which in a test fixture is a fictional number. Sequential calling means the list keeps moving until someone says yes; only a guardrail block holds regardless of outcome. Obvious in hindsight, and worth stating for anyone adapting this.
2. **The CLI was not printing the CALL-E call id**, so the audit trail ended at our side of the boundary. Fixed in `21e80b4`.

I could not find the API-placed call in the dashboard's **Recents** afterwards; it appears to list assistant-initiated calls only. If that is intended, ignore it — if not, it is a small gap for anyone debugging an API integration.

LiamLacey95:
Two updates.

**Added a screenshot of a completed run to the app README** (`8b4344c`). The rules let judges score on the description and images alone, and there was no picture of the thing working.

**A correction on my own checklist.** I wrote that `scripts/validate_repository.py` passes. That is true on Linux, and it is what your CI will show — but I should be precise, because on Windows this branch alone fails:

```
ERROR: Generated outbound skill checker smoke test failed:
ERROR: Missing YAML frontmatter: ...\generated-callback-skill\SKILL.md
```

Every one of those failures is the generated-skill smoke test. **None of them mentions `apps/web/waitlist-backfill`** — my files validate clean. The failure is the CRLF bug in `check-generated-skill.mjs` that #33 fixes, and I had verified "passes" on a working copy that already had that fix applied, which was not a fair test of this branch on its own.

So, accurately: this branch passes validation everywhere your CI runs, and on Windows it passes once #33 is merged. Sorry for the imprecision.

Ray-56:
Thanks for the contribution. The repository validator and the 19 individual test cases pass, but I cannot merge this yet because the core safety guarantee can still fail:

1. On any `placeCall()` exception the loop continues to the next contact. A timeout or lost response can happen after the provider accepted a still-running call, so this can put two calls in flight and promise one slot twice. Stop on ambiguous transport errors and reconcile the existing call/idempotency key before advancing.
2. Simulation enters request mode `live`, so `server.mjs` uses the real clock instead of `scenario.demoNow`. Outside the sample calling window it skips the scripted calls and does not produce the deterministic acceptance/suppression run promised by the README.
3. `checkConsent()` fails open when `scopes` is missing or not an array; a `grantedAt` value alone becomes callable. Require an explicit scope match.
4. `/api/run` is a side-effecting GET endpoint. A browser prefetch or cross-site request that knows the slot id can start live calls; the typed UI confirmation is not proof of a user gesture at the server boundary. Use a state-changing request with origin/confirmation protection.
5. `/api/scenario` sends raw waitlist phone numbers to the browser, and provider summaries/errors are emitted without phone redaction. This contradicts the claim that the server never sends raw numbers. The same untrusted scenario/provider fields are inserted with `innerHTML`, creating DOM-XSS risk.
6. The documented and package test command, `node --test test/`, fails because Node treats the directory as a module. Use a working command such as the explicit test file/glob and verify `npm test`.

LiamLacey95:
Thanks, this was a genuinely useful review. All six are fixed in `174fab6`. Point by point.

**1. Ambiguous transport errors no longer advance the loop.**
You were right that this was the one that mattered. `classifyTransportError()` in `src/calle.mjs` is fail-closed: only an outright 4xx rejection counts as "the request was never actioned", and that alone is safe to skip past. Timeouts, dropped connections, 5xx, 408, 429 and anything unrecognised are all treated as ambiguous, because for every one of them the provider may hold a call we cannot see. Those halt the run and emit `run_halted` carrying the idempotency key, so the existing call can be reconciled before anything else is dialled. 401/403 halt as well: nothing was placed, but every later call would fail the same way and the operator needs to fix credentials rather than watch a loop retry.

**2. The simulation clock now follows the transport, not the requested mode.**
Exactly as you describe. A simulated run asks for mode `live` on purpose, to exercise the intent gate and the real loop, so `mode === "live" ? new Date() : demoNow` handed it the real clock and quiet hours skipped everyone outside the sample window. It is now keyed on whether the transport is genuinely live, so anything against the fake transport gets `scenario.demoNow`. There is a server test that runs the whole simulated backfill and asserts the acceptance and the suppression, so this cannot silently regress with the time of day.

**3. `checkConsent()` fails closed on scope.**
The old `if (Array.isArray(c.scopes) && !c.scopes.includes(scope))` skipped itself whenever `scopes` was absent or not an array, so a record holding only a `grantedAt` was callable. A missing scope is not a permissive scope, and a malformed record is the case where we know least about what was actually agreed. Absent, non-array and empty `scopes` are now all refusals with distinct reason codes, and a run configured with no scope at all cannot match anything either.

**4. `/api/run` is no longer a side-effecting GET.**
`/api/run` and `/api/cancel` are now POST-only, require a same-origin `Origin` when one is sent, reject a cross-site `Sec-Fetch-Site`, and require `content-type: application/json` so a plain form or an `` cannot reach them without a preflight. You are right that the typed confirmation proves nothing at the server boundary; it is still there as operator intent, but it is no longer doing security work. `EventSource` can only issue GET, so the front end now reads the SSE stream off the POST response body instead, which keeps the live view.

**5. Raw numbers and DOM-XSS.**
`/api/scenario` used to spread the scenario wholesale, which shipped every waitlist number to the browser and made the app's own claim false. It now sends `phoneMasked` and drops `phone` entirely, so masking is done before the value leaves the process rather than in the page. Provider summaries and transport error text go through a new `redactPhones()`, since those are strings we did not compose and they quote the number that was dialled. Separately, all rendering of untrusted values moved off `innerHTML` onto `textContent` and explicit DOM construction, which closes the injection path those same fields opened.

**6. `npm test` works.**
`node --test test/` makes Node try to load the directory itself as a module and fail before running anything. The script now names the test files explicitly, and both README references are updated.

Two things worth your attention rather than buried:

- **One existing test asserted the behaviour finding 1 calls a bug.** "a transport failure does not abort the run or fill the slot" encoded the old continue-on-error path, so it is replaced rather than kept: there are now separate tests for the ambiguous case (halts, places nothing further, reports a reconcile key) and the definitive case (skips that contact, carries on). Please read that as an intentional change, not a dropped test.
- **`scripts/validate_repository.py` currently fails**, on the generated outbound skill checker smoke test (`Missing YAML frontmatter: .../generated-callback-skill/SKILL.md`). I confirmed it fails identically on a clean tree with this branch stashed, so it is not from these changes, but you may want to look at it separately.

Test count is 32, up from 19: the original unit tests plus 7 new server-boundary ones covering the CSRF rejections, the scenario redaction, and the deterministic simulated run. Still no credentials and no network.

Ray-56:
Thanks for addressing the six earlier findings. The scoped-consent, deterministic simulation, POST endpoints, raw-number redaction, and DOM rendering changes are present, but three safety blockers remain:

1. The one-run lock has a race: `activeRun` is checked, then `loadScenario()` is awaited, and only afterward is `activeRun` assigned (`server.mjs:103-124`). Two concurrent POSTs can both pass the check and start separate call loops. Reserve the run synchronously before the first await and add a concurrent-request regression test.
2. The live HTTP surface has CSRF-style headers but no operator authentication. Both `Origin` and `Sec-Fetch-Site` are optional, `/api/scenario` exposes the slot ID, and the server listens on all interfaces (`server.mjs:59-88,162-206`). A direct network client can fetch the ID and POST a live confirmation. Bind the local demo to loopback and/or require operator authentication for live run/cancel endpoints; browser-origin checks alone do not prove user intent.
3. `classifyTransportError()` treats every 4xx except 408/429 as proof that no call exists and advances (`src/calle.mjs:49-70`). A 409 can represent an existing idempotent/in-progress call, and provider-specific 4xx semantics are not established here. Fail closed unless a documented SDK error proves the request was rejected before call creation.

The repository validator passes and all 25 pure unit tests passed locally. The seven HTTP tests could not start their child server inside the restricted review sandbox, so please add the concurrency/auth cases to the runnable suite and keep CI executing the full test command.

LiamLacey95:
All three are fixed in `ffa4bf7`, and the test-runnability point is addressed by restructuring rather than by adding more spawned servers.

**1. The one-run lock was a flag, not a lock.**
Correct, and the window was exactly as you describe. It is now check-and-set inside a single synchronous turn, with no await between reading `activeRun` and reserving it, so there is no point at which the event loop can interleave a second request into the gap. The regression test holds the first run open on a gate that only the test can release, so the second request provably arrives while the first is still in flight, and asserts it is refused with a 409. There is a companion test that the lock is released afterwards, since "never double-runs" is trivially satisfiable by never running twice at all.

**2. Header checks are not authentication, and the demo should not be on every interface.**
Agreed on both, and the framing was the useful part: the headers establish that a request did not come from another website, which is a different question from whether an operator sent it. Anything that can dial now requires a bearer token. It is generated per start when `CALLE_OPERATOR_TOKEN` is unset and printed to the console, so forgetting to configure one produces a token you have to go and read rather than an open endpoint; preview and simulate require none, because neither can reach a telephone. The server binds `127.0.0.1` unless `HOST` is set deliberately, and warns at startup when it is. On the slot id specifically: in live mode `/api/scenario` no longer returns it, since publishing the value a live confirmation has to echo handed over half of the gate to anything that could reach the port. The UI does not need it, because the operator types it and the server is what checks it.

**3. `classifyTransportError()` was asserting something the SDK does not support.**
This one was the clearest of the three, and I could not justify the old behaviour when I went looking. `apiErrorFromResponse` lifts `code` from whatever the response envelope contained and falls back to `internal_error`; nothing documents a mapping from status to whether a call was created. Your 409 example is the sharp one, since an existing idempotent or in-progress call is precisely the case where advancing double-books. So the allowlist is now a single entry: `CalleAuthenticationError` (401/403), which the SDK does define and which means the request was rejected at authentication before anything could be created. That still halts, because every later call fails identically, but it needs no reconciliation. Every other failure, 4xx included, halts and reports the idempotency key. If CALL-E later documents statuses or codes that guarantee pre-creation rejection, the comment says where to add them and asks for the citation.

**On the tests that could not run in your sandbox.**
Rather than making them skip, I moved what they were protecting. The HTTP layer is now `src/http.mjs`, a handler with its dependencies injected, so the concurrency, authentication, disclosure and cross-site cases are exercised by calling it directly with stub request and response objects. They need no port, no socket and no child process, so they run wherever `node --test` runs.

`test/server.test.mjs` is deleted rather than left skipping. Its coverage is now in-process, and the two things only a real listen could show — the loopback bind and the token minting — are pure functions (`resolveHost`, `resolveOperatorToken`) with direct tests. It was also leaking child processes, since `node:test` has no `test.after`, which hung the suite locally and would eventually have done the same in CI.

45 tests, all passing, no credentials, no network, nothing spawned. `npm test` runs the whole thing.

One thing unchanged from last round, flagged again so it does not look ignored: `scripts/validate_repository.py` fails for me on the generated outbound skill checker smoke test (`Missing YAML frontmatter: .../generated-callback-skill/SKILL.md`). It fails identically with this branch stashed, and you report it passing, so I take it to be environment-specific at my end rather than something this branch introduces. Happy to look if it turns out to be real.

Ray-56:
The current head still has five safety/behavior blockers:

1. `runBackfill()` advances whenever `structuredResult.can_take_slot` is missing or unexpected. A completed call with an unparseable result may still have accepted the slot, so halt and reconcile unless the result definitively establishes non-acceptance.
2. `classifyTransportError()` treats every 401/403 from the whole `createAndWait()` operation as proof that no call exists. Authentication can also fail during post-creation polling; retain the idempotency key and treat it as ambiguous unless the create phase itself is known to have rejected the request.
3. Tab-close cancellation is not reliably implemented: the handler attaches `req.close` after consuming the request body, while an SSE client disconnect must be observed on the response/socket. Either implement and test that path or remove the README claim.
4. `loadScenario()` runs before the releasing `try/finally`; a read/parse failure leaks `activeRun` and makes every later run return 409 until restart.
5. On a live-configured server the UI sends no bearer token for Preview, but the server requires the operator token for every `/api/run` request, so the advertised safe preview path always receives 401.

Repository validation and all 45 credential-free tests pass, but the suite does not cover these cases.

Ray-56:

Remaining Must Fix only:

- runBackfill — Stop before the next slot-offer call when the first call lacks a definite non-acceptance; current missing/unknown result advances.
- classifyTransportError — A polling-phase 401/403 must not be treated as definite create rejection.
- loadScenario / activeRun — A scenario-load failure leaks the active guard and prevents subsequent runs.
- public/app.js and server auth — Preview on a live-configured server omits the required token and fails.
- README — Correct unimplemented tab-close cancellation claim or implement response disconnect handling.

PR #35, opened 2026-07-26
feat(forgerelay-supplier-clarification): add safe RFQ call workflow

junsenliu:
## Summary

- add a portable Agent Skill for bounded manufacturing RFQ clarification calls
- require recipient-specific outreach basis, E.164 validation, masked summaries, explicit approval, deduplication, and durable result output
- include a dependency-free local JSON validator that defaults to dry-run and never places a call
- document fictional examples, cancellation behavior, commercial boundaries, provider onboarding, and result finalization

## Safety

- no real phone numbers, credentials, or private data
- no recurring calls or automatic retries
- no quotation acceptance, negotiation, order placement, supplier approval, or other commercial commitments
- real calls remain blocked until the CALL-E MCP route, OAuth, compatible tools, exact preview approval, and runtime gates are verified

## Validation

- `node skills/forgerelay-supplier-clarification/scripts/validate-clarification-input.mjs skills/forgerelay-supplier-clarification/assets/example-request.json`
- `node skills/outbound-call-skill-creator/scripts/check-generated-skill.mjs --skill-dir skills/forgerelay-supplier-clarification`
- `python scripts/validate_repository.py`

All checks pass locally.


Ray-56:
Thanks for the contribution. The repository, generated-skill, and happy-path fixture checks pass, but the safety contract is not yet enforced strongly enough to merge:

1. The approval preview contains only question IDs, not the question prompts or outreach basis, and the idempotency key also omits prompts, allowed context, and caller identity (`scripts/validate-clarification-input.mjs:102-133`). Editing what will be spoken can therefore reuse the same preview/key. Please bind every safety-relevant field to the exact preview and dedupe key.
2. The documented rejection cases currently pass validation: any non-empty `outreachBasis` is accepted, including a public-website-only basis, and a price/commitment question is accepted as any non-empty prompt. Either use structured fields that can be validated deterministically or return a pending-safety-review state until an agent reviews the free text; do not report the request as ready for a dry run.
3. `resultTarget` accepts any non-empty string, while the runtime contract requires a writable new local CSV path. The shipped valid fixture uses prose rather than a path. Please align and validate the contract and fixture.
4. JSON `null` crashes instead of returning a blocked validation result, and the repository-root command in `SKILL.md` should be `python3 scripts/validate_repository.py`, not `python ../../scripts/validate_repository.py`.

Please add one small runnable negative check covering these trust-boundary cases.

Ray-56:
Superseding the earlier review under the proportional Community/Experimental policy: all previously listed Must Fix items are resolved on `06016472`. Repository validation, the generated-skill check, the approved fixture, and all 7 trust-boundary cases pass. The content is ready from review; this remains unmerged only because the PR is still a draft.

Ray-56 (approved):
The frozen Must Fix list is clear on this unchanged head. Repository validation, the 7-case trust-boundary suite, and GitHub CI pass; approving under the Community / Experimental tier.

PR #39, opened 2026-07-29
feat: add verify-by-phone skill

StephenSook:
## Summary

Adds `verify-by-phone`, a skill for the case where a record claims something and only the
published line can confirm it. It places one disclosed CALL-E call, asks the directory question,
and returns either an answer grounded in the transcript or an explicit abstention.

The part that belongs in this repository is the abstention. An agent that always produces an
answer is easy to build and quietly harmful, because a confident wrong answer is
indistinguishable from a right one at the call site. This skill separates "the call happened"
from "the call established an answer", which are different facts that phone agents routinely
conflate. The platform reporting high confidence that a call completed says nothing about
whether anyone answered the question.

How it decides:

- **Span grounding.** Every extracted value carries the character offsets of the transcript text
  it came from. If no span supports a value, there is no value.
- **Graded hedging.** A lexicon separates "yes" from "I think so" from "you would have to ask
  the office manager". Hedged speech lowers the score instead of being rounded up to a yes.
- **A calibrated gate.** `calibrate.py` derives an answer-or-abstain threshold from labelled
  examples using split conformal prediction with the finite-sample correction, so the retained
  answers carry an empirical coverage level rather than a vibe. `gate.py` applies it. Bundled
  scenario data in `references/sample-scenarios.jsonl` lets a reviewer run this end to end.

Voicemail is treated as a non-answer and no message is left, since a directory fact cannot be
established from a greeting and nobody should find a robot message on their line.

## Type

- [x] New skill

## Checklist

Notes on two of those:

**Cancellation.** Nothing here recurs. There is no scheduler and no batch mode, by design, since
verification is per listing. `SKILL.md` states this under Side Effects And Cancellation: exactly
one outbound call per `--live` invocation. The returned idempotency key is documented so a retry
cannot double dial.

**Dry run is the default, not a flag you remember to pass.** Dialing requires opting in with
`--live` plus `CALLE_API_KEY`. From a clean checkout with no credentials:

```
$ python3 scripts/place_verify_call.py --org "Example Family Medicine" \
    --phone "+15550101234" --claim-accepting-new-patients yes --claim-plan "Example PPO"
recipient: +15******234
task:
You are placing a short verification call to Example Family Medicine on behalf of a records
verification service. Open with: 'Hi, this is an automated assistant calling to verify directory
information for Example Family Medicine. This call may be recorded.' ...

DRY RUN: no call placed. Re-run with --live to dial.
```

The recipient is masked in output. Every sample number in the skill is in the reserved fictional
`+1 555 0100` block.

## Disclosure

Disclosure is in the opening line rather than in a later paragraph, so the person can decline
before answering anything. If they object to an automated caller, the call thanks them and ends,
and the result records a refusal as unverifiable rather than as a failure to retry.

## Verification

Run from the repository root against this branch:

```
python3 scripts/validate_repository.py     # Repository validation passed.
```

Built during the CALL-E hackathon. The skill is self contained and has no dependency on the
application it was extracted from.


Ray-56:
Thanks for the substantial contribution. I found these current blockers:

- place_verify_call.py cannot safely retry an ambiguous create: it generates a new idempotency key every run, offers no way to reuse one, and prints it only after the request succeeds, so a lost response can lead to a duplicate call.
- gate.py can serve a polar answer when the calibrated singleton prediction set is {unknown}; require the singleton label to equal the extracted answer and reject unknown.
- reconcile_record.py invokes extraction without qhat, so extraction fail-closes every field to unknown and the documented reconciliation can never verify or contradict a record.
- Multi-question extraction scans all later user turns, so a later insurance answer can be attributed to the earlier new-patient claim. The workflow also needs to establish that the respondent represents the named organization before treating an answer as directory evidence.
- The documented quick start fails on the repository/default Python 3.9 because the scripts use newer runtime annotation syntax without declaring a minimum Python version.
- poll_result.py writes full phone/transcript payloads with ordinary 0644 permissions; save sensitive results with restrictive permissions and document retention/deletion.

Please also update the bundled calibration example: it currently reports 60.0% abstention, not the documented 43.3%.

StephenSook:
Thank you, this was a genuinely useful review. All seven were real, and two of them were bugs in the mechanism the project is built on. Pushed as one commit per finding, each with a regression test that I ran against the old behaviour first to confirm it fails rather than passing vacuously.

**1. `gate.py` could answer when the calibrated set said `{unknown}`.** Correct, and the rule is now `pset != {answer} or answer == "unknown"`. Worth recording why our own tests missed it: the reference harness derives the answer as the argmax of the class scores, and the argmax is `unknown` across the entire region where the set is `{unknown}`, so the two forms are provably identical there. A sweep over a million `(trust, qhat)` points finds zero divergences under argmax and 110826 under an extracted answer. The skill and the served path both use the extracted answer, so they could reach what the harness could not. No reported number changed, and I verified that rather than assuming it.

**2. Idempotency.** Fixed all three parts. The key is now `sha256(org, phone, claims, UTC day)` rather than a fresh `uuid4`, it is printed **before** the request rather than after it succeeds, and a failed create exits with `--idempotency-key <key>` spelled out plus a plain statement that the call may already have been placed. The reasoning for deriving rather than randomising: the natural recovery from a lost response is to run the same command again, and that has to be the safe action rather than a second dial.

The UTC-day scope is deliberate and I would rather state the tradeoff than imply the derivation is total. A key over the parameters alone would be permanent, so re-verifying the same listing next month would collide with the first call and return a stale answer without dialing. For a verification tool a silently stale result is worse than a duplicate call. It is not airtight across a UTC midnight, which is exactly why `--idempotency-key` exists and why the failure message prints it.

**3. `reconcile_record.py` never passed `--qhat`.** Correct, and it was worse than inert. On the bundled fixture the recipient says "Yes. We're accepting new patients." with a transcript span behind it, and the script labelled that field "no evidence" and printed a 50 percent posterior with UNVERIFIABLE. That reads exactly like a real reconciliation that found nothing. `--qhat` is now required rather than defaulted: failing closed is right for extraction, which has other uses, but a script whose only output is a verdict cannot distinguish "the call established nothing" from "nobody looked", and printing one as the other is the failure this project exists to prevent.

**4. Multi-question attribution.** Fixed, and this was the one I found most alarming. A claim's answer window now closes when the agent asks a different tracked question. Previously, a respondent who deflected on new patients and later answered the insurance question had the new-patients claim reported as `yes`, trust 0.9, **not abstaining**, span-grounded to a sentence plainly about insurance. A wrong answer carrying a citation is worse than no answer, because the citation is what invites belief. Fixed in every place that extracts multiple claims from one transcript.

**The respondent-identity half of that finding was the best catch in the review**, because nothing in the design addressed it at all. Numbers get reassigned, and "Yes, we are open and taking new customers" is a perfectly true sentence from whoever holds the number now. The call script now establishes it reached the named organization before asking anything and ends the call if it did not; extraction reports `organization_denied` and drops every span when it is set; and reconciliation stops before the arithmetic rather than weighing the answers down, since reaching another party is not weak evidence but no evidence about that listing.

**5. Python floor.** Reproduced on 3.9.6: `TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'` at import, before argparse ran. Every script now has `from __future__ import annotations`, and SKILL.md states Python 3.9 or newer plus the two dependency exceptions (`calle-ai` for the dialing step, `cryptography` for the optional attestation checker). Verified by running the entire documented quick start on 3.9.6 rather than inferring it from the syntax.

**6. Payload permissions.** Now `0600`, applied on open and again with `chmod`, since `O_CREAT` only honours its mode when it actually creates the file and re-running over an existing `0644` path would otherwise keep it loose. Tested both cases. Console output was already masked, which is part of why this was easy to miss: the masking implied care that the file did not receive. `safety.md` now says what the file contains, that it is the one unmasked artifact, and that nothing after extraction needs the raw transcript.

**7. The calibration example.** You were right, and this is the one I am least comfortable with, because the project's entire claim is that its numbers regenerate from the code with one command. A hand-typed figure contradicting the program refutes that claim better than any critic could. Corrected to 60.0 percent by regenerating the block from the command rather than editing the digits.

Both worked examples are now diffed against real command output in the source repository's CI, so an example that drifts fails on the commit that causes it instead of surviving to review. I also swept the remaining hand-typed numbers in the docs and pinned the verdict thresholds to the constants they come from.

One thing your review surfaced that I want to name: the reconciliation documentation was **right** and the code was wrong. `examples.md` already documented the `+1.36 bits` and 72 percent posterior that only the fixed code produces. The docs described the intended behaviour and the code silently never performed it, which is a failure mode I had not been testing for at all, and is now what the output-diffing tests exist to catch.

Ray-56:
Thanks for addressing the previous review; those fixes now validate. I found these remaining current blockers:

- Identity is still fail-open: extraction detects explicit denial but does not require positive confirmation that the respondent represents the named organization before accepting directory answers.
- If the agent asks both tracked questions in one turn, both extraction windows open on the same reply, so one answer can be assigned to both claims. Require one question and response at a time or abstain on ambiguous combined turns.
- A voicemail or non-answer with no transcript still exits with an error instead of emitting the promised explicit abstention records.
- A live run with neither supported claim still places a generic listing-current call, but the extractor cannot produce that result; reject such a run before dialing.
- Add the repository-required medical, legal, financial, and emergency boundaries, and clarify that the live calle-ai step requires Python 3.11 even though the standard-library scripts support Python 3.9.

StephenSook:
Thanks, and thanks for the second pass. All five were real, and one of them turned out to be two bugs. Same discipline as before: verified against source before acting, one regression test each, every test run against the old behaviour first to confirm it fails rather than passing vacuously.

**1. Identity was fail-open.** You are right, and the distinction matters more than I had it: detecting an explicit denial is not a check, because absence of a denial is not confirmation. Wrong numbers, answering services, shared reception desks and reassigned lines all produce cooperative respondents who answer clearly and are not the listing. Extraction now takes `--org` and requires positive confirmation, abstaining without it.

Writing that check surfaced a second bug in the check itself, which is worth reporting rather than quietly fixing. My first version accepted "Yes, we are." as confirmation. That is the ordinary way to answer "are you accepting new patients", so every cooperative respondent confirmed themselves and the gate was decorative. My own test caught it. Confirmation now requires the respondent to say the practice's distinctive name, or to affirm an actual identity question, tracked by which question the agent last asked:

| transcript | confirmed |
|---|---|
| cooperative, never identifies itself | no |
| "Northside, this is the front desk" | yes |
| "Is this Northside?" then "Yes, it is" | yes |
| affirms the new-patients question instead | no |

Worth noting: neither of the two real calls I have recorded contains a positive confirmation, so both now abstain. That is the correct reading rather than a regression. The respondent answered the question but never established which organization was speaking.

**2. Combined turn.** Fixed, and this was the one I found most uncomfortable, because the old output looked perfect. A single "Yes." went to both claims, each non-abstaining, each citing the same span. At least one of those is wrong and both look identically confident. It now abstains on both, and `build_task` asks one question at a time so it should not arise in the first place.

**3. Voicemail and no-transcript.** Fixed. It emits an explicit abstention record per claim with gate `no-transcript` and exits 0. You are right that this was backwards: nobody answering is the most common real outcome, it is precisely the case the explicit abstention exists for, and erroring broke every downstream caller including reconciliation on the ordinary path rather than an exceptional one.

**4. Claimless run.** Rejected before dialing. `build_task` now raises rather than offering the generic listing-current fallback, since no claim pattern could ever match it. A call that cannot produce a recordable answer is worse than a crash, because it spends real money and a real person's time to learn nothing.

**5. Boundaries and Python floors.** Added the required medical, legal, financial and emergency boundaries. They are load-bearing for this skill rather than boilerplate, so they say explicitly that it handles directory logistics only, collects no clinical detail in either direction, is never a route to urgent care, and points at local emergency services and 988 for mental-health crises.

On the versions, you identified something genuinely misleading rather than merely absent. There are two floors and I had documented one: **3.9** for the standard-library scripts, which I verified by running the whole quick start on 3.9.6, and **3.11** for placing a live call, because that is `calle-ai`'s own declared `requires-python` and not a choice this skill makes. SKILL.md now states both in a table.

**Root cause worth naming**, since it explains finding 1 rather than excusing it: this skill's `build_task` had drifted from the reference implementation it was extracted from. The reference has always asked the agent to confirm the organization before asking anything; this copy never did. So the call script was not producing the confirmation that extraction should have been demanding, and neither end noticed. Both now ask identity first and one question at a time.

I also added `references/sample-call.json` so every step after the call runs with no credentials, and rewrote the reconciliation example to show the interesting case instead of the flattering one: one field agreeing, one contradicting, net 34 percent, reported as UNVERIFIABLE rather than rounded in either direction.

Ray-56:
Thanks for the latest fixes; the prior blockers now validate. Two current issues remain:

- Positive identity is still fail-open for a negated organization name. For example, `No, this is not Example Family Medicine` is not matched by `IDENTITY_DENIALS`, then the distinctive-name token makes `organization_confirmed()` return true. If the call continues, later answers can be attributed to the wrong party. Scope confirmation to an affirmative response to the identity question, make explicit negation win, and add this regression case.
- The operator-facing follow-ups are stale: `references/examples.md`, `calibrate.py` output, and `poll_result.py` still show extraction without the now-required `--org`, so those documented commands produce `identity-unconfirmed` abstentions instead of the displayed successful result. Please update every generated/example command consistently.

Repository validation, compilation, calibration, reconciliation, combined-question/no-transcript behavior, idempotency, and 0600 payload handling otherwise pass with local credential-free tests; no real call was placed.

StephenSook:
Both fixed. Thank you for catching the first one, because it was worse than I had characterised my own bug: not fail-open but **inverted**.

**1. Negated organization name.** You are right that `IDENTITY_DENIALS` missed it and that the distinctive-name token then confirmed it. So "No, this is not Example Family Medicine" produced `organization_confirmed = True`, and the call would have continued attributing answers to a party that had just told us it is not the listing. A wrong answer that arrives with confidence attached is the failure mode this project exists to prevent, and I had built a fresh instance of it while fixing the previous one.

Denial now wins, is evaluated first, and applies to the whole call. A name mention confirms only when it is not preceded by a negation.

The detail I want to flag, because it is where a naive fix goes wrong: the negation window is scoped to 40 characters **before** the name rather than applied to the whole turn.

| transcript | denied | confirmed |
|---|---|---|
| "No, this is not Example Family Medicine" | yes | no |
| "No, this isn't Example Family Medicine, this is Riverside Auto", then answers helpfully | yes | no |
| "Yes, this is Example Family Medicine, **not** the Buckhead office" | no | yes |
| "Example Family Medicine, how can I help you?" | no | yes |
| cooperative, never identifies itself | no | no |

Row three is the reason for the scoping. A whole-turn negation check fixes the inverted case by breaking the ordinary one, and I would rather not trade one silent wrong answer for another. All five are regression cases now, including the denial-then-cooperative-answer sequence.

`main()` also now passes `org` into `organization_denied`, without which the new denial would have been caught in one call site and missed in the other.

**2. Stale operator-facing commands.** Correct, and this one is worth naming as a pattern rather than a slip. `calibrate.py` and `poll_result.py` both printed a next-step command without `--org`, and `examples.md` documented one, so an operator following the output exactly would get `identity-unconfirmed` abstentions instead of the result the surrounding output implies. `poll_result.py` was also omitting `--qhat`.

Every printed and documented command now carries the flags needed to reproduce what it displays, and there is a test that walks both scripts and both docs to enforce it.

This is the same class of defect as the documented-number bug from your first review: **output that describes a behaviour the command as written cannot produce.** Both times the code was right and the thing telling the operator what to run was wrong. Adding a required flag is a change to every generated command that mentions the tool, and I had treated it as a change to one script.

254 tests, all gates green, and the eval output regenerates byte-identical, so nothing in the reported numbers moved across any of the three review rounds.

Ray-56:
Thanks for the focused fix. One identity blocker remains: negation is checked only before an organization token. A response such as `Example Family Medicine? No, this is Buckhead Clinic` contains an affirmative-looking token before the correction, so `_name_mentions()` reports an affirmative mention and the call is still attributed to the wrong organization. Bind confirmation to the identity question/response semantics and make explicit denial or correction win regardless of whether it appears before or after the name; add this exact regression case.

The updated commands, repository validation, compilation, and calibration pass locally. No provider call was placed.

StephenSook:
Fixed, and this pass caught the exact blind side of the defense I made two rounds ago. I justified the before-the-name window with the row "Yes, this is Example Family Medicine, not the Buckhead office", and position-dependence is precisely the failure that scoping could not survive: your transcript opens with the name, nothing precedes it, and the echo read as an affirmative mention.

Confirmation is now bound to what the turn does rather than where the tokens sit. Three rules, in order:

1. **An explicit correction ("No, this is ...") wins wherever it appears in the turn**, and is classified by WHO it names: somebody else denies; the listing itself confirms. The second half fixes a bug in the other direction that the window had been hiding: "No, this is Example Family Medicine", correcting a mispronounced name, used to be swallowed as a false denial.
2. **A name mention inside a question asserts nothing.** "Example Family Medicine?" is the respondent asking, not identifying. The ordinary front-desk greeting keeps its comma ("Northside, how can I help you?") and still confirms.
3. Only then does the per-clause before-the-name negation window decide, which preserves the ordinary confirmation cases.

| transcript | denied | confirmed |
|---|---|---|
| "Example Family Medicine? No, this is Buckhead Clinic" | yes | no |
| "Example Family Medicine?" alone, then cooperative answers | no | no, every claim abstains identity-unconfirmed |
| "No, this is Example Family Medicine." | no | yes |
| "Northside, how can I help you?" | no | yes |
| "Yes, this is Example Family Medicine, not the Buckhead office" | no | yes |

Your exact transcript is the named regression case, and the other four rows are pinned beside it. 258 tests, lint and type gates green, `scripts/validate_repository.py` passes on this branch, the script still parses under a Python 3.9 feature gate, the documented example outputs are unchanged, and no provider call was placed.

Ray-56:
Current blocker:\n\n- Identity questions that repeat the listing name after a comma are still treated as positive identity confirmation. For example, “Example Family Medicine, right?”, “Example Family Medicine, is that who I reached?”, and “Example Family Medicine, what do you want?” all produce organization_confirmed=true. A question or echo must not establish identity, otherwise later answers can be attributed to the wrong listing.\n\nPlease require an actual affirmative response or unambiguous self-identification instead of using comma presence to distinguish a greeting from a question.

StephenSook:
Fixed, and the comma is gone rather than adjusted. You were right that it was doing work punctuation cannot do: it was there to protect the ordinary front-desk greeting from the echo rule, and a tag question has the same comma, so the protection covered the attack.

Confirmation is now bound to the speech act, and only two things establish identity:

1. **An unambiguous self-identification.** A declarative statement of the name ("Northside Family Medicine, this is the front desk"), a greeting phrase that cannot itself be a question ("thanks for calling X", "you've reached X", "welcome to X"), or an affirmative that states the name ("Yes, this is X, how can I help you?"). All of these are anchored to the start of the clause and must run to the name, so a quoted or discussed phrase is not a greeting.
2. **An actual affirmative answering an identity question the agent really asked, that also named this listing.** Tightened in the same pass: a bare "confirm" in the agent's turn used to arm this path, so "Can you confirm you are accepting new patients?" followed by "Yes" read as identity. That is the decorative check from the second review, one layer up.

| respondent turn | confirmed |
|---|---|
| "Example Family Medicine, right?" | no |
| "Example Family Medicine, is that who I reached?" | no |
| "Example Family Medicine, what do you want?" | no |
| "Example Family Medicine?" | no |
| "This is Example Family Medicine?" | no |
| "Thanks for calling Example Family Medicine, how can I help you?" | yes |
| "Yes, this is Example Family Medicine, how can I help you?" | yes |
| "Example Family Medicine, this is the front desk." | yes |

**One consequence I would rather state than have you find.** A bare "Northside, how can I help you?" no longer confirms. It is the same construction as "Northside, what do you want?", and I could not find an honest rule that keeps one and rejects the other, so it abstains. The agent asks "Is this the office of X?" as its first question, so the ordinary route to confirmation is an answer to that.

**Then I ran an adversarial pass over my own fix, and it found seventeen more.** Reporting them because several are the same class you have been probing since round two, and because two of them predate this round entirely. Each was reproduced against the code and survived an independent refutation pass; thirteen further candidates were refuted and dropped.

- **Your own sentence with the punctuation removed.** "Northside Family Medicine, is that who I reached" confirmed. My question test had three signals and this tripped none: no "?", the first word is not interrogative, no trailing tag. I had written in the test docstring that the rule was punctuation-independent; it was independent only for the two shapes those tests happened to cover. An interrogative anywhere in the clause now counts, and a tag question is never rescued by a greeting ("You've reached Northside, right?" was confirming).
- **The name in a position that is not self-identification.** "Let me transfer you to Northside", "For Northside, press two", "Northside is upstairs, this is the pharmacy" and "Northside is at 555-0102" all confirmed. Naming a place is not being it. The name must now open its clause or follow a self-identifying phrase, and a third-person predicate about the listing is a description rather than an identification.
- **A repudiation in a different clause.** "Who? Northside Family Medicine. Never heard of them." confirmed on the middle clause, because the negation window is scoped to one clause and cannot reach across a sentence boundary.
- **One distinctive word standing in for the whole name.** "New Life Health Center" reduces to {new, life}, so "we are not accepting new patients" both matched the name and negated it: an ordinary refusal denied the call, and an ordinary yes confirmed an identity nobody had claimed. Every distinctive word of the listing must now appear in the clause.
- **The plainest denial was missing.** The agent's first question is "Is this the office of X?" and a respondent who answered "No." left the call merely unconfirmed rather than denied.
- **Predating this round, found by writing the test for the rule beside it:** "No, we are full at the moment" matched the correction shape ("no ... we are ...") and denied the whole call on an ordinary refusal. The correction trigger is now only a phrase that introduces an entity.
- **Also predating it, and in both extractors:** a single typographic apostrophe flipped verdicts, because every contraction in the cue lexicons is written with the ASCII quote. "No, it's Northside" confirmed and the same sentence with U+2019 denied, and separately "we aren't taking new patients" with U+2019 extracted the opposite answer. Both extractors now fold typographic punctuation through one shared definition, and the fold is one character to one character so every cited span offset is unchanged.

Two earlier ones from a second-model pass are in the commit before this: gating the name check on the question test lost a denial the previous version caught ("We're not Northside, right?"), and let one greeting outvote an explicit denial in the same breath. Negation is now decided before the question test, because the fail-closed direction never needs a punctuation argument.

One case I deliberately left fail-closed rather than fix: "You have reached Buckhead Clinic. For Northside Family Medicine, press two." does not confirm, and it also does not deny. A rule general enough to deny it would deny a real voicemail greeting that names the practice after a comma, which is a false denial on a legitimate call, so it abstains instead.

302 tests, every new case verified failing against the previous code first. Lint, format and type gates green, `scripts/validate_repository.py` passes on this branch, the script still parses under a Python 3.9 feature gate, the evaluation harness regenerates byte-identical so no documented number moved, and no provider call was placed.

StephenSook:
Thank you for merging, and for five rounds of genuinely load-bearing review.

Worth saying plainly, since it is unusual: every blocker you filed was real, and four of them were in the mechanism the whole project rests on. The gate that could answer while the calibrated set said `{unknown}`, the answer credited to an earlier question, identity being fail-open, and then the comma standing in for the difference between a greeting and a question. Our own harness could not have found the first one, because it derives the answer as the argmax and the bug only exists for an extracted answer.

The pattern I am taking away is that each round landed exactly on the scope boundary of the previous fix, so before shipping the last one I attacked it myself and found seventeen more, including my own new rule failing on your sentence with the question mark deleted.

Every case is pinned as a regression test that was run against the old code first. If anything surfaces in the skill after the merge, ping this thread and I will turn it around the same day.

PR #40, opened 2026-07-29
Add invoice-exception manager briefing skill

JAYAPLY:
## Summary
- Adds a portable, safety-first Agent Skill for a one-time manager briefing about an invoice exception.
- Keeps every business decision in the existing human review application.
- Documents dry-run-first planning, recipient and scope verification, idempotency, cancellation, redaction, and server-side credential handling.

## Safety
- No automatic, recurring, or duplicate calls.
- No credentials, personal phone numbers, recordings, voicemails, or transcripts in this contribution.
- A failed or unanswered call never triggers an automatic retry.

## Validation
- python3 scripts/validate_repository.py
- git diff --check

### DIFF
--- a/skills/invoice-exception-manager-briefing/SKILL.md
+++ b/skills/invoice-exception-manager-briefing/SKILL.md
@@ -1,0 +1,89 @@
+---
+name: invoice-exception-manager-briefing
+description: Prepare a controlled, one-time CALL-E briefing for an authorized manager about an invoice exception, while keeping the human decision in the existing review application.
+license: MIT
+---
+
+# Invoice Exception Manager Briefing
+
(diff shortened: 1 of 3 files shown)


Ray-56:
Thanks for the safety-first contribution. I found these current blockers before it can merge:

- The workflow verifies the stored manager role and destination, but it does not require the callee to positively confirm that they are the authorized manager before any invoice context is disclosed, and it has no do-not-leave-voicemail rule.
- The idempotency guidance is not concurrency-safe and does not handle an ambiguous provider timeout. Require a durable atomic reservation/result record, reuse the same key while the outcome is unknown, and never create a new-key retry until the original is confirmed absent.
- Cancellation is documented only before dispatch. State the point of no return and what the operator can do after provider acceptance; also resolve the conflicting instructions that the kill switch must be off versus enabled.
- Add the repository-required medical, legal, and emergency off-topic boundaries alongside the existing financial decision boundary.

PR #41, opened 2026-07-29
feat(apps): add phone approval gate app and deployment-approval-call skill

zkasuran:
## Summary

Adds `apps/typescript/phone-approval-gate`, a runnable app that puts one CALL-E
call in front of an irreversible automated change, plus the roadmap's
`deployment-approval-call` skill, which drives the same CLI from an agent.

The problem: a pipeline or an agent stops in front of a production deploy, a
database restore or a refund batch, and the owner is not at a keyboard. In
practice that either waits hours or gets tapped through on a phone screen without
anybody reading what they approved. The gate rings the enrolled owner, reads the
change out loud and reports approval only when a live person reads back a
one-time code that was printed on the request. A bare yes is never an approval.

What the app does:

- one CALL-E call task per approver, in ladder order, stopping at the first
  decision, with `single` or `dual` control and a distinct one-time code per
  approver
- `result_schema` for a strict decision contract, plus its own read of the
  recipient transcript turns. The extracted result is corroboration. When the two
  disagree the gate refuses rather than guessing
- fail-closed reasons for voicemail, no answer, no transcript, wrong code, low
  completion confidence, timeout, window expiry and API errors
- an idempotency key bound to the request, the approver, the attempt and a digest
  of the canonical call payload, so a retried workflow step reuses the same call
  while an edited request gets its own key. Each attempt's secret is reserved on
  disk with `O_EXCL` before the phone rings, so two concurrent runs cannot hold two
  codes for one provider key
- an ambiguous create or poll is never reported as a failure. The gate replays the
  key to find the call CALL-E already holds, and if that cannot be settled it
  records `state_unknown`, stops the ladder and names the call to reconcile
- hash-chained approval records. `verify` re-links the chain, re-reads the
  decision from the stored excerpt and recomputes every recorded verdict from the
  recorded inputs with the same functions the gate ran live
- exit codes 0, 10, 20 and 30 for a pipeline to gate on, a composite GitHub
  Action and an example deploy workflow

Side effects and safety: exactly one call task per approver per run, nothing
recurring is ever created, so there is no schedule to cancel. Stopping the
process stops the ladder. Preview is the default and contacts nothing. Live mode
needs `--live` plus `CALLE_API_KEY` from the environment, never from the request
file. Numbers are masked in output and records, transcripts are trimmed to the
decisive turns and the code is never stored, only a salted digest of it. The call
script ends on voicemail or a menu system without describing the change, and it
refuses every instruction made on the call other than this one approval.

`docs/threat-model.md` states plainly what a phone approval proves and does not
prove. The out-of-band handling follows NIST SP 800-63B (six digit secret from an
approved generator, single use, a ten minute ceiling on the window, capped failed
attempts, no compare-and-approve prompt), including the part where SP 800-63B
calls the PSTN a restricted channel for out-of-band verification.
`docs/audit-format.md` documents the record and the three checks.

Two directories in one PR because they are one contribution: the app is the
implementation and the skill is the agent-facing entry point that shells out to
it, with no duplicated logic. Happy to split them if you would rather review them
separately.

## Verification

```text
npm test        101 tests pass
npm run demo    4 end-to-end cases, chain verified, forged verdict rejected
```

CALL-E in `fake/calle-server.ts`, which speaks the documented wire contract
(201 on create, `Idempotency-Key`, snake_case payloads, error envelopes). The
committed `examples/audit.example.jsonl` is the unedited output of `npm run demo`
and passes `verify`.

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

Also adds one new skill, `skills/deployment-approval-call/`, as the agent entry
point for the same app.

## Checklist


Ray-56:
Thanks for the safety-focused contribution. I found these current blockers:

- `allow_structured_only` can approve with no transcript, contradicting the stated fail-closed rule that structured output only corroborates live transcript evidence.
- Create/wait ambiguity is recorded as an API failure and the ladder continues. An accepted call can still be running while another approver is called; reconcile the original call under the same key before advancing.
- The stable key binds only request/approver identifiers, not canonical request content, and there is no durable atomic reservation across concurrent runs. Concurrent retries can generate different secrets for the same provider key and place extra ladder calls.
- The approval window is checked before a call but not after its result, so a late response can approve after expiry. The audit record is optional and append is not concurrency-safe.
- Arbitrary `--base-url`/`CALLE_BASE_URL` receives `CALLE_API_KEY` without HTTPS/trusted-host validation. Also establish a live expected approver before reading change details.

`diff --check`, repository validation, and typecheck pass. Tests/demo could not run here because the sandbox denied the `tsx` IPC socket, and the PR has no lockfile/CI checks. No real call was placed.

zkasuran:
Thanks for the review. All five were real, so here is what changed for each, in
your order. Six commits on the branch, ending at d0668ec.

**1. `allow_structured_only` could approve with no transcript.**

The option is gone rather than narrowed, because the rule the module states is
that the extraction corroborates the recipient turns and never replaces them.
`src/decide.ts:99` refuses any completed call with no recipient turns
(`no_transcript_evidence`) and `src/decide.ts:204` reads `structured_result` for
corroboration only, so it can no longer supply `reached_person` or a code match.
`policy.allow_structured_only` is out of the request schema, the resolved policy
and the record (`src/types.ts`, `src/config.ts`, `src/audit.ts`). The test that
used to assert the opt-in approved now asserts the refusal, including that
`reached_person` and `code_match` stay false: `test/decide.test.ts`, "a missing
transcript blocks approval, whatever the extracted result says".

**2. Create/wait ambiguity recorded as an API failure while the ladder
continued.**

`src/calle.ts:51` splits the two cases: a 4xx the server chose to send means no
call was placed, while no reply at all, a 408, a 409 on the key or a 5xx can each
sit on top of a call that was accepted. `placeCall` in `src/gate.ts:126` handles
the second class by replaying the same idempotency key first
(`src/gate.ts:146`), which returns the call CALL-E already holds for that key.
Only when that also fails is the attempt recorded with `call_status:
state_unknown` (`src/decide.ts:159`) and reason `call_state_unknown`
(`src/decide.ts:67`). The ladder then stops (`src/gate.ts:303`) with a line
naming the call to reconcile. A refused create still advances the ladder, since
nothing was placed. The poll path got the same treatment: a `waitForResult`
failure reads the call once. If that read fails the record keeps the call id, so
somebody can still reconcile it.

Tests: `test/gate.e2e.test.ts` "a create whose reply was lost is reconciled under
the same key" (one call created, approved, record verifies), "a call state that
cannot be settled stops the ladder and says so" (second approver never called),
"a refused call is not ambiguous, so the ladder moves to the next approver", and
`test/gate.test.ts` "a poll that cannot be read back leaves the state unknown and
stops the ladder".

Chasing this turned up a bug of mine underneath it: an attempt with no call built
its outcome without going through `attemptOutcome`, so a record of a refused
create failed verification because the replay read `api_error` as `timed_out`.
Both `api_error` and `state_unknown` are decided in `attemptOutcome` now, covered
by two records tests in `test/audit.test.ts`.

**3. The key bound only identifiers, with no durable atomic reservation.**

`src/script.ts:160`: the key keeps `pag-<request>-<approver>-<attempt>` and adds
12 hex of sha256 over the canonical JSON of the create payload, using the audit
module's `canonicalJson` so a key and a record hash are built the same way. An
edited request now gets a different key instead of replaying a call about
something else.

`src/reserve.ts` is the reservation. Each attempt writes
`<state>/<request_id>/<approver>-<attempt>-<content digest>.json` with `O_CREAT`
and `O_EXCL` (`flag: "wx"`, `src/reserve.ts:125`) before the phone rings. The
first run owns the attempt and every later run adopts its code, so two runs
cannot hold two secrets for one provider key. The directory defaults to
`.phone-approval-gate` beside the record file, `--state` moves it and `null`
turns it off for unit tests. The file holds the code in clear at `0600`: that
trade is written down in `docs/threat-model.md`. The short version is that it is
the same code already printed on the request channel and it is useless once the
window closes. Approver ids are now validated to the same characters
`request_id` takes (`src/config.ts:129`), because they travel into the key and
into that file name.

Tests: `test/reserve.test.ts` (6 checks, including the adopt path and a corrupt
reservation failing closed) and `test/gate.e2e.test.ts` "a second run adopts the
reserved secret instead of inventing one", where the second run generates a
different code and the gate still uses the reserved one. Before the reservation
that run walked the ladder on a code mismatch.

**4. Window checked before a call but not after its result. Record optional,
append not concurrency safe.**

Three separate changes.

- `decidedInWindow` (`src/gate.ts:74`) runs on every result and lands in the
  evidence as `within_window`, which `src/decide.ts:92` refuses on. It reads two
  clocks because they catch different things: the local clock catches a result
  that arrived after the window closed, the call's own `completed_at` catches a
  call that finished outside this run's window at all, which is what a replayed
  key returns. 60s of skew is allowed on the provider timestamp.
- `--audit` is required for a live run (`src/cli.ts:149`) and `action.yml` ships
  a default `audit-file`, so a live run always leaves a record. `runGate` keeps
  `auditPath` optional for unit tests only.
- `withAuditLock` (`src/audit.ts:148`) wraps the read-the-tail-then-append, since
  that is where the race actually is: two runs would both link to the same tail
  and verification would then call the chain broken. Lock file created `wx`, 30s
  stale break, 10s wait, released in a `finally`.

Tests: `test/gate.test.ts` for a decision that lands after the window closed and
for a call that finished before the window opened, both `window_expired`;
`test/decide.test.ts` for a rejection still counting when the window has closed;
two lock tests in `test/audit.test.ts` (a held lock appends nothing and throws, an
abandoned lock is broken).

**5. Arbitrary base URL received the key. Establish the approver first.**

`assertTrustedBaseUrl` (`src/calle.ts:70`) runs inside `createSdkPort` before the
SDK import and before any client holds the key (`src/calle.ts:104`). https
anywhere, plain http only for `localhost`, `127.0.0.1` or `::1` so the local fake
CALL-E and the tests keep working, everything else refused with a `ConfigError`
that names both `--base-url` and `CALLE_BASE_URL` and exits 30. It refuses rather
than warns, since a warning still sends the key.

On the approver: `src/script.ts:42` has the caller ask who is on the line before
any change text. `src/script.ts:45` tells it to read none of the change to anyone
else. The question is open ("who am I speaking with?") rather than "am I
speaking with Alice?" on purpose. A yes or no prompt there would put an approval
word in a recipient turn before the decision was even asked for. Those are the
turns the gate scores.

Tests: `test/calle.test.ts` (5 checks, including the cloud metadata address and
`localhost.attacker.example`) and a script test asserting the question's position
precedes both the title and the summary.

**On your environment note.**

`npm test` is now `node --import tsx --test test/*.test.ts`. The tsx CLI owns that
IPC socket: it is the parent, so it dies when it cannot have one (I reproduced an
EACCES crash on the pipe by denying the directory). The loader path only attempts
the connection and swallows the failure, `node_modules/tsx/dist/client-*.mjs`
registers `r.on("error")` and resolves with nothing, so a denied socket does not
stop the run. The same 101 tests pass both ways here.

No lockfile added. No TypeScript app in this repo commits one, the three under
`apps/typescript/` ship a `package.json` only. The Python apps commit `uv.lock`. I followed the repo rather than introducing the first npm lockfile.
Say the word if you want a `package-lock.json` for the action's `npm install` and
I will add it.

**Verification** (Node 22.22.2, from `apps/typescript/phone-approval-gate`):

```text
npm test                                101 tests, 101 pass, 0 fail (73 before)
npm run demo                            exit 0, 4 records, chain and verdicts hold,
                                        forged verdict rejected
```

No real call was placed. `examples/audit.example.jsonl` is the demo's own output,
unedited.

Still open, documented rather than solved: the reservation and the record lock
live on one filesystem, so two runners with separate checkouts of the same request
still lean on the provider idempotency key alone. `docs/threat-model.md` says so
in the row about two runs racing.

Ray-56:

- After waitForResult times out or fails, any successful getCall is treated as resolved. If it returns queued or in_progress, the attempt becomes timed_out and the ladder advances. An isolated stub reproduced two approvers being called while the first call remained live. Nonterminal read-backs must remain unresolved and stop the ladder.
- The O_EXCL secret reservation coordinates only one filesystem. In code_from_request mode the provider payload and idempotency key do not contain the off-channel secret, so runners with separate state directories can generate different codes for the same provider call; the losing runner can see a mismatch and advance its ladder. This needs a durable shared ownership boundary, not only a per-checkout file.
- assertTrustedBaseUrl accepts every HTTPS host, so a mistyped or attacker-controlled HTTPS base URL still receives CALLE_API_KEY, contrary to the README claims. Also, appending an existing audit file does not enforce the documented 0600 mode.

Typecheck, 101 tests, the local fake demo, repository validation, and diff check pass. No credentials or real calls were used.

zkasuran:
Thanks. All three were real. Two commits on the branch, ending at `396997a`.

**1. A nonterminal read-back resolved the attempt and the ladder moved on.**

Your stub is a test now: `test/gate.test.ts:123`, a port whose `waitForResult`
throws and whose `getCall` answers `in_progress`. It asserts one call created, one
attempt, so the second approver is never rung.

The terminal statuses live in one set, `src/decide.ts:56`, read by the gate when
it reads a call back and by `attemptOutcome` when it turns a status into an
outcome, so the two cannot drift. `src/decide.ts:99` refuses to resolve anything
outside it: the attempt records `call_status: state_unknown` with the call id and
the status it saw, reason `call_state_unknown`. The ladder stops at
`src/gate.ts:340` with a line naming the call to reconcile. `timed_out` is gone as
a reason, because nothing can produce it once a live call stays unresolved.
`test/decide.test.ts:231` runs nine statuses through it with an approval sitting in
the transcript and none of them resolve.

Sweeping that class turned up one of mine underneath it. The set carried
`no_answer`, `busy` and `voicemail`. Those are not CALL-E call statuses at all.
`CallStatus` in the schema is `queued`, `in_progress`, `completed`, `failed`,
`canceled`. A no answer arrives as `failed` with a failure code, which is how
the app already reads it. So the set was three values wide of the API and the
README documented them as real. Both fixed.

**2. The reservation coordinated one filesystem only.**

The reservation is no longer the boundary. The code is derived rather than minted:
`deriveSecret`, `src/secret.ts:141`, is HMAC-SHA-256 over the request digest, the
approver and the attempt, keyed with operator key material, with the digits taken
out of the MAC the way RFC 4226 section 5.3 does it and the phrase words picked by
rejection sampling so the pick stays uniform. Two runners that share nothing but
that key derive the same code, so there is no race to lose.

Key material comes from `CALLE_APPROVAL_CODE_KEY` or `--code-key-file`
(`src/config.ts:296`). 128 bits is the floor, which is what RFC 4226 section 4
sets for a shared secret. A key file other accounts can read is refused before the
key is read (`src/config.ts:269`), rather than fixed for the operator, because a
key others could read has already been exposed. `code_from_request` refuses to run
with no key at all. `liveness_phrase` needs none, since the phrase is spoken on the
call and never printed on the request channel. `src/gate.ts:263` derives over the
same content the reservation is keyed on, so the code and the reservation cannot
end up bound to different requests. The O_EXCL file stays as the local fast path
and the local audit trail.

Your losing-runner case is closed from the other side too. A code that comes back
and is not this run's is now its own fact: `unmatched_code_spoken`
(`src/decide.ts:132`) separates nobody reading a code back (`secret_not_returned`)
from somebody reading back a code this run does not hold (`code_mismatch`). The
second stops the ladder at `src/gate.ts:357` instead of ringing the next approver
behind that person's back. The foreign code is redacted out of the record as
`[digits]`, because it is still live for whoever does hold it.

Tests: `test/gate.e2e.test.ts:256` (two runs, separate state directories, same
derived code, one call created), `:271` (a different code per approver and per
attempt), `test/secret.test.ts:96`, `:114`, `:123`, `test/config.test.ts:105`,
`:124`, `test/gate.test.ts:159`, `test/decide.test.ts:108`,
`test/secret.test.ts:86`.

The trade, written down rather than hidden: an operator now holds one more secret,
and if it leaks every code it covers becomes predictable. `docs/threat-model.md`
has a row for that with the rotation story. What is still not durable is the record
file itself. Two runners with separate checkouts keep separate chains and the
provider idempotency key is what stops the second call. The same table says so.

**3. Any https host, plus the record mode on a file that already existed.**

`assertTrustedBaseUrl` is an allowlist (`src/calle.ts:109`). Allowed:
`api.heycall-e.com`, which is the SDK default, plus loopback for the local fake.
Anything else has to be named in `--allow-host` or `CALLE_ALLOWED_HOSTS`
(`src/cli.ts:178`). Matching is exact and lowercased with the port ignored, never a
suffix, so `api.heycall-e.com.attacker.example` fails. A wildcard entry is refused
rather than read literally, since somebody who writes one means it to match
subdomains. It still runs before the client that carries the key exists
(`src/calle.ts:153`).

`appendRecord` (`src/audit.ts:135`) opens the record file and fchmods the
descriptor to 0600 on every append. You are right that the old code only passed a
mode to the write, which does nothing once the file exists, so the README promise
held for a new file only. fchmod on the descriptor rather than chmod on the path,
so no path can be swapped in between. A target that is not a regular file is
refused. One more place in the same class: an adopted reservation whose mode is
wider than 0600 is refused rather than read (`src/reserve.ts:74`), because nothing
here ever writes one that way.

Tests: `test/calle.test.ts:37`, `:49`, `:81`, `test/audit.test.ts:133`, `:145`,
`test/reserve.test.ts:81`.

**Docs.** The README says exactly the above for the base URL, the record mode and
the new code key. The retry section leads with the derivation. In
`docs/threat-model.md` the credential row is rewritten, the two-runs row is
rewritten, there is a new row for the code key leaking and data handling covers
the new redaction plus the mode.

**Verification** (Node 22.22.2, from `apps/typescript/phone-approval-gate`, repo
root for the validator):

```text
npm test                                120 tests, 120 pass, 0 fail, 0 skipped (101 before)
npm run demo                            exit 0, 4 records, chain and verdicts hold,
                                        forged verdict rejected
```

No real call was placed. Everything above is against the local fake CALL-E over
the real `@call-e/calle` client, with no credentials and nothing beyond localhost.

Ray-56:
Thanks for the update. Two retry and expiry blockers remain:

- In liveness_phrase mode, separate runners generate different random phrases because no shared code key is resolved. The phrase changes the call payload digest and provider idempotency key, so concurrent retries can place duplicate approval calls. Derive the phrase deterministically or persist one content-bound key across runners.
- decidedInWindow accepts missing or malformed completedAt as current. A stale idempotent replay can therefore approve a fresh window. Missing or invalid provider completion time must fail closed.

zkasuran:
Both blockers are fixed on b3888ab. `main` is merged into the branch first: #43 landed and rewrote the app index rows, so the conflict was `README.md` plus `apps/README.md`. I kept the call-on-behalf entry that just landed and put the phone-approval-gate row beside it in the same style.

## 1. liveness_phrase derived nothing, so two runners placed two calls

True. It is a hole in my own last answer. I wrote that phrase mode needs no key material because the phrase is spoken on the call rather than printed on the request channel. The request channel was never the whole story. The caller speaks the phrase, so the phrase sits in the task text, so it sits in the payload digest that `idempotencyKey` hashes. Two runners inventing two phrases held two different provider keys for one attempt, so both created a call at the same handset.

`resolveCodeKey` no longer takes a binding and no longer returns null (`src/config.ts:296`). Either binding refuses a live run without `CALLE_APPROVAL_CODE_KEY` or `--code-key-file` (`src/config.ts:308`), so no branch is left that can reach the random fallback. `src/cli.ts:171` resolves the key for every live run. `deriveSecret` already produced the phrase from the same HMAC and the same rejection sampling as the code (`src/secret.ts:141` for the digits, `src/secret.ts:156` for the words), so the derivation itself did not change. What changed is that phrase mode now gets a key at all.

Tests: `test/config.test.ts:105` pins the refusal for a missing key, an empty env value and an empty file path. That is the one new assertion set that fails against the previous head, because that head returned null in phrase mode. `test/gate.e2e.test.ts:299` runs the gate twice in phrase mode with separate state directories and one key, then asserts one call was created and both runs recorded the same secret digest plus the same call id. It passes on the old source when a key is handed to `runGate` directly, which is the point: the gate could always derive a phrase, the CLI never asked it to.

README, `docs/threat-model.md` and the CLI help no longer claim phrase mode needs no key.

## 2. decidedInWindow accepted a missing or malformed completedAt

True. It returned true for a null timestamp and true again for an unparseable one, on the reasoning that the local clock had already agreed. A call replayed under an old idempotency key is exactly the case where the local clock cannot help, so a decision from an older window could open a fresh one.

`completionTime` (`src/decide.ts:73`) is now the only place that decides whether a provider timestamp can be used. Not a string, empty, whitespace only, unparseable, non-finite: all null. `decidedInWindow` refuses on null before it looks at anything else (`src/gate.ts:97`). The local clock check and the minute of provider skew are unchanged.

The refusal carries its own reason so a record does not read as an expired window: `completion_time_unknown` (`src/types.ts:34`), with `completion_time_usable` in the evidence (`src/types.ts:187`), checked ahead of the window in `attemptOutcome` (`src/decide.ts:142`). `evaluateAttempt` reads the same field the window check reads and ties the two together, so `within_window` can never be true on a record whose completion time could not be read (`src/decide.ts:293`).

Tests: `test/gate.test.ts:90` is a table over undefined, null, empty string, whitespace, `"not a date"`, NaN, a number, an object, an array and true. Every one has to fail closed while a real timestamp inside the window still passes. `test/decide.test.ts:193` pins the recorded reason and evidence for a completed call with no usable time. `test/decide.test.ts:372` pins the same split at `attemptOutcome`, which is the function audit verification replays. All three fail against the previous head.

One thing worth flagging in the diff: the old window test asserted that a null timestamp and `"not a date"` both counted as inside the window. Those two assertions pinned the behaviour you flagged, so they are gone and the table replaces them.

`examples/audit.example.jsonl` is regenerated from the demo, so it carries both evidence fields and the `secret_not_returned` reason from the last round.

## Verified locally on b3888ab

- `npx tsc --noEmit`: clean
- `npm test`: 123 pass, 0 fail, 0 skipped. 120 before this round: three new tests, plus rewritten assertions in three existing ones
- `npm run demo`: four cases against the local fake, chain and verdicts hold, the forged copy still fails verification
- `python3 scripts/validate_repository.py`: passed, after the merge, since the index files are what it reads
- `git diff --check`: clean

No credentials, no real call. The suite talks to the local fake CALL-E server and to stub ports only.

Nothing left open from this round.

PR #42, opened 2026-07-29
feat(apps): add multi-party scheduler with two-phase booking over calls

zkasuran:
## Summary

Adds `apps/typescript/multi-party-scheduler`, a runnable app that books one
appointment across several people who only answer the phone, as a two-phase
protocol over CALL-E calls.

The problem: a repair that needs a tenant, a plumber and a building
superintendent in one place. Nobody shares a calendar, so somebody sits with a
phone and works down the list. The failure is always the same shape. Two people
said yes, the third could not make it and the first two were never told.

A single pass cannot fix that. Saying "you are booked for Thursday" while asking
leaves a false booking behind when the third person refuses. Never committing
means calling everybody again. So this app gathers availability without promising
anything, chooses one time, then commits in a separate round that can be rolled
back.

What the app does:

- **Gather.** One call per party, in request order. The script says plainly that
  nothing is booked. Each answer intersects the feasible set, so the next party
  hears a shorter list and an impossible schedule is found on the call that makes
  it impossible. Option numbers are stable across every call and in the ledger.
- **Confirm.** The earliest surviving slot is named to every party, one time, one
  word back. A yes only counts when it comes after the confirmation question was
  asked on that call, so an ordinary "yes, speaking" cannot commit anybody. The
  outcome is `verbally_confirmed`, never `booked`: this app writes to no calendar
  and has no booking adapter.
- **Recovery.** Every call is on disk before it is placed: the exact idempotency key,
  the attempt it belongs to, a digest of the call payload and the provider origin and
  account it went to, all before the create, then the accepted call
  id before anything waits on the call, so a process death in that window still leaves
  a call that can be found. `resume` settles a call whose create response was lost by
  re-issuing the key the ledger recorded for it, then places the release calls a
  crashed run owed. The key is read rather than derived a second time, so an upgrade
  between the crash and the resume cannot turn one call into two. Re-issuing it is
  refused unless the payload digest and the provider still match, because a key the
  provider never received places whatever body is sent under it. A call with no
  recorded key at all or a key with no attempt record behind it is refused rather
  than dialled. A retry of the one call this app places twice gets an attempt number of
  its own, but only once the previous attempt's outcome is settled and `--retry-release`
  asked for it. Nothing else mints a new key: a coordination lives in one ledger and
  runs once, so `run` pointed at a ledger that already records it hands back a finished
  round with no call placed and sends anything still open to `resume`, rather than
  deriving a fresh key for every call in that round. The resume digest is taken over the
  request whole, so an edited request is refused rather than dialled.
- **Release.** If any party does not confirm, nothing is booked and everybody who
  already confirmed gets a call telling them so, most recent first. Release calls
  ignore the coordination window on purpose, because telling somebody their
  afternoon is free again is a duty and not part of the booking. A party the
  release call cannot reach is reported in `unreleased` for a human to chase.
- **Budget and window.** Worst case is `2n + (n - 1)` calls. The request is
  refused at load time when `policy.max_calls` is below that, so a run never
  starts a protocol it cannot finish.
- **Ledger and replay.** Every call, every narrowing, the chosen slot and every
  release is one JSON line. `replay` recomputes the feasible set from the recorded
  answers, the slot that choice implies and whether the outcome follows. A ledger
  that says Thursday when the answers do not intersect on Thursday fails, which a
  plain log cannot catch. It also reports a call that was attempted with no result
  behind it, which is what a crash between the create and the answer leaves.

Two reading rules, both conservative in the same direction. Availability is
credited only when CALL-E's extracted `available_options` and this app's own read
of the transcript agree, so a mishearing cannot invent a free slot. A
disagreement is recorded rather than hidden. A confirmation has to come from the
transcript: the extracted answer can veto a confirmation but never create one.
When CALL-E returns no schema-valid result at all, which the API documents as a
real outcome, the transcript read is used alone and the ledger says so.

Timezones are never inferred. `meeting.timezone` is a required IANA name, slots
are full ISO 8601 instants with an offset. A request whose offset disagrees
with that zone at that instant is refused rather than read out wrong on a call.

Side effects and safety: at most one call per party per phase, nothing recurring,
so there is no schedule to clean up. `plan` and `replay` place no calls and need
no credentials. `CALLE_API_KEY` comes from the environment only. Numbers are
masked in output and in the ledger, only the decisive turns are kept and every
script refuses to negotiate a new time or accept any other instruction given on
the call.

## Verification

```text
npm test        151 tests pass
npm run demo    4 coordinations, the availability grid, replay, a tampered ledger
                that replay rejects, plus a run killed mid-commit and recovered
```

CALL-E in `fake/calle-server.ts`, which keys its scripts by phone and phase and
speaks the documented wire contract (201 on create, `Idempotency-Key` reuse,
snake_case payloads, stable error envelopes). The committed
`examples/ledger.example.jsonl` is unedited output from `npm run demo` and
replays cleanly.

Note for review: this is a separate contribution from my other open PR, #41. They
share no files. Both wrap the CALL-E SDK behind a small port, so if you would
rather have one shared adapter under `apps/shared/`, say so and I will factor it
out in whichever lands second.

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist


Ray-56:
Thanks for the thorough protocol work. The current head still has blockers:

- Confirmation parsing scans every user turn, so an ordinary pre-question `Yes, speaking` can become the confirmation signal. Bind confirmation to the exact confirmation question and a post-question answer.
- A lost create response or crash during commit can leave a party believing the appointment is confirmed without a mandatory release. The ledger only replays; add durable resume/recovery that reconciles ambiguous calls and completes owed releases.
- Idempotency keys bind only request/phase/party identifiers, not canonical content, and the optional ledger is not an atomic concurrency reservation.
- `plan --json` prints full phone numbers, required medical/legal/financial/emergency boundaries are absent, and there is no enforceable per-party consent/calling-hours or in-flight cancellation policy.
- Arbitrary `--base-url`/`CALLE_BASE_URL` receives the API key without HTTPS/trusted-host validation. The app also reports `booked` without creating a booking in any system; label this as verbal confirmation unless a real booking adapter succeeds.

`diff --check` and repository validation pass, but no lockfile or CI checks are present and typecheck/tests/fake-provider recovery could not be completed in this restricted environment. No real call was placed.

zkasuran:
Thanks, all five were real. Here is what changed, in your order.

## 1. Confirmation parsing scanned every user turn

True. `src/read.ts:215` walked every `user` turn and `CONFIRM_PATTERNS` includes
`/\byes\b/` (`src/read.ts:51`), so "Yes, speaking" set the answer to confirm and
`evaluateCommit` credited it.

Fixed in `src/read.ts`. `readConfirm` now finds the turn where the caller asked the
confirmation question (`CONFIRM_QUESTION_PATTERNS`, matching the lines
`confirmTask` actually reads out) and only a `user` turn after that index can
confirm. If the question never appears in the transcript, no confirmation is
possible. A decline still counts wherever it is said, because it can only stop a
commitment, never create one. The clause and negation handling is unchanged.
`CommitReading.questionAsked` is recorded on every commit result as
`question_asked`, so the ledger shows whether the call ever asked.

The fake server now speaks the real scripted lines per phase
(`fake/calle-server.ts`), so a confirm transcript contains the question and the
tests exercise the binding rather than a shortcut.

Tests: "a yes before the confirmation question is not a confirmation" and "a yes in
a call that never asked the question is not a confirmation" in `test/read.test.ts`.
Both fail against the old reader.

## 2. A lost create response or a crash could leave a party owed nothing

True. The sharpest case is not the crash. In the old code a confirm call whose
create response was lost was recorded as not confirmed (`src/coordinate.ts:414`),
so that party was never added to `confirmedParties` and the release loop
(`src/coordinate.ts:443`) skipped them, while the call itself may have gone ahead
and got a yes. Nothing was ever going to tell that person. `replay` could prove it
afterwards (`src/ledger.ts:74`) and could not fix it.

New `src/resume.ts`, wired as `resume --request <file> --ledger <file> --live`:

- it refuses a ledger whose `request_digest` does not match the request, because
  the request is what rebuilds the same idempotency keys
- `inspectLedger` folds the ledger into confirmed parties, released parties, owed
  releases and unsettled calls. A call is unsettled when it has no call id (the
  create response was lost) or a non terminal status (it was still running)
- a call with an id is settled by waiting on that id, which places nothing. A call
  with no id is settled by re-issuing the same idempotency key: CALL-E answers with
  the call it already has or places the one the run owed. That one is charged to
  the call budget, because from outside the two cannot be told apart
- it then places the release calls that are owed, most recent yes first, through
  the same `releaseRound` a fresh run uses
- it appends `resume_started` and `reconcile` entries and closes with a fresh
  `outcome`. `replay` now folds multiple rounds and rejects entries that follow an
  outcome without a `resume_started` opening a new round

It never gathers availability again, never chooses a different slot and once any
release call has gone out the appointment stays off, so a yes found late cannot
bring it back. A confirm call for a slot that has already started is not placed.
Anything it cannot settle is named in the outcome note.

Release stays exempt from the coordination window and stays subject to the call
budget, as before. It is now also subject to calling hours and a party who cannot
be called yet is reported in `unreleased`.

Tests in `test/resume.test.ts`, over the real client against the fake server: a run
killed between the yes and the release call and the lost create response case
where the reconciliation finds the same call by the same key and that party is then
released. `npm run demo` section 4 shows the same thing end to end.

## 3. Keys bound identifiers only and the ledger is not a reservation

True on both counts. The second one deserves a precise answer.

Content binding: `idempotencyKey` (`src/script.ts:212`) is now
`mps-<request_id>-<phase>-<party>[-<slot>]-<digest>`, where the digest is the first
12 hex characters of sha256 over the canonical JSON of the payload that determines
the call (task, result schema, recipient, metadata). It reuses `canonicalJson` and
`digestOf` from `src/ledger.ts` rather than a second implementation, so a key and a
request digest agree on what canonical means. `coordinate.callInput` builds the body
once and the key is taken from that exact object, so the key and the request cannot
drift apart and `resume` rebuilds both.

Concurrency: the reservation that stops a second dial is the idempotency key at
CALL-E, not the ledger. The ledger is optional, so it can never be that. What the
ledger can do is stop two processes interleaving lines into one history, so
`acquireLedgerLock` creates `<ledger>.lock` with `O_EXCL` and holds it for the whole
run and for the whole of resume. It is taken before the first dial. Two runs against
the same ledger now fail fast with the holding pid instead of writing a history that
replays as nonsense.

Tests: "idempotency keys are stable, specific to phase and slot and bound to the
content" in `test/script.test.ts`, plus "one writer per ledger and the lock is taken
before any call" and "a run refuses a ledger another run holds, before it dials
anybody" in `test/ledger.test.ts`.

## 4. Full numbers in `plan --json`, missing boundaries, no consent, hours or cancellation

All true.

`plan --json` printed the parsed request verbatim (`src/cli.ts:92`). There is now
`redactRequest` in `src/format.ts` and the CLI prints that, so JSON masks numbers
the same way the ledger and the plan text do. Test in `test/format.test.ts`.

Boundaries: `boundaryRules()` in `src/script.ts` is appended to all three scripts.
No medical, legal or financial advice and no opinion on any of them. Anybody who
says there is an emergency, that somebody is hurt or that a fire, gas leak or flood
is happening now is told to hang up and call their local emergency number and the
call ends. No payment, card, bank or government id details and one that is offered
is not repeated back. Test asserts all four in every phase.

Consent and calling hours are enforced in code, not prose. `parties[].consent_recorded`
must be `true` or the request is refused before any call (`src/config.ts`).
`parties[].calling_hours` is `{start, end, timezone}`, parsed in `src/hours.ts` and
checked in that zone by the coordinator before every call, release calls included.
The default is 09:00 to 20:00 in the meeting timezone, which is a floor rather than
an inference and a window that wraps past midnight is refused rather than guessed.
A call the window refuses is not placed and costs nothing from the budget. Tests in
`test/hours.test.ts`, `test/config.test.ts` and an end to end one that proves the
tenant is not dialled at 22:00 local.

Cancellation: `runCoordination` takes an `AbortSignal` and the CLI wires SIGINT and
SIGTERM to it. Once it fires, no new gather or confirm call is placed and the
release round still runs, because canceling the appointment does not cancel the
duty to tell anybody who said yes. A second interrupt gives up on that too. The
Developer API has no cancel endpoint, so a call already connected cannot be hung up:
the run records it with the status it had, which makes it unsettled and `resume`
settles it. Test: "canceling in flight stops new calls and still releases everyone
who said yes".

## 5. Untrusted base URL and `booked` without a booking

Both true. `src/cli.ts:130` passed any `--base-url` or `CALLE_BASE_URL` straight
into `createSdkPort` (`src/calle.ts:41`), which put the key in an `Authorization`
header against that host.

`assertTrustedBaseUrl` in `src/calle.ts` now runs before the client is constructed,
so it runs before any request that carries the key. It requires `https:` and allows
`http:` only for `localhost`, `127.0.0.1` and `::1`, which is what the local fake
server and the demo use. Anything else is refused with a message naming both
`--base-url` and `CALLE_BASE_URL` that says nothing was sent. It refuses rather than
warns. Tests in `test/calle.test.ts`.

On the label: you are right that nothing here creates a booking. The outcome is now
`verbally_confirmed` and `booked_with` is `confirmed_with`. Renamed together across
`src/types.ts`, the coordinator, the ledger and replay, the result renderer, the
exit code table, the README, the protocol doc, both app indexes, the package
description, the example ledger and the tests, so no document contradicts another.
`renderResult` prints "every party said yes on a call. Nothing is booked in any
system." and the README says the app writes to no calendar and has no booking
adapter. There is no adapter in this change: if one is ever added, the outcome
should only become something stronger when that adapter succeeds.

## Your environment note

`npm test` no longer needs the `tsx` CLI. The script is now
`node --import tsx --test test/*.test.ts`, which registers the loader in process
instead of going through `tsx/dist/cli.mjs`, the part that opens an IPC server. Same
82 tests, no socket.

No lockfile. No Node app in this repo commits one (only the Python apps commit
`uv.lock`), so adding a `package-lock.json` here would be the odd one out. Say the
word and I will add it or add it repo wide in a separate PR. Same for CI: the only
workflow is `validate.yml` running `scripts/validate_repository.py` and no app has
its own job, so I have not added one for this app alone. Happy to add a small
typecheck and test job for `apps/typescript/**` if you want it.

## Verification

Run from `apps/typescript/multi-party-scheduler` on Node v22.22.2 and from the repo
root for the validator. No credentials, no network beyond localhost, no real call
placed.

```
npx tsc --noEmit                      exit 0, no output
npm test                              82 tests, 82 pass, 0 fail (was 57)
npm run demo                          4 runs against the local fake CALL-E, then 2 replays
python3 scripts/validate_repository.py Repository validation passed.
git diff --check                      clean
```

`npm run demo` now includes a run that is killed between the yes and the release
call: the crashed ledger fails replay with "no outcome entry, the run did not
finish", `resume` places the two owed release calls and the ledger then replays
clean at 11 entries.

## Still open

Nothing from the review is left and the two things I deliberately did not do are
the lockfile and the per app CI job, both for the repo consistency reason above.

Ray-56:

- Live run still allows --ledger to be omitted. With ledgerPath null every recovery entry is discarded, so a crash or second interrupt after a verbal confirmation leaves resume no way to reconcile the call or deliver required releases. Require durable state before the first live call.
- Recovery treats every terminal release as delivered. A failed or canceled, unacknowledged release is added to released, so inspectLedger can erase an explicit unreleased debt and resume can report that nothing remains. Only acknowledged delivery should clear that debt, and recorded outcome.unreleased must be preserved.
- Confirmation is checked against the window only before placing the call, not after the result returns. A late final response can still produce verbally_confirmed.
- The base URL guard accepts any HTTPS host, so it validates transport encryption rather than host trust; CALLE_API_KEY can still be sent to an arbitrary configured HTTPS host. Existing ledgers also are not chmodded to the documented 0600 mode.

Typecheck, 82 tests, the local fake demo, repository validation, and diff check pass. No credentials or real calls were used.

zkasuran:
Thanks. All four were real. Branch head `dabf506`.

**1. A live run still allowed `--ledger` to be omitted.**

Refused in two places now, because the CLI was not the only way in. The CLI
refuses before the key is read (`src/cli.ts:120`). `assertDurableState`
(`src/coordinate.ts:111`) refuses inside the engine at `src/coordinate.ts:498`,
before the ledger lock is taken and before the first dial, so a library caller
holding a live port cannot dial with no durable state either. The in memory path
stays for unit tests and for `plan`, which places no call.

Tests: `test/cli.test.ts:42` (a live run with no ledger is refused before anything
is dialled), `:50`, `:57`, `:63`.

**2. Recovery treated every terminal release as delivered.**

You were right and the fix is in three places.

A release settles only when a person acknowledged it. `replay`
(`src/ledger.ts:323`, `:329`) counts a release as telling somebody only on
`acknowledged`. `inspectLedger` (`src/resume.ts:136`) does the same: a
terminal call that reached a machine, failed or was canceled leaves the party
owed.

A recorded `outcome.unreleased` is folded back in as debt (`src/resume.ts:150`)
rather than read as a settled fact. That was the erasure you named. The debt only
grows: nothing later in the ledger can remove a party from it except an
acknowledged release call.

`replay` also fails a history that confirmed somebody and then neither released
them nor listed them as unreleased (`src/ledger.ts:353`), so the gap is caught by
verification rather than only by recovery.

Tests: `test/resume.test.ts:236` (a terminal release that reached nobody leaves the
debt owed), `:268` (resume finishes a release nobody acknowledged instead of
writing it off), `test/ledger.test.ts:185`, `:154`.

**3. The window was only checked before the call.**

New `src/window.ts`. `decidedInWindow` (`src/window.ts:43`) reads two clocks
because they catch different things. This run's clock catches an answer that landed
after the round could act on it. The call's own `completed_at` catches a call that
finished outside this window at all, which is what a replayed idempotency key hands
back. The provider timestamp gets one minute of skew at each end
(`src/window.ts:20`), because it is somebody else's clock. Every confirm result
records `within_window` (`src/coordinate.ts:413`), a false one cannot be a
confirmation and the run ends as `window_expired`.

One decision inside this that is worth stating plainly: a late yes still owes a
release call. The person said yes on a phone call. From where they are sitting the
time is on. This app refusing to act on it is exactly what the release round is
for. So the release list is everybody who said yes on a call (`saidYes`,
`src/window.ts:70`), which also covers a yes the confidence floor discards and a
yes the extracted answer contradicts. `replay` holds a terminal outcome to that
same list. `resume` inherits the coordination window from the `run_started` entry
instead of opening a fresh one. It will not ask anybody new to commit once that
window has closed, because that is a call nobody should get.

Tests: `test/window.test.ts:21`, `:25`, `:30`, `:37`, `:44`,
`test/coordinate.e2e.test.ts:367` (an answer from outside this run's window
confirms nothing), `:414`, `test/ledger.test.ts:223`.

**4. The base URL guard checked the scheme, not the host. Ledger mode.**

`trustedHosts` (`src/calle.ts:77`) and `assertTrustedBaseUrl`
(`src/calle.ts:113`): the trusted set is `api.heycall-e.com` plus `localhost`,
`127.0.0.1` and `::1`, compared as exact lowercased hostnames with no suffix match
and no wildcard. Plain http reaches this machine only, which is what the fake
server and the demo use. Another host has to be named in `CALLE_ALLOWED_HOSTS` or
with `--allow-host`, which is repeatable. An opted in host still has to be
https. The refusal still happens before the client that carries the key exists
(`src/calle.ts:138`).

`appendEntry` (`src/ledger.ts:78`) opens the ledger and fchmods the descriptor to
0600 on every append. You are right that the old code only passed a mode to the
create, so an existing 0644 ledger kept it. fchmod on the descriptor is the same
file the write goes to, so there is no path left to swap in between. The lock file
gets the same treatment and a target that is not a regular file is refused rather
than written to.

Tests: `test/calle.test.ts:38` (a host that ends in a trusted name is refused),
`:51`, `:71`, `:82`, `test/ledger.test.ts:251`, `:263`, `:270`.

**Two things our own pass found that you did not list.**

A ledger torn by a crash mid append. `readLedger` (`src/ledger.ts:159`) names a
broken line with its line number instead of throwing a raw `SyntaxError`. The last
line is the exception: a crash between the write and the newline leaves half an
entry there, so replay reads the history without it and `repairTornTail`
(`src/ledger.ts:195`) drops it under the ledger lock in `resume` and says so in the
note. Leaving it would put a broken line mid history the moment anything else was
appended. Tests: `test/ledger.test.ts:281`, `:293`.

A correction to one of my own commit messages. `49c353d` says the terminal status
set "names the six the API can end on". That was wrong and a later commit on the
branch fixed the code: `CallStatus` in the schema is `queued`, `in_progress`,
`completed`, `failed`, `canceled`, so `TERMINAL_STATUSES`
(`src/coordinate.ts:98`) holds `completed`, `failed` and `canceled` only. A no
answer or a voicemail arrives as `failed` with a failure code, which is how the
reader already handles it. Flagging it because the commit body is still on the
branch and it claims an API surface that does not exist.

**Docs.** The README described the old base URL check and promised 0600 without
saying it is re-applied on every append. Both match the code now. `docs/protocol.md`
gained the window rule: both clocks, the minute of skew and the decision that a
late yes still owes its release call.

**Verification** (Node 22.22.2, from `apps/typescript/multi-party-scheduler`, repo
root for the validator):

```text
npm test                                107 tests, 107 pass, 0 fail, 0 skipped (82 before)
npm run demo                            exit 0, runs against the local fake, then replays
```

**Still open.** Nothing from your list. The two things I deliberately have not
done are the same two as last round, the npm lockfile and a per app CI job, both
because no app in this repo has either. Say the word on either and it is a small
change.

No real call was placed. Everything above is against the local fake CALL-E over
the real `@call-e/calle` client, with no credentials and nothing beyond localhost.

Ray-56:
Thanks for the update. Two ambiguous-outcome blockers remain:

- decidedInWindow accepts missing or malformed completedAt, so a stale idempotent replay can satisfy a new coordination window. Missing provider completion time must fail closed.
- placeCall converts ambiguous create, wait, and read failures into ordinary api_error or a normal nonterminal result. A call can remain active while the workflow continues to another party. Preserve these outcomes as unresolved, stop the round, and reconcile the same idempotency key before any further call.

zkasuran:
Both round-3 blockers are fixed at 048fadd. `main` is merged into the branch as well: after #43 landed, the only conflict was the app index rows in `README.md` and `apps/README.md`, resolved by keeping the call-on-behalf row that just merged and putting the multi-party-scheduler row beneath it in the same style.

## Live run still allows a missing or malformed completion time

You were right. It was worse than the parse case alone. Against the pre-fix source every one of these returned `within_window=true` when the local clock was inside the window: missing, `null`, `""`, `"   "`, `"some time on Thursday"`, `NaN`, `Infinity`, a number of milliseconds, a `Date` and an object. A stale idempotent replay carrying no usable timestamp satisfied any window it was measured against.

`judgeWindow` (`src/window.ts:58`) now returns a verdict rather than a bare boolean and fails closed on the timestamp. A `completed_at` is refused unless it is a non-empty string that parses to a finite instant (`src/window.ts:70` and `src/window.ts:74`). The field is typed `unknown` on the way in, because it is the provider's field and this app has to check it rather than trust its declared type. The local clock check and the minute of provider skew are unchanged. `decidedInWindow` stays as the boolean wrapper (`src/window.ts:82`).

Each refusal now carries its own reason: `no_window`, `late_result`, `no_completion_time` or `outside_window`. It is recorded as `window_reason` on every commit and reconcile entry, so a line that failed for want of a timestamp no longer reads like a plain expired window. A yes that lands late still earns its release call: `saidYes` is untouched and the release list is still built from what the person said.

Pinned by `test/window.test.ts`: a table over the six shapes you named plus `Infinity`, a `Date` and an object, asserting `within` false and `no_completion_time` for each. A second test checks that every refusal reports its own reason. One old assertion said the opposite ("no usable completion time leaves the local clock to decide"). It pinned the hole, so it is gone.

## placeCall converts ambiguous failures into ordinary results

Also true. `placeCall` wrapped create, wait and read in one try block and returned `{ call: null, errorCode }` for everything, so a create with no reply looked like a call that never happened and a failed read looked like an API error. The sharpest edge was in the confirm phase: an unresolved confirm ended the round as `not_confirmed` and then the release round rang everybody who had already said yes, while a call that might still agree the time was live.

Split by whether the server chose to answer. `CalleCallError` now carries the status and an `ambiguous` flag for a null status, 408, 409, 429 and 5xx (`src/calle.ts:47`, `src/calle.ts:53`). The SDK adapter passes `CalleAPIError.status` through, so a connection error with no status is ambiguous.

- `placeCall` (`src/coordinate.ts:200`) re-issues the same idempotency key on an ambiguous create. That is the one request which cannot ring a second time, so CALL-E hands back the call it already holds for the key. A definite refusal returns straight away with no second attempt.
- A wait failure reads the call rather than guessing. A read that also fails comes back `unresolved` with the call id kept. So does a call CALL-E has not finished with.
- The recorded status is `unresolved` (`src/coordinate.ts:301`), never `api_error` and never a raw nonterminal status. The last status CALL-E reported is kept in `notes` and the error chain in `failure_code`.
- The gather round stops before the next party and does not narrow the feasible set off a call nobody can account for (`src/coordinate.ts:700`). The confirm round stops and skips the release round (`src/coordinate.ts:781`, `src/coordinate.ts:830`), recording everybody who said yes in `unreleased` so the debt survives into the ledger.
- The terminal set and the `unresolved` literal live in `src/call-state.ts` now, imported by the coordinator, recovery and replay, so the three cannot drift.
- Recovery owns exactly those calls. `settleExisting` classifies the same way (`src/resume.ts:190`). While an unresolved confirm stands, `resume` decides nothing and places no release call either (`src/resume.ts:343`, `src/resume.ts:404`, `src/resume.ts:430`). A debt still clears only on acknowledged delivery, so an unresolved release cannot write one off.
- `replay` expects the feasible set to be unchanged across an unresolved gather entry (`src/ledger.ts:288`).

Pinned by `test/unresolved.test.ts`, seven cases over a stub port that can drop a reply, answer a create twice or hold a call open. Six fail against the pre-fix source. The seventh is the control: a 402 is a definite refusal and must still be read as one.

Two existing tests changed their expectations, because those expectations were the behaviour you asked to change. The lost create response now stops the confirm round, so nobody is rung to say it is off. `resume` then finds the yes that call had all along and the appointment stands. The non-live stub port now reconciles once under the same key before it stops.

One deliberate exception. The release round keeps going after an unresolved release call. The appointment is already off by then and each release call is a duty to a different person, so stopping would leave somebody who agreed to a time believing it is going ahead. The unresolved call stays listed in `unreleased` and `resume` retries it.

## Verification

From `apps/typescript/multi-party-scheduler`, after the merge:

- `npx tsc --noEmit` clean
- `npm test` 115 pass, 0 fail, 0 skipped, up from the 107 baseline. Blocker 1 added 2, blocker 2 added 7 in a new file, one stale window assertion was replaced
- `npm run demo` runs all four scenarios against the local fake CALL-E and the crash then resume replays clean at 11 entries
- `git diff --check` clean

No credentials and no real calls were used. `examples/ledger.example.jsonl` is regenerated from the demo run so it carries the new fields.

Nothing left open on either blocker.

zkasuran:
One follow up at `5824db8`, so this matches the shape #41 landed rather than inventing a second one.

The refusal is now named `completion_time_unknown` and it is decided before the window and before the local clock (`src/window.ts:58`). The evidence field `completion_time_usable` sits next to the verdict on every commit and reconcile entry, so a ledger line can never claim the window was weighed against a time nobody could read. Reading the timestamp last, which is what I posted above, let `late_result` hide the same failure. The minute of provider skew and the local clock check are still unchanged.

Verified again from `apps/typescript/multi-party-scheduler`: `npx tsc --noEmit` clean, `npm test` 116 pass 0 fail 0 skipped, `npm run demo` exit 0, `python3 scripts/validate_repository.py` passes, `git diff --check` clean. `examples/ledger.example.jsonl` is regenerated from that demo run.

Ray-56:
One ambiguous-create blocker remains on the current head:

After the first create attempt has an ambiguous outcome, a second same-key create that returns a definite 4xx is treated as proof that no call exists. That is not a valid reconciliation: the first request may already have been accepted, while the second 401/403 can fail before idempotency lookup. coordinate.ts then returns unresolved: false and the round may call another party while the first call is live.

Once the first create outcome is ambiguous, only successfully retrieving the original call can resolve it. Any failed reconciliation, regardless of the second error class, must remain unresolved and stop the round.

zkasuran:
Fixed at `a86dcad`. You were right, and the hole was exactly where you said.

`placeCall` reconciled an ambiguous create by re-issuing the same idempotency key, then branched on the class of the second answer: an ambiguous second failure stayed `unresolved`, and a definite one returned `{ unresolved: false, errorCode }` with the comment "the second answer is definite, so there is no call after all". That is a wrong inference twice over. The first request may already have been accepted, and a 401 or a 403 on the reconciliation can be decided before the idempotency lookup happens at all, so it carries no information about the request that went unanswered. `coordinate.ts` then treated the party as simply not reached, and the gather round moved to the next party while the first call could be ringing.

The rule is now the one you stated. Once a create is ambiguous, getting the call back is the only thing that resolves it. Any second failure, whatever its class, returns `unresolved: true` with the chain `<first>, then <second>` in `failure_code`, so the round stops before the next party, the feasible set is not narrowed and the debt stays in the ledger. Nothing else about the split changed: a definite refusal on the **first** attempt still means no call exists and the round still reads it as a refusal, which is the control case in the test file.

Re-issuing the same key stays the retrieval mechanism rather than a second dial. A lost create response never named a call id, so there is nothing to `GET`, and the key is the reservation, so CALL-E answers with the call it already holds. What changed is only how a failure of that retrieval is read.

`resume` inherits the rule, since its lost-create settlement goes through the same `placeCall`. Its other path, `settleExisting` for a call that does have an id, already stayed unresolved when the read failed.

Pinned by a new case in `test/unresolved.test.ts`, "a definite refusal on the second attempt resolves nothing": an unanswered create followed by 401, 403, 400 and 402 in turn. Each asserts the outcome is `unresolved`, that exactly two creates happened and both were for the same party, that nobody else was called, that the ledger entry is `unresolved` with no call id and the full error chain, that the feasible set is unchanged and that replay still passes. It fails against the pre-fix source for all four statuses and the other seven cases in the file still pass. The README and `docs/protocol.md` sections on accounting for a call now state the rule instead of "only when that fails too".

Verified from `apps/typescript/multi-party-scheduler`: `npx tsc --noEmit` clean, `npm test` 117 pass 0 fail 0 skipped (up from 116), `npm run demo` exit 0 with all six sections, `python3 scripts/validate_repository.py` passes, `git diff --check` clean. No credentials and no real calls. `examples/ledger.example.jsonl` is unchanged, because this touches a failure path the demo does not enter and no ledger field changed.

zkasuran:
Both of last round's fixes are in the code rather than only in the commit message. I went looking for the same mistake in the rest of the app before saying so. It was there once more, in the place where it costs the most. Fixed at `768dd30`.

## The ambiguous create, checked in the source

`placeCall` (`src/coordinate.ts:205`) sorts the first create failure. Definite means no call exists, so it returns `unresolved: false` with no second attempt. Ambiguous re-issues the same key. The only branch that resolves it is the one where CALL-E hands the call back. Any second failure returns `unresolved: true` with the chain `<first>, then <second>`, whatever its class, so the round stops before the next party and the debt stays in the ledger. `test/unresolved.test.ts:122` runs 401, 403, 400 and 402 as the second answer and asserts unresolved, exactly two creates both for the same party, nobody else called, a ledger entry with no call id and the full chain, an unmoved feasible set and a clean replay.

## The completion time, checked in the source

`judgeWindow` (`src/window.ts:75`) reads `completed_at` first, before the window is tested for computability and before the local clock is consulted. `test/window.test.ts:70` pins that order rather than the outcome: a null timestamp with `now` past the deadline has to report `completion_time_unknown` and not `late_result`. An unreadable one with `windowStart: NaN` has to report it and not `no_window`. A check that was merely present and last would satisfy the outcome assertions and fail these two.

## The same class, one step further in

`saidYes` decided who is owed a release call and it required `reached_person`, which requires `call.status === "completed"`. A confirm call CALL-E reports as `failed` or `canceled` can still carry the confirmation question and a yes after it: that is what a line dropping once somebody has agreed looks like. So the transcript said the tenant agreed to Thursday, the status said the call did not complete and the run believed the status. The tenant came off the release list, the run ended `not_confirmed` and the person who had cleared their afternoon was never going to be called. Replay could not catch it, because it reads the same rule off the same ledger.

Against the pre-fix source that run places one release call, to the plumber. It now places two, the tenant first, with `unreleased` empty. A duty is read from the transcript (`src/window.ts:117`): the yes has to come after the confirmation question with no machine on the line. No status decides it. Nothing loosened on the other side. `confirmed` still needs a completed call, both readings agreeing and the window.

Two more of the same shape came out of that pass.

**A refusal is recorded as the check that refused it.** Every uncredited yes ended the run as `window_expired` with a note saying the window closed, including a yes refused because the completion time could not be read. That is the invented reason `5824db8` took out of the ledger line, still sitting in the outcome. `window_expired` is now kept for `late_result` and `outside_window`. Everything else is `not_confirmed` with the reason named, `completion_time_unknown`, `low_confidence`, `extracted_decline` or `call_failed`. The release call goes out either way.

**A debt clears only on delivery the recording shows.** `acknowledged` was true whenever the extracted answer said yes, even with nothing acknowledged anywhere in the transcript, so an extraction could write off what this app owes a person. The transcript leads now, the same rule as a commitment. An explicit `no` in the extracted answer still vetoes it.

The asymmetry there is deliberate. A duty is created by what the person said, so it is read from the transcript and a provider status cannot cancel it. A duty is discharged only on positive evidence, so a release still needs a completed call, a person on the line and an acknowledgment.

`test/duty.test.ts` is that one rule from three angles, four cases, every one of them failing against the pre-fix source: a yes on a `failed` or a `canceled` call still earns its release call (`:88`), a completion time nobody can read is not reported as a window that closed (`:142`), a release nobody acknowledged stays owed (`:195`). The first two run end to end over the real client against the local fake server.

## Against the two that merged

The reconciliation rule matches `phone-approval-gate` as merged: its `src/gate.ts` also keeps a second failure unresolved whatever the class. `completion_time_unknown` is the same name in the same position in its `decide.ts` ladder. Two deliberate differences here. This app also treats a 429 as ambiguous, because a rate limit can sit on top of an accepted call. It also records the whole chain rather than the second code alone. `call-on-behalf` as merged still branches on the class of the second answer (`src/errand.ts:186`) and files a definite one as a call that was not created. That is one call with no release duty, so the consequence is smaller, but it is the same inference. Happy to send it as its own PR if you want it.

## Every round and where it stands

| Round | What you raised | Landed in | Test |
| --- | --- | --- | --- |
| 1 | Confirmation parsed from any user turn | `18526db` | `test/read.test.ts` "a yes before the confirmation question is not a confirmation" |
| 1 | A lost create response or crash leaves a party owed nothing | `18526db`, `4fe071c` | `test/resume.test.ts` "a confirm whose create response was lost stops the round, then resume finds the yes" |
| 1 | Keys bind identifiers only, the ledger is not a reservation | `18526db` | `test/script.test.ts` "idempotency keys are stable, specific to phase and slot and bound to the content", `test/ledger.test.ts` "one writer per ledger and the lock is taken before any call" |
| 1 | `plan --json` leaked numbers, no boundaries, consent, hours or cancellation | `18526db` | `test/format.test.ts`, `test/hours.test.ts`, `test/script.test.ts` "every call script refuses advice, emergencies and payment details" |
| 1 | Untrusted base URL took the key, `booked` without a booking | `18526db` | `test/calle.test.ts`, `test/format.test.ts` "a confirmed run is printed as a verbal confirmation, not a booking" |
| 2 | A live run still allowed `--ledger` to be omitted | `7da64ce`, `6d4e560` | `test/cli.test.ts` "a live run without --ledger is refused before anything is dialled" |
| 2 | Recovery read every terminal release as delivered | `49c353d` | `test/resume.test.ts` "a terminal release call that reached nobody leaves the debt owed", `test/ledger.test.ts` "a release call that reached nobody does not count as telling them" |
| 2 | The window was only checked before the call | `b17475d` | `test/window.test.ts`, `test/coordinate.e2e.test.ts` "a confirm answer that landed outside this run's window confirms nothing" |
| 2 | The base URL guard checked the scheme, ledger mode not re-applied | `beaa487` | `test/calle.test.ts` "nothing is suffix matched, so a host that ends in a trusted name is refused", `test/ledger.test.ts` "a ledger that already exists is put back to 0600 on the next append" |
| 3 | A missing or malformed `completed_at` satisfied the window | `0bf0c11`, renamed and reordered at `5824db8` | `test/window.test.ts:48` and `:70` |
| 3 | `placeCall` turned ambiguous failures into ordinary results | `e98b85f` | `test/unresolved.test.ts`, seven cases |
| 4 | A definite 4xx on the second create read as proof of no call | `a86dcad` | `test/unresolved.test.ts:122` |
| this pass | A call status read as proof that nobody said yes | `768dd30` | `test/duty.test.ts`, four cases |

## Verification

From `apps/typescript/multi-party-scheduler` on Node 22.22.2, repo root for the validator:

```text
npm test                                121 tests, 121 pass, 0 fail, 0 skipped (117 before)
npm run demo                            exit 0, all six sections
```

The upstream `validate` check is green on `768dd30`. No credentials and no real calls. `examples/ledger.example.jsonl` and `results/*.jsonl` are unchanged, because the demo does not enter any of these paths and no ledger field was added.

Nothing from your list is open. The two things I have deliberately not done are the same two as every round, the npm lockfile and a per app CI job, because no app in this repo has either. Either is a small change if you want it. This is the last of the three: #41 and #43 are merged, so it is ready for another look whenever you have time.

Ray-56:
A P1 resume-safety gap remains. `requestDigest` omits party phone, name, role, region, locale, and calling hours, but unresolved creates without a call ID are later reconstructed from the current request. Editing a party under the same request ID therefore passes resume validation, produces a different payload and idempotency key, and can dial another recipient while the original ambiguous call may still exist. Bind every call-affecting party and payload field into the durable digest, and refuse resume when any of them differs.

zkasuran:
Fixed at `151b39d`. Confirmed as you described it, and it is the failure the protocol exists to prevent, so it is worth spelling out.

`requestDigest` bound the meeting, the policy, slot ids and starts, and party **ids**. Everything else about a party sat outside it: name, phone, role, region, locale, recorded consent and calling hours. Since an unresolved create with no call id is rebuilt from the request in hand, editing a phone under the same request id passed the digest check, produced a different payload, took a different idempotency key with it, and placed a fresh call to a different recipient while the original ambiguous call might still have been live.

**Parties now go into the digest whole**, rather than field by field. That is the part I want to flag as a deliberate choice: an explicit field list would have fixed today's hole and left the same trap for whoever adds the next call-affecting field to `Party`. Binding the object means a new field is covered by default, and the failure mode of forgetting one becomes a refused resume instead of a call to the wrong person.

Slots gain `option`, because that is the number the callee hears, so slot order is part of the contract. `startMs` and `spoken` stay out on purpose: both are pure functions of `start`, `option` and the meeting timezone, and all three are bound already.

A test pins eleven edits that must each move the digest, including party order and each calling-hours field separately. Tests 121 -> 122, `tsc --noEmit` clean.

Ray-56:
The digest fix is still incomplete: `requestDigest()` omits `requestId`, while rebuilt idempotency keys and metadata include it. Editing only `request_id` therefore passes the resume check but changes the key, so reconciliation of an accepted-but-lost create can place a second call. Bind `requestId` in the stored digest and add a resume regression test for it.

zkasuran:
Fixed at `4a85518`. Confirmed exactly as you described it. Going looking for the
rest of that class turned up one more instance of it that the digest can never
close, so this push closes that a different way.

## 1. `requestId` was outside the digest

`requestDigest` bound the meeting, the policy, the slots and the parties. It did
not bind the id. The id is the first thing every key is built from
(`mps-<request_id>-<phase>-<party>[-<slot>]-<digest>`, `src/script.ts:249`) and it
also sits in the metadata, which is inside the payload the digest on the end of
that key is taken over, so editing only `request_id` moved the key twice while the
resume check passed. Reconciling an accepted-but-lost create would then have gone
out under a key CALL-E has never seen: a second call, to the same person, while the
first one may still be live.

**The digest is now taken over the request whole** (`src/ledger.ts:72`). That is the
part worth flagging as deliberate. An explicit list would have fixed `requestId` and
left the same trap for the next field. This is the second round where the omission
was the list rather than any one field, the party fields first and the id now. So
there is no list. Anything added to `CoordinationRequest` is bound the day it is
added and the cost of a mistake there is a refused resume, not a call to the wrong
person.

`startMs` and `spoken` are bound with it, which reverses what I told you last round.
The reason I gave then, that both are pure functions of `start`, `option` and the
meeting timezone, is wrong about `spoken`: it is rendered by `Intl.DateTimeFormat`
(`src/slots.ts:41`), so it comes from the runtime's locale data as much as from the
request. Binding it turns a resume that would read a different line out loud into a
refusal rather than a rebuilt key.

Tests. `test/ledger.test.ts` "every field of the request is bound into the digest"
is typed `Record<keyof CoordinationRequest, ...>`, so adding a request field fails
`tsc` there until somebody says how it moves the digest. It also compares that key
set against a real request, so a field that reaches the object without reaching the
type is caught. `test/resume.test.ts` "resume refuses a ledger when only the
request id was edited" is the regression test you asked for: same request in every
other respect, refused, nothing dialled. Against the previous head those two fail,
plus the two slot rows added to the party-field test, 3 failures of 28.

## 2. The same reconstruction was wrong for a reason no digest can cover

The derived key carries a digest of the payload and the payload carries the task
text, which lives in `src/script.ts` rather than in the request. So a run crashes on
an ambiguous create, the app is upgraded, one line of a call script changes, and
`resume` derives a key that does not match the one the lost create used and places a
second call. The request is untouched in that story, so no digest over the request
can catch it.

**The key is durable state now.** Every ledger entry for a call records the
`Idempotency-Key` it went out under (`idempotency_key` on `GatherResult` and
`CommitResult`, `src/types.ts:194` and `:242`) and `resume` re-issues that exact
string (`src/resume.ts:362` for a call with an id, `:406` for a create with none,
taken by `placeCall` at `src/coordinate.ts:227`). Nothing is derived twice. If the
body behind that key no longer matches, CALL-E answers `idempotency_conflict`, which
is ambiguous, so the call stays unresolved and the round stops instead of dialling.
Nobody is rung twice either way.

Tests. "resume re-issues the key the ledger recorded rather than deriving a new one"
stamps the ledger with a key nothing could derive and asserts that is the string on
the wire and in the `reconcile` entry. "a recorded key re-issued with a different
body stops rather than ringing again" drives the real client at the fake server,
takes the 409 and asserts one call created rather than two. Both fail when the
recorded key is ignored.

## Also in this push

- `main` merged. The only conflict was the app index row in `README.md` after
  freshchain-resolver landed in the same place, resolved the way the last one was:
  the row that just merged stays where it is and the multi-party-scheduler row goes
  beneath it. `apps/README.md` merged on its own.
- `apps/typescript/multi-party-scheduler/package-lock.json` went in by accident with
  last round's commit. Removed. No other app in the repo commits one.
- The README said 82 tests. It says 126.

Verified from `apps/typescript/multi-party-scheduler`: `npx tsc --noEmit` clean,
`npm test` 126 pass 0 fail 0 skipped, `npm run demo` exit 0,
`python3 scripts/validate_repository.py` passes, `git diff --check` clean.
`examples/ledger.example.jsonl` is regenerated from that demo run and replays
cleanly. No credentials and no real calls were used.

Ray-56:
The latest digest/key fix closes the edited-request gap, but a crash window still leaves an untracked call. placeCall sends createCall and waits before the caller appends the gather/commit ledger entry, so a process death after provider acceptance but before that append leaves neither the exact key nor an accepted call ID for resume. Persist a content-bound call-attempt record with the exact key before create, update it with the accepted ID before polling, and fail closed on legacy ambiguous entries whose key is missing instead of deriving a new one.

zkasuran:
Fixed at `e059d17`. Confirmed exactly as you describe it, including where the window
opens and closes.

## 1. The window

`placeCall` sent the create and waited. The `gather` or `commit` entry was appended
by its caller only once that wait came back (`src/coordinate.ts:230` at `4a85518`,
then `:775` and `:849` for the appends), so a process death after CALL-E accepted the
call and before that append left nothing on disk. No key, so there was nothing to
re-issue. No id, so there was nothing to read.

Nothing downstream could see it either, because everything downstream read the phase
entries. `inspectLedger` found no unsettled call. `resume` released the parties who
had said yes and closed the ledger with a fresh outcome. `replay` then passed: the
call count matched, everybody who said yes was released, ok=true. In the case I now
test, the missing call was the superintendent's confirm and it had said yes, so the
old path rang two people to tell them a time all three had agreed to was off, then
produced a clean history saying so.

## 2. Two lines, written by the code that places the call

`call_attempt` (`src/types.ts:290`) is appended before `port.createCall`
(`src/coordinate.ts:256`). It carries the phase, the party, the masked number, the
slot, the exact `Idempotency-Key` and `payload_digest`, a sha256 over the canonical
JSON of the same `callInput` payload the key is taken over (`src/ledger.ts:50` for
the digest, `src/script.ts:249` for the key). That is what content bound buys: the
payload is recomputable from the request plus this code, so the record can be shown
to belong to that call rather than to a party and a phase that could sit on a
different body. In the committed example the last 12 hex of the key are the first 12
of the digest on the same line, which is the same binding read twice:

```json
{"kind": "call_attempt", "at": "2026-08-04T17:00:00.000Z", "phase": "confirm", "party_id": "plumber", "phone_masked": "+14*******01", "slot_id": "thu-14", "idempotency_key": "mps-ash-lane-3b-leak-confirm-plumber-thu-14-38972b6c1481", "payload_digest": "sha256:38972b6c14810b7c83de6b4cc52902aef23b629318e85d0be5deee4b26797004"}
```

One record per `placeCall` rather than one per create. An ambiguous create re-issues
the same key with the same payload, so it is the same attempt and a second line for
it would only invite a reader to think two calls exist.

`call_accepted` (`src/types.ts:291`) is appended as soon as a create returns an id,
before anything waits on the call (`src/coordinate.ts:292`). It sits outside the
try/catch around the create on purpose: an append that throws in there would be
caught and read as an ambiguous create, which is a ledger problem reported as a
telephony one.

Neither line records an answer and neither is a call placed, so `calls_placed`, the
budget and the outcome still come from the phase entries. `PlaceOptions` gained
`record` and `now` for this (`src/coordinate.ts:180` and `:187`). The clock is
injected because the demo pins one and `examples/ledger.example.jsonl` is committed,
so a wall clock read in here would move a committed file on every run.

## 3. What recovery does with them

`openAttempts` (`src/ledger.ts:263`) is the one reader of that state, shared so
replay and recovery cannot drift apart on what is open. An attempt is open when no
entry carries its key and no entry records that party and phase at all. The second
half matters for a hand edited ledger: a phase entry under a different key is still
the result of that call, not a call nobody can account for.

- A confirm or release orphan becomes an unsettled call (`src/resume.ts:220`). There
  is no result to read, so the `CommitResult` the existing loop needs is synthesized
  from the attempt and says as much: `call_status` `unresolved`, the recorded key, the
  accepted id when there is one, `failure_code` `no_result_recorded`, everything else
  false or null (`src/resume.ts:101`). Nothing about what the call did is inferred,
  because nothing is known. With an id it is settled by a read, which places nothing.
  With no id the recorded key goes back out and CALL-E answers with the call it holds.
- A gather orphan is not recovery's to finish. `resume` never gathers again, which is
  what `nameOpenCall` already tells a person mid run (`src/coordinate.ts:790`). It
  goes into a new `RecoveryState.unsettledGathers` (`src/resume.ts:88`), then into
  `stuck` and the note (`src/resume.ts:437` and `:604`) and into
  `resume_started.ambiguous` as `gather:<party>`, so the open call is on the record
  even though this run leaves it alone.
- `replay` reports an open attempt by party and phase, with the accepted id when
  there is one (`src/ledger.ts:465`). A history holding one is not ok whatever its
  outcome line claims.

## 4. The fallback from last round is reversed

Last round put the key on the result entry and derived one when the entry had none
(`src/resume.ts:406` at `4a85518`). You are objecting to the derive and you are
right. That path is a legacy `commit` entry with no id and no key. Nothing on disk
says a derived string is the one the lost create used. It is refused now
(`src/resume.ts:456`): a progress line, the party in `stuck` and in the note, no
create. `key: recordedKey` at `src/resume.ts:514` is unconditional.

The trade is worth stating plainly, because refusing is not free. That call may still
be live and this run will never know. What the run does instead is discharge the duty
it can: the parties who did say yes get their release calls in the same pass, so
refusing costs a manual check rather than leaving somebody expecting an appointment.
Waiting for a call that can never be settled would cost the opposite.

## 5. Tests

131 now, 126 before. Every one was run against `4a85518` by restoring the four
touched `src` files from that commit and running the two test files that hold them:
28 pass, 5 fail, then restored.

| Test | Where | Against `4a85518` |
| --- | --- | --- |
| a crash between the create and the ledger append leaves the key resume settles that call with | `test/resume.test.ts` | fails: no `call_attempt` entry to cut the crash at |
| a crash after the accepted id was recorded settles that call without placing one | `test/resume.test.ts` | fails: no confirm attempt for superintendent |
| a gather call nothing settled is reported for a person, never dialled again | `test/resume.test.ts` | fails: no `call_attempt` entry to cut the crash at |
| an unsettled call with no key and no call id is refused rather than dialled | `test/resume.test.ts` | fails on behaviour: note is `resumed and settled, every party confirmed thu-14 by voice` |
| a call attempt nothing settles is reported | `test/ledger.test.ts` | fails: `placeCall` records no attempt before the create |

Being straight about that table: four of the five cannot fail on behaviour at
`4a85518`, because the record they read does not exist there. That absence is the
defect, so the failure is the right one, but it is a fixture failure and I am not
going to dress it up as more.

The fifth is behavioural. The old path derived a key for the keyless entry and
settled the call off it, so the resume reported everything confirmed. The derived key
matched the lost create there only because nothing in this repo changed between the
run and the resume. That match is the assumption the ledger cannot back, which is the
whole reason for failing closed.

How the crash states are built, since it decides what the tests prove: a real
coordination runs against the fake server through the real client, then the ledger is
cut at the line the process would have died on. The call is genuinely at the provider
under the key the attempt names. The same port answers the resume, so re-issuing that
key returns the call CALL-E already holds. The first test asserts one create on the
wire under the recorded key with no new call at the provider, the second asserts no
create at all because the accepted id is readable, `placed_call: false` and 5 calls
charged rather than 6.

## Also in this push

- The ledger gains two lines per call, so the demo entry counts moved. The README
  quotes a fresh `npm run demo`: `entry 17` on the crashed ledger and `25 entries`
  after resume, where it said 7 and 11. The confirmed run is 21 entries.
- `examples/ledger.example.jsonl` is regenerated from that run. 21 lines,
  `replay --ledger examples/ledger.example.jsonl` exit 0, no full number in it.
- `docs/protocol.md` documents both entry kinds, the refusal, the gather rule and the
  new replay check. The README's "When a run does not finish" and "The ledger and why
  replay matters" say the same. Its test count line is 131.

Verified from `apps/typescript/multi-party-scheduler`: `npx tsc --noEmit` clean,
`npm test` 131 pass 0 fail 0 skipped, `npm run demo` exit 0,
`python3 scripts/validate_repository.py` prints `Repository validation passed.`,
`git diff --check` clean. No credentials and no real call.

Ray-56:
The new pre-create attempt and accepted-ID checkpoints close the original crash window, but current recovery still has blockers:

- A release retry derives the same key and payload as the prior terminal unacknowledged release, so provider idempotency returns the same machine/no-answer call; the test counts a second ledger/API attempt, not a second provider-created call. Give each verified release retry a separately persisted attempt identity while retaining ambiguity reconciliation.
- payload_digest is recorded but never carried into recovery or compared with the rebuilt payload, so an upgrade can replay the old key with changed, unapproved task/schema when the first request never reached the provider.
- The ledger binds neither provider origin nor account, so resume under another base URL or API key can create the call in another idempotency namespace while the original remains ambiguous.
- If cancellation arrives during an ambiguous first create, placeCall still performs its create retry; when the first request never landed, that starts a new gather/confirm call after cancellation.

Fail closed on payload/provider mismatch, honor abort before the create retry, and make release retries durable and genuinely new only after the previous outcome is known.

zkasuran:
Fixed at `d178694`. All four were real and three of them are the same shape: a key
outlives the process that made it, so everything about it has to be on disk or it is
not checked at all. `main` is merged in as well, so the branch is clean again.

## 1. The release retry was not a retry

Confirmed, including why the test did not see it. Every part of that key is stable
across attempts by design, which is what makes it a reservation. So the retry derived
the string the first attempt had used, CALL-E answered with the call that reached the
machine and nobody was called again. A second `release` entry and a second charge
against the budget looked like a retry from the inside.

The key now ends `-a<attempt>` and the number comes from the ledger:

```
mps-<request_id>-<phase>-<party>[-<slot>]-<digest>-a<attempt>
```

`nextAttempt` in `ledger.ts` is one more than the highest `call_attempt` the file
holds for that party, phase and slot. It is read off disk rather than counted in
memory, because the run that placed the earlier attempts is usually a process that no
longer exists.

Then the other half of what you asked for, genuinely new only after the previous
outcome is known. `releaseRound` calls `unaccountedAttempt` before every call and
leaves the party owed when one stands, so the attempt is reconciled under its own key
instead. That covers both shapes: an attempt record nothing names and a recorded
result this app cannot read as finished. The guard is in the round rather than in its
callers, because a fresh coordination and a resume both place release calls and only
one of them reads the ledger first.

## 2. `payload_digest` was written and never read

Confirmed and the reason it matters is the case you named. While CALL-E holds the
key, a changed body comes back `idempotency_conflict` and the round stops, which is
what the docs claimed as the protection. When the first request never reached CALL-E
the key is unknown there, so nothing is replayed and the create just places a call
with whatever body goes out now. An upgrade to one line of a call script between the
crash and the resume is enough to make that a call saying words nobody approved.

`resume` now rebuilds the payload it would send under that key, digests it and
compares it with the digest the attempt record holds. A mismatch is a refusal and a
line in the note, not a call. The doc claim was wrong the same way the code was, so
that paragraph is rewritten rather than left standing.

## 3. Neither the provider nor the account was bound

Confirmed. `--base-url` or a different `CALLE_API_KEY` on the resume was all it took.
The create there is not a reconciliation at all: it creates the call in that
namespace while the original stays ambiguous where it was placed, which is two calls
to one person wearing the look of recovery.

`CallePort` now reports `origin` and `account` and `call_attempt` records both
beside the key. `account` is `key:<16 hex of sha256>`, so the ledger can say which
credential without holding it. Resume compares both before it touches the call, the
read of a known call id included and refuses on either.

One more from the same class while I was in there. A recorded key with no
`call_attempt` behind it is refused too. Nothing would say which payload or which
provider that string went out against, so there is nothing to check, which is the
same reasoning that already refused a call with no key at all.

## 4. Cancellation stopped short of the create retry

Confirmed. It reads like a lookup and it is a create.

No create goes out after the signal fires now and the two cases are not recorded the
same way, because they are not the same. Before the first create nothing has been
sent, so no key is claimed, nothing is appended and the call reads `not_placed`.
Before the reconciliation the first request may already have been accepted, so the
call comes back unresolved under the key already on disk with
`<code>, then canceled` and resume settles it. Writing that one off as canceled
would be the same mistake in the other direction.

## The tests you were right about

Two of them and they were hiding each other.

`openAttempts` matched a result to an attempt by party and phase, not by key. That
was fine while one call had one attempt. With two, the entry settling the first read
as the answer for both, so a second attempt that crashed before its result became
invisible to replay and to recovery. It matches by key now. A result carrying no key
at all is still matched by party and phase, since that is all a pre-key ledger has.

The stub port replayed by script identity rather than by idempotency key, so it
answered any second create for the same party and phase with the same call whatever
key was sent. It could not tell a retry from a replay, which is exactly why the
release test could only count entries. It honours the key the way the API does now:
a key it has seen returns that call, a key it has not is a new call with a new id,
and it exposes the calls it actually created so a test can count those instead.

The release test now asserts against `fake.created`: two provider created calls, two
different call ids, keys ending `-a1` and `-a2`, the phone ringing twice and the two
attempt records on disk numbered 1 and 2.

## New tests

`test/reissue.test.ts` is what has to still hold before a recorded key goes back on
the wire, six cases: a payload that no longer matches, another provider origin,
another account, a key with no attempt record, an unaccounted attempt reconciled
under its own key rather than retried and the round's own guard. Origin and account
are real ports from `createSdkPort`, a second fake server for one and a second
credential for the other, so the fingerprints differ the way they would in a live
resume.

Cancellation is in `test/unresolved.test.ts`: the abort fires the moment the create
fails and the create is not tried again. Plus the one step earlier, a `placeCall`
with a signal already fired, which sends nothing and writes nothing.

Every one of them was run against the previous head first and fails there.

## Two things worth flagging

The key format changed, so `examples/ledger.example.jsonl` was regenerated from a
demo run. Old ledgers still resume: recovery reads the key rather than deriving it,
and an entry with no `attempt` field reads as attempt 1.

`unsettledCall` moved from `resume.ts` to `call-state.ts` and is re-exported. Three
places need the same answer now, recovery, the release round and replay and that
module exists so they cannot drift apart on what finished means.

Verified locally on `d178694`: 140 tests pass (131 before), `tsc --noEmit` clean,
`python3 scripts/validate_repository.py` passes, `npm run demo` regenerated and every
ledger it writes replays.

Ray-56:
Current blocker at d178694:

- A second live run against the same completed durable ledger is treated as a new attempt for every gather and confirm call. I reproduced this with the localhost fake: the first run created 6 calls with -a1 keys and the second created 6 more with -a2 keys. A completed coordination must be returned or refused by run, with resume reserved for recovery; only an explicitly authorized, settled release retry should mint a new attempt key.

zkasuran:
Fixed at `4cd232e`. Confirmed. Reproduced with the same setup before I touched
anything: the first run created 6 calls with `-a1` keys, the second created 6 more with
`-a2`. Twelve calls to three people about one appointment and the file stopped
replaying, because every line of the second round follows an outcome with no
`resume_started` opening one.

```
run 1: the provider created 6 calls
  call_fake1 gather  +14155550101 mps-ash-lane-3b-leak-gather-plumber-395e865b6730-a1
  ...
  call_fake6 confirm +14155550102 mps-ash-lane-3b-leak-confirm-superintendent-thu-14-c30d33ccc779-a1
run 2: the provider created 6 calls
  call_fake7  gather  +14155550101 mps-ash-lane-3b-leak-gather-plumber-395e865b6730-a2
  ...
  call_fake12 confirm +14155550102 mps-ash-lane-3b-leak-confirm-superintendent-thu-14-c30d33ccc779-a2
total provider calls 12
replay of the shared ledger: ok=false, 42 entries, outcome verbally_confirmed
  entry 22: run_started follows outcome verbally_confirmed with no resume_started entry opening a new round
  entry 42: calls_placed 6 does not match the 12 call entries in this ledger
```

## Why the second run rang everybody

The attempt number in the key is counted from the ledger, which is what makes a release
retry a call the provider has never seen. `run` read the ledger for that number and for
nothing else. So a file that already held a finished round was a history to count from
rather than a coordination to read: every gather and confirm call derived a key one
attempt higher, CALL-E had never seen that key, so nothing was replayed and each person
heard the same question again about an answer already on disk. Making the retry real is
what made this reachable, which is worth saying plainly.

## The rest of the class

Before writing anything I went looking for what else re-mints an outcome that is already
settled on disk, because gather and confirm on a completed ledger is one instance of it.
The same second run against an **interrupted** ledger is worse than a duplicate call:

```
run 1 died: power cut
run 1: the provider created 5 calls          three gathers, two confirms, both -a1
run 2 outcome verbally_confirmed, calls_placed 6
run 2: the provider created 6 calls          the two -a2 keys and the third confirm
total provider calls 11
```

The first round had two yeses on disk and nobody had been told it was off. The second
round rang all three again and closed the file as `verbally_confirmed`, which wrote that
release call out of the history: nothing was left owing it, so nothing would ever place
it. A duplicate call is a bad afternoon. That one loses a duty.

So the class I closed is "an outcome already settled on disk being minted as a fresh
attempt", not "gather and confirm on a second run".

## What run does now

It reads the ledger under its own lock, before it dials anybody. There are three
answers and no fourth. This is the shape you named.

- A round that finished with every call settled is handed back as it stands. No call is
  placed and **nothing is appended**, so running twice is reading twice and the file
  still replays as one round. The outcome, the slot, who confirmed and the call count
  are the recorded ones: this run placed nothing, so claiming a call of its own would be
  a call in the accounting and none on a phone. The note says where it came from.
- A round that did not finish or one that finished still owing somebody a release call is
  refused with `resume` named and the reason spelled out, down to which party is owed
  what. `resume` is the path that settles a call by sending the key it went out under,
  which is exactly why a second run must not try.
- A file that is not this coordination's is refused outright: another request's digest,
  lines no `run_started` opens or a torn last line. That last one is what a crash during
  an append leaves and reading a ledger drops it, so without the check a second run
  would see a shorter history and append a round to it. `resume` repairs that line under
  the lock.

Same repro, same fake, at `4cd232e`:

```
run 2 outcome verbally_confirmed, calls_placed 6, note: every party confirmed the time by
  voice, 3 of 3; read from the ledger, which already records this coordination, so this
  run placed no call
run 2: the provider created 0 calls
total provider calls 6
replay of the shared ledger: ok=true, 21 entries, outcome verbally_confirmed
```

and on the interrupted one:

```
run 2 refused: <ledger> records this coordination and no entry closes it, so the run did
  not finish and this is a recovery rather than a new run. resume settles every call
  under the key the ledger recorded for it and places the release calls that are owed,
  where a second run derives a new key for each one and rings everybody again. Run resume
  --request <file> --ledger <ledger> --live. Nothing was dialled.
run 2: the provider created 0 calls
```

## What may mint an attempt key

One rule, `attemptToMint` in `ledger.ts`. Both places in the app that derive a key
ask it, so a fresh coordination and a recovery cannot answer it differently. Recovery
derives no key at all, it re-issues the string the ledger recorded.

A key is minted for a call this ledger holds no attempt for. The single exception is a
release retry that was asked for and whose last attempt is settled.

Nothing else. That is the rule rather than a carve-out for the case you found. A
gather or a confirm call is placed once per coordination, so a second attempt at one is a
second phone call about an answer the ledger already holds. A release call is the one
call this app places twice, because a release nobody acknowledged leaves the person owed.
There both conditions have to hold: the last attempt has to be accounted for, since
one that may still be on the phone to somebody is reconciled under its own key instead
and the retry has to be asked for with `resume --retry-release`. A new key is a handset
ringing rather than a lookup. Nothing else recovery does can reach anybody who is not
already being called, so that one call is the only thing that needs authorizing. Without
the flag the debt is reported and stays owed. `run --retry-release` is refused rather than
ignored: a run has no earlier attempt to retry and a flag that looks like it authorizes
a call and does nothing is worse than no flag.

`inspectLedger` moved from `resume.ts` to `ledger.ts` for this, so the check `run` makes
and the state `resume` recovers from are the same reader. Two readers would have drifted
on what "finished" and "owed" mean, which is the whole question here.

## The tests

`test/settled.test.ts`, nine cases. Every one of them counts `fake.created`, the calls
the provider actually made, rather than ledger lines or API calls.

Your repro first: one run, then a second against the same completed ledger. 6 provider
calls, all `-a1`, then **0** more, the recorded outcome handed back, the file byte for byte
what the first run left and the ledger still replaying. Before the fix that second run
created 6 calls and left a history with 23 replay issues.

Then the class:

- the interrupted ledger above: refused, 0 new calls, the file untouched, then `resume`
  finishes it and both parties who said yes are told, most recent first
- a finished round that still owes a release call: refused, naming the party and the debt
- a ledger a different request wrote and a ledger torn by a crash mid append: refused
  before the first call
- a release retry nobody asked for: 0 provider calls, the debt reported, the progress line
  names `resume --retry-release`
- the same ledger with the retry authorized: one more provider call, two different call
  ids, keys `-a1` then `-a2`, so the phone rang twice
- an authorized retry while the earlier attempt is unaccounted for: still nothing sent,
  because being asked for is not the only condition
- the rule on its own, including that authorizing a retry does not reach a gather or a
  confirm call, since there is nothing there to retry

`test/resume.test.ts` covers the authorized retry end to end and now passes the flag. Two
CLI cases pin the refusal of `run --retry-release` and the help text. `README.md`,
`docs/protocol.md` and the `--help` output all describe the new behaviour, including two
new sections in the protocol doc for the two rules.

I cannot claim the new file fails at `d178694`: it imports the rule and the refusal, so it
does not load there at all. What I ran against the old head is the repro above and it
produced the 6 plus 6 and the 5 plus 6.

## Verification

At `4cd232e`, all run locally:

- `npm test`: 151 tests pass, 0 fail, 0 skipped. 140 before this round.
- `npx tsc --noEmit`: clean.
- `python3 scripts/validate_repository.py`: `Repository validation passed.`
- `npm run demo`: unchanged, the resumed ledger still replays `ok=true, 25 entries`.
  `examples/ledger.example.jsonl` is untouched, since nothing about the key format or the
  entries moved.

Ray-56:
Current blocker at `4cd232e`: the completed-ledger handback checks only for unsettled calls and release debt, then trusts the closing `outcome`. It does not run the existing semantic replay checks that reject `verbally_confirmed` without every confirmation, a wrong chosen slot, or a mismatched call count. A syntactically complete but inconsistent ledger can therefore be returned as a successful coordination without dialing. Run the replay/semantic verification before returning the recorded result.

zkasuran:
Fixed at `02fd696`.

`recordedRound` now runs the full `replay(entries)` before it hands a completed ledger back. It refuses when the replay is not ok. The check sits right after the existing `inspectLedger` outstanding check (unsettled calls, owed releases) that routes to `resume`. That ordering is the point: by the time the new check runs, the ledger already has a closing outcome, no unsettled call, no owed release, so an interrupted-but-valid ledger has already gone to `resume`. Anything replay still finds at that point is a genuine semantic inconsistency in a file that claims to be finished. That case is refused, not returned and not resumed.

So the three shapes you named are now caught by the same gate that `replay` already had:

- `verbally_confirmed` with a party who was never credited: replay says `<party> never confirmed`.
- a recorded outcome slot that is not the one the answers choose: `confirmed slot X is not the chosen slot Y`.
- a `calls_placed` that does not match the call entries: `calls_placed N does not match the M call entries`.

On `!replay.ok` it throws `AlreadyCoordinatedError` (still a `LedgerError`, so the CLI reports it as a usage error) listing every issue by entry with its problem. It says plainly that a run will not return an outcome the history does not support and that nothing was dialled. I ran the class both ways. Each inconsistent ledger is refused with zero provider calls. A genuinely consistent completed ledger is still handed back idempotently with no call, the file byte for byte unchanged. The interrupted-ledger path and the owed-release path still route to `resume` exactly as before.

One decision worth calling out. A gather call attempted with no result behind it (`unsettledGathers`) is not in `inspectLedger`'s outstanding list, because `resume` never re-gathers, so it is nobody's to finish. But `replay` flags it as an open attempt. I let that refuse here rather than hand back: a gather call may have been live, so an outcome resting on it is not trustworthy. Routing to `resume` would be wrong anyway since `resume` cannot re-gather. Fail closed.

New tests in `test/settled.test.ts`, five cases: the three inconsistent ledgers above, the gather-orphan ledger, plus a regression guard that a consistent completed ledger still hands back with no call. Each builds a real completed ledger, asserts it replays clean, edits one entry so the file is still syntactically complete but no longer replays, then asserts the run refuses while `fake.created` does not move. I confirmed the refusal cases fail against the prior handback (4 of the 5 fail without the fix; the consistent-ledger guard passes either way) and pass with it.

Verification at `02fd696`, all local:

- `npm test`: 156 pass, 0 fail, 0 skipped (151 before this round).
- `npx tsc --noEmit`: clean.
- `python3 scripts/validate_repository.py`: `Repository validation passed.`

Ray-56:
Current blocker at `02fd696`: the new replay gate still does not validate every field returned from the closing outcome. I changed only `confirmed_with` in a clean completed ledger to `[]`; `replay(entries)` still returned `{ ok: true, issues: [] }`, so `recordedRound()` can hand back `verbally_confirmed` with an empty confirmation list. `unreleased` is trusted the same way, and replay derives parties/slots from the ledger rather than the supplied request. Derive or compare all returned handback fields against the request and replayed history before trusting the outcome.

zkasuran:
Fixed at `f9bb298`.

You were right that the gate only checked some of what the handback returns. `replay` verified `calls_placed`, the credited set for `verbally_confirmed` and that owed parties were covered, but it never read the outcome entry's own `confirmed_with`. Its `unreleased` check was one-directional too. `recordedRound` then lifted `confirmed_with`, `unreleased` and `slot_id` straight off the entry. So your edit (only `confirmed_with` to `[]`) replayed ok and was handed back as a verbal confirmation with an empty list.

I closed it on both sides so a field cannot leave `recordedRound` unless replay validated it.

In `replay`, at the outcome entry, every returned field is now derived from the replayed history and compared:

- `confirmed_with` must equal the parties the recorded answers credit. For `verbally_confirmed` that is every party (a missing one is already refused by the credited check). For every other outcome it is nobody. Your exact repro now produces `confirmed_with empty does not match the parties the recorded answers credit (plumber, tenant, superintendent)` and `ok: false`.
- `unreleased` must not name anybody the answers do not owe a release call. The existing check already caught a missing owed party (a subset check), so the two directions together pin the set exactly: a dropped name fails the old check, a spurious name fails the new one.
- `slot_id` and `calls_placed` stay checked as they were.

In `recordedRound`, the returned `confirmed_with` and `slot_id` are now built from the request and the outcome kind the way a live `coordinate` run builds them, not lifted off the entry. `slot_id` is the one field replay does not constrain outside `verbally_confirmed`, so deriving it here closes that last gap directly. `unreleased` and `calls_placed` are still lifted, which is safe because replay now proves each equals the derived value.

On your parties/slots point: `replay` does derive parties and slots from `run_started`, but `recordedRound` has already refused any ledger whose `run_started.request_digest` does not equal `requestDigest(request)` (the digest binds the whole request). So the derived history the outcome fields are checked against is request-bound, it is not ledger-only data that could disagree with the supplied request. The fields I rebuild in the handback come from `request` directly.

Tests in `test/settled.test.ts`: your repro (`confirmed_with` emptied) and a spurious `unreleased` name both refuse now with zero `fake.created` calls. The consistent-ledger guard still hands back idempotently. I confirmed both new cases fail against `02fd696` before the fix and pass after.

Verification at `f9bb298`, all local (Node 22):

- `npm test`: 158 pass, 0 fail, 0 skipped (156 before this round).
- `npx tsc --noEmit`: clean.
- `python3 scripts/validate_repository.py`: `Repository validation passed.`

PR #43, opened 2026-07-29
feat(apps): add call-on-behalf errand caller with a disclosure budget

zkasuran:
## Summary

Adds `apps/typescript/call-on-behalf`, a runnable app that makes one delegated
phone call for somebody who cannot make it themselves, and hands back the answers,
the transcript and an exact record of what was said about them.

Some things can only be arranged by phone: a clinic with no online booking, a
garage, a landlord, an office whose web form has been down for two years. For
somebody deaf or hard of hearing, or whose language does not match the line they
have to call, that is not an inconvenience, it is a door that does not open.

The design starts from two limits rather than treating them as disclaimers. The
call never claims to be a person: the first sentence says it is an automated
assistant calling on behalf of a named person, with their permission. And it is not
a relay service, so `docs/scope-and-limits.md` says what Telecommunications Relay
Service is and why this does not replace it.

What the app does:

- **A disclosure budget with three gates.** Delegating a call means handing over
  the right to say things about you, so the errand file carries an explicit list of
  details the caller may give, and nothing else may be said. Gate one refuses a
  request file that hides an identifier in a question, and refuses a payment card,
  a password or a national identifier in the budget outright, Luhn check included.
  Gate two scans the generated script before any call is created, and a finding
  refuses the errand with no call placed. Gate three scans what the caller actually
  said and reports anything outside the budget to the person it belongs to.
  Findings are masked everywhere, because a privacy report that quotes the leak is
  not a privacy report.
- **Commitment bounded by authorization.** `goal.commitment` is the whole
  authority: `none`, `slot_within_windows` or `confirm_existing`. A time offered
  outside the authorized windows comes back as a proposal with nothing agreed, and
  if the extracted result claims a time was accepted outside them, the report says
  `outside_authorized_window` and tells the person to check and cancel it rather
  than hiding it behind a success message.
- **A refusal is an answer.** Plenty of businesses will not deal with an automated
  caller. That is detected from the transcript, recorded as
  `callee_declined_automated`, and the next step tells the person to call another
  way. The app does not argue and does not dial back with a different script.
- **The report is the product.** A person who could not hear the call gets the
  answer per question, what was agreed with any confirmation code, which authorized
  details were actually spoken and which were not needed, the privacy findings and
  the verbatim transcript, labelled for a reader. Written with mode `0600`.

CALL-E usage at runtime: `calls.create` with a result schema built from the
questions (one field per question plus commitment, offered time, confirmation code
and whether the callee refused an automated caller), the recipient locale from
`policy.language`, `metadata` carrying the errand id, and `cob-<errand_id>` as the
`Idempotency-Key` so a retried run reuses the call instead of ringing again. Then
`calls.waitForResult`, with `calls.get` to snapshot a call that outlived the
timeout. The transcript turns are read locally for the two things a structured
result should never be the only source of: whether a person was there at all, and
whether that person told the caller to go away.

Side effects: one call per errand, nothing recurring, so there is no schedule to
clean up. `preview` and `show` place no calls and need no credentials, and preview
is the consent step, printing the exact script and the disclosure list before
anything rings. `CALLE_API_KEY` is read from the environment only.

## Verification

```text
npm test        87 tests pass
npm run demo    6 cases: a clean errand, a clinic refusing an automated caller,
                a time outside the authorized windows, the three privacy gates,
                an ambiguous call reconciled by its key, and an unsupported claim
```

CALL-E in `fake/calle-server.ts`, which speaks the documented wire contract (201 on
create, `Idempotency-Key` reuse, snake_case payloads, stable error envelopes). The
committed `examples/report.example.json` is unedited output from `npm run demo`.

Note for review: this is my third contribution, separate from #41 and #42, and the
three share no files. All three wrap the CALL-E SDK behind a small port, so if you
would rather have one shared adapter under `apps/shared/`, say so and I will factor
it out in whichever lands last.

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,7 @@ Runnable demo apps live under [`apps/`](apps/). They are not a CALL-E SDK and do
 
 | App | Language | Purpose |
 | --- | --- | --- |
+| [`apps/typescript/call-on-behalf`](apps/typescript/call-on-behalf/) | TypeScript | Delegated errand caller with a disclosure budget: says only the details the person authorized, commits only inside authorized windows, and returns the answers plus the transcript. |
 | [`apps/python/callback-window-coordinator`](apps/python/callback-window-coordinator/) | Python | Consent-first callback-window coordinator with masked preview, stable idempotency, and structured CALL-E results. |
 | [`apps/python/batch-runner`](apps/python/batch-runner/) | Python | JSONL batch runner using CALL-E CLI auth state, FastMCP, Rich output, and MCP tool-call metadata. |
 | [`apps/python/broker-login-client`](apps/python/broker-login-client/) | Python | CALL-E brokered login client with local token cache and MCP HTTP calls. |
@@ -154,6 +155,7 @@ Plugins should be explicit about inputs, outbound call side effects, credential 
 ### Safety patterns
 
 - [`Safety reference`](skills/call-reminder/references/safety.md) - Consent, E.164 phone-number handling, credential boundaries, cancellation, duplicate-job prevention, and medical reminder boundaries.
+- [`Disclosure budget`](apps/typescript/call-on-behalf/docs/privacy-budget.md) - Authorizing what a caller may say about a person, checking the script before the call and checking what was actually said after it.
 - [`Design principles`](docs/design-principles.md) - Repository-wide architecture principles for safe phone-call workflows.
 
 ## Contributing
--- a/apps/README.md
+++ b/apps/README.md
@@ -10,6 +10,7 @@ Current apps:
 
 | App | Language | Purpose |
 | --- | --- | --- |
+| [`typescript/call-on-behalf`](typescript/call-on-behalf/) | TypeScript | Delegated errand caller with a disclosure budget: says only the details the person authorized, commits only inside authorized windows, and returns the answers plus the transcript. |
 | [`python/callback-window-coordinator`](python/callback-window-coordinator/) | Python | Consent-first callback-window coordinator with masked preview, stable idempotency, and structured CALL-E results. |
 | [`python/batch-runner`](python/batch-runner/) | Python | JSONL batch runner using CALL-E CLI auth state, FastMCP, Rich output, and MCP tool-call metadata. |
 | [`python/broker-login-client`](python/broker-login-client/) | Python | CALL-E brokered login client with local token cache and MCP HTTP calls. |
--- a/apps/typescript/call-on-behalf/README.md
+++ b/apps/typescript/call-on-behalf/README.md
@@ -50,9 +50,11 @@ The extraction proposes. The transcript decides.
   after the caller asked that question. The report prints that turn. An answer
   nobody can be quoted saying comes back as not answered, with a note that CALL-E
   claimed one.
-- Something is agreed only when the transcript shows somebody agreeing. Otherwise
-  the commitment reads `unconfirmed` and the next step says to treat nothing as
-  booked. The confirmation code belongs to the agreement, so it is dropped too.
+- Something is agreed only when the transcript shows somebody agreeing, after the
+  caller raised the arrangement and about the time the report would print.
+  Otherwise the commitment reads `unconfirmed`, the next step says to treat nothing
+  as booked and the note quotes whatever they did say. The confirmation code
+  belongs to the agreement, so it is dropped too.
 - A call this app could not read comes back as `outcome_unknown`, never as a
   failure and never with "nothing was said". So does a call CALL-E has not finished
   with: only a terminal status is read as a result, and a call still `queued` or
--- a/apps/typescript/call-on-behalf/README.md
+++ b/apps/typescript/call-on-behalf/README.md
@@ -54,7 +54,11 @@ The extraction proposes. The transcript decides.
   caller raised the arrangement and about the time the report would print.
   Otherwise the commitment reads `unconfirmed`, the next step says to treat nothing
   as booked and the note quotes whatever they did say. The confirmation code
-  belongs to the agreement, so it is dropped too.
+  belongs to the agreement, so a code nobody read out is dropped even when the
+  agreement stands.
+- A time no callee turn named is not read back as an offer, and CALL-E's own note
+  is printed with its name on it. Nothing in the report is both unchecked and
+  unlabelled.
 - A call this app could not read comes back as `outcome_unknown`, never as a
   failure and never with "nothing was said". So does a call CALL-E has not finished
   with: only a terminal status is read as a result, and a call still `queued` or
--- a/apps/typescript/call-on-behalf/README.md
+++ b/apps/typescript/call-on-behalf/README.md
@@ -56,12 +56,12 @@ The extraction proposes. The transcript decides.
   as booked and the note quotes whatever they did say. The confirmation code
   belongs to the agreement, so a code nobody read out is dropped even when the
   agreement stands.
-- A time no callee turn named is not read back as an offer, and CALL-E's own note
+- A time no callee turn named is not read back as an offer. CALL-E's own note
   is printed with its name on it. Nothing in the report is both unchecked and
   unlabelled.
 - A call this app could not read comes back as `outcome_unknown`, never as a
   failure and never with "nothing was said". So does a call CALL-E has not finished
-  with: only a terminal status is read as a result, and a call still `queued` or
+  with: only a terminal status is read as a result. A call still `queued` or
   `in_progress` gets no verdict, no commitment and no privacy finding. Re-running the
   same errand file reads the same call back, because the idempotency key covers the
   content of the call and has not changed.
@@ -177,8 +177,11 @@ npm run errand -- show --report report.json
 the file and the receipt changes, which is the point: the consent belongs to a
 preview somebody read.
 
-Reports are written with mode `0600`. They hold the full transcript on purpose:
-the person is entitled to what was said on their behalf.
+Reports are written with mode `0600`, including over a report file that already
+exists: the mode is set on the descriptor, so a file left at `0644` comes back
+`0600` and a path that is not a regular file is refused rather than followed. They
+hold the full transcript on purpose: the person is entitled to what was said on
+their behalf.
 
 ## The errand file
 
@@ -224,8 +227,11 @@ required to make them write it down.
 - `preview` and `show` place no calls and need no credentials.
 - `CALLE_API_KEY` is read from the environment only, never from the errand file.
 - The key goes out on every request, so the base URL is checked before the client
-  is built. HTTPS anywhere, plain HTTP only on `localhost`, `127.0.0.1` or `::1`
-  for the local fake. Anything else is refused by name and the key is not sent.
+  is built. The host has to be `api.heycall-e.com`, `localhost`, `127.0.0.1` or
+  `::1` for the local fake, which is also the only place plain HTTP is allowed. Any
+  other host has to be named exactly in `CALLE_ALLOWED_HOSTS` or with
+  `--allow-host`: no wildcards, no suffix matching, port ignored. Anything else is
+  refused by name and the key is not sent.
 - The report holds the transcript and the disclosure record. Numbers are masked.
   Findings are masked. Keep the file the way you keep anything with your own
   details in it.
--- a/apps/typescript/call-on-behalf/README.md
+++ b/apps/typescript/call-on-behalf/README.md
@@ -155,6 +155,9 @@ may accept, the result contract, the privacy check and a receipt for that exact
 preview. It contacts nothing. The person sees what will be said about them before
 anything rings.
 
+The receipt covers every line the preview printed, including the reason for
+delegation. That field is hashed on this machine and still never sent to CALL-E.
+
 ## One live call
 
 The receipt is what carries consent from the preview to the call.
--- a/apps/typescript/call-on-behalf/README.md
+++ b/apps/typescript/call-on-behalf/README.md
@@ -1,0 +1,184 @@
+# Call On Behalf
+
+Some things can only be arranged by phone. A clinic with no online booking, a
+garage, a landlord, a government office where the web form has been "temporarily
+unavailable" for two years. If you cannot use a phone line, that is not an
+inconvenience, it is a door that does not open.
+
+This app makes one delegated call and hands back what was said. It states in its
+first sentence that it is an automated assistant calling on behalf of a named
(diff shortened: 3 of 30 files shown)


Ray-56:
Thanks for the strong disclosure-budget design. I found these current blockers:

- One broad catch treats an ambiguous create/wait failure as `api_error` and states that nothing was said, although the call may be running or complete. Report `outcome_unknown`, reconcile by the same key, and do not suggest a blind retry.
- Commitments and per-question answers are accepted from `structured_result` whenever any user turn exists; transcript evidence is not bound to each question or the claimed acceptance. A hallucinated extraction can therefore be reported as an agreement.
- `call --live` does not require a receipt/hash from the displayed preview, so the documented preview consent step can be skipped. The delegated permission and calling-time authorization are otherwise only self-asserted input prose.
- The task sends `reason_for_delegation` to the provider and the pattern scanner cannot detect ordinary sensitive prose; clinical/legal/financial details can also enter through goal/question text despite the stated boundary.
- The idempotency key binds only `errand_id`, and arbitrary `--base-url`/`CALLE_BASE_URL` receives the API key without canonical-content binding or HTTPS/trusted-host validation.

`diff --check` and repository validation pass. No lockfile/CI checks are present, so typecheck/tests/fake-provider validation was incomplete here. No real call was placed.

zkasuran:
Thanks, all five were real. Pushed as four commits on `feat/call-on-behalf`
(`b58c87f`, `96f52ae`, `2ca22fa`, `f7aa501`).

**Ambiguous create or wait.** Fixed. `src/errand.ts` no longer wraps create, wait
and read in one catch. `CalleCallError` now carries the HTTP status and says
whether the failure leaves the call unknown (`src/calle.ts:38-44`: no status, 408,
429 or 5xx is ambiguous, anything else is a refusal). An ambiguous create
reconciles by re-issuing the same idempotency key, which returns the existing call
instead of ringing again (`src/errand.ts:165-179`). A wait or read failure keeps
the call id and tries a direct read (`src/errand.ts:186-204`). If the call still
cannot be read the outcome is `outcome_unknown` with `call_status: "unknown"`, the
call id kept and exit code 40 (`src/errand.ts:215-247`, `src/cli.ts:144-146`). The
next step says nobody knows yet whether the call was made and that running the
same errand file reads that same call back, because the key is unchanged. It does
not suggest a blind retry. It also says not to edit the file first, since an
edited file is a different call. The report no longer prints "nothing was said", "no
person on the line", "not answered" or a privacy verdict for an unknown call:
those become "not known" (`src/report.ts:99-160`). A refusal still reports
`api_error`, because then nothing was created.

**Evidence binding.** Fixed. `src/read.ts` gained `supportingTurn` and
`agreementTurn`. A per-question answer is reported only when a callee turn
supports that specific question. That turn is now in the report's `quote` field,
which used to be a placeholder (`src/errand.ts:260-271`). A `yes_no` answer
is bound to the question: the caller must have asked it in the transcript and the
answer is read from the next two callee turns, with the first polarity marker
deciding, so "No, but we do take Aetna" supports a no and not a yes. A text or
datetime answer is bound by its own words, at 60 percent token overlap with one
turn. An unsupported answer comes back `answered: false` with the extraction
dropped and a note saying how many CALL-E claimed (`src/errand.ts:336-340`). A
claimed acceptance now needs a turn that shows somebody agreeing, otherwise the
commitment reads `unconfirmed`, `committed_datetime` is null, the confirmation
code is dropped with it (`src/errand.ts:281-297`, `src/errand.ts:357`) and the next
step says to treat nothing as booked. Report shape is unchanged: one new
`commitment` value, one new `outcome` value, `quote` finally populated. The
matching is lexical, so it is sometimes too strict; a paraphrase it cannot see
reads as unsupported. `docs/scope-and-limits.md` says so and says why that is the
direction to be wrong in.

**Preview consent as a receipt.** Fixed. `previewReceipt` is a 16 character sha256
over the canonical JSON of the script, the disclosure list and the windows
(`src/script.ts:240-256`). The preview prints the exact command to run with it
(`src/report.ts:84-90`) and `call --live` refuses without it or with a stale one,
before the API key is read and before any client exists (`src/cli.ts:107-121`). The
refusal does not print the correct receipt, because handing it over would defeat
the step. `test/cli.test.ts` spawns the CLI for all of that. You are right that
the delegated permission itself is still self-asserted prose: the receipt binds
consent to a script, not to a person. `docs/scope-and-limits.md` now says that in
the consent section rather than implying more.

**`reason_for_delegation` and prose leakage.** Two parts.

Fixed: the reason is no longer sent. It was in the first line of the task
(`src/script.ts:69` now ends at "who asked you to make it"). The example reason is
a disability disclosure, so shipping it to the provider was the app's own thesis
being broken in its own example. It stays in the errand file, is shown in the
preview marked as never sent. `test/script.test.ts` asserts it is absent from both
the task and the whole payload.

Partly a correction: goal and question prose was already scanned by the same
detectors, so an identifier in either refuses the call before it is placed.
Questions are checked at load (`src/config.ts:155`) and the goal is inside the
script that gate 2 scans (`src/errand.ts:63-65`). What genuinely got through is
ordinary prose, which no pattern scanner catches. Added: a clinical, legal and
financial keyword scan over `goal.summary` and each question, reported in the
preview next to the field, at warn severity only (`src/disclosure.ts:209`,
`src/errand.ts:62-70`). It has to warn rather than block, since "book a routine
check-up" is the errand. `docs/privacy-budget.md` states plainly that it catches
words and not meaning. It also says that "she has been unwell since the spring"
trips nothing.

**Idempotency key and base URL.** Both fixed. The key is now
`cob-<errand_id>-<12 hex of sha256 over the canonical JSON of the payload>`
(`src/script.ts:188-189`), so a retry reuses the call and an edited errand is a new
call rather than a 409. One canonical JSON helper serves both the key and the
receipt (`src/script.ts:231`). One `buildCallInput` builds the payload both of them
hash (`src/script.ts:202`). Base URL: `assertTrustedBaseUrl` runs before the client
that carries the key is built (`src/calle.ts:60-78`, called at `src/calle.ts:80`).
`https:` anywhere, `http:` only for `localhost`, `127.0.0.1` and `::1` so the local
fake still works, everything else refused with both `--base-url` and
`CALLE_BASE_URL` named. It refuses to send the credential rather than warning.

## Verification

Run in `apps/typescript/call-on-behalf` and at the repo root:

```
npm test                                87 tests, 87 pass, 0 fail, 0 skipped
npm run demo                            6 cases, exit 0
python3 scripts/validate_repository.py  Repository validation passed
```

Baseline before this round was 53 tests. The 34 new ones cover the trust gate, the
CLI consent gate, the evidence binding, the unknown outcome and the content hashes.
Against the pre-review `src/` the new suite fails 13 times, which is where each fix
is pinned.

On the environment note: `npm test` was `tsx --test`, which wants an IPC socket of
its own. It is `node --import tsx --test` now, same tests, no socket, so it runs in
a restricted sandbox. `npm run demo` too. No app in this repo commits a lockfile
(`apps/python/*` ship `uv.lock`, none of the four TypeScript apps ship one), so I
have not added one here rather than break that pattern. Say the word if you want
lockfiles across the TypeScript apps and I will do it as a separate change.

## Still open

No live call has been placed from this branch: everything above is verified
against the local fake CALL-E over the real `@call-e/calle` client, with no
credentials and nothing beyond localhost.

Ray-56:
Thanks for the update. The current head still has blockers that the green suite does not cover:

- After a wait timeout, getCall may return queued or in_progress. That snapshot is evaluated as normal because only failed and canceled are special-cased. An isolated stub reproduced call_status in_progress being reported as outcome goal_met with commitment committed. Nonterminal states must be outcome_unknown.
- Text and datetime answers do not require their question to have been asked. When askedAt returns -1, supportingTurn scans every callee turn; an unrelated matching sentence can therefore become a reported answer. Require an anchor for every answer shape and bound evidence to that question.
- agreementTurn accepts booking language anywhere and does not bind it to the commitment prompt or extracted offered_datetime. An unrelated agreement plus a different in-window extracted time can be reported as committed.
- The preview prints reason_for_delegation, but the receipt does not hash it, so changing that field leaves the receipt valid despite the exact-preview claim.

Typecheck, 87 tests, the local fake demo, preview, repository validation, and diff check pass. No credentials or real calls were used.

zkasuran:
Thanks. All four were real. You were right that the green suite did not cover
any of them. Branch head `9955c6e`.

**1. A nonterminal read-back after the wait timeout was evaluated like a result.**

Your stub is a test: `test/errand.stub.test.ts`, "a call CALL-E has not finished
with is no result, whatever the extraction claims". The read back says
`in_progress` and carries a full transcript plus a confident extraction. It now
comes back `outcome_unknown` with `commitment: unconfirmed`, no answered answers, no
disclosures and no privacy finding.

The terminal statuses live in one list (`src/calle.ts:63`,
`isTerminalCallStatus` at `:72`) and only a terminal status is read as a result
(`src/errand.ts:228`). Anything else is `outcome_unknown` with `call_status`
unknown, the call id and the provider call id kept, exit code 40 and a next step
that says the call is still open rather than claiming nobody knows whether it was
made. "no state short of finished is evaluated, whichever one it is" runs ten
statuses through that.

Sweeping the same class turned up one of mine. That list also carried `no_answer`,
`busy` and `voicemail`. `CallStatus` has none of them: it is `queued`,
`in_progress`, `completed`, `failed`, `canceled`, which is why the local fake only
ever emits the last three. A no answer or a voicemail arrives as `failed` with a
failure code or as a completed call whose transcript is a machine, both of which
are read further down and are unchanged. The three invented values are gone, so a
status spelled that way now leaves the call unresolved. `docs/scope-and-limits.md`
documented the six as real and documents the three now.

**2. Text and datetime answers did not need their question to have been asked.**

True. The hole was exactly where you pointed: when `askedAt`
(`src/read.ts:125`) returned -1 the scan ran over every callee turn, so an
unrelated matching sentence could be reported as the answer to a question the
caller never got to.

`supportingTurn` (`src/read.ts:149`) now requires an anchor for every answer shape
and bounds the evidence to the two callee turns after that question, which is the
rule a yes or a no already followed. The 60 percent token overlap still decides
what counts as support inside that window. No anchor means unanswered, whatever
the extraction claims.

`docs/scope-and-limits.md` says which way this errs: a callee who circles back four
turns later reads as unsupported. A report that says less costs a phone call to
check. A report that says more costs somebody an appointment they do not have.

Tests: `test/read.test.ts:44`, `:60`, `test/errand.stub.test.ts:129` (an answer to a
question nobody asked is not an answer).

**3. `agreementTurn` took booking language anywhere and was not bound to the time.**

`agreementEvidence` (`src/read.ts:394`) binds twice. The agreeing turn has to come
after the turn where the caller raised the arrangement. When the extraction claims
a datetime, that time has to be named in the agreeing turn or in the turn either
side of it. `mentionsDatetime` (`src/read.ts:291`) reads the wall clock in the forms
a transcript carries: 9:40, nine forty, half past nine, ten o'clock. A turn that
names a day or a weekday has to name the claimed one, so the same time on another
authorized day is not this agreement.

Your exact case is `test/errand.stub.test.ts:144`: booking language in the call plus
a different extracted time that happens to fall inside an authorized window. Also
`:176` and `:186`, `test/read.test.ts:91`, `:102`, `test/errand.e2e.test.ts:320`.

The decision inside this one, since it is a judgement call rather than a fix. An
agreement about some other time reads `unconfirmed`, not `committed`. It does not
read `outside_authorized_window` either, because both of those print or act on a
time that came from the extraction alone. The report quotes what they did agree to,
so `unconfirmed` is not read as nothing having happened.

Blockers 2 and 3 are the same bug in two places, so the rest of the report went
through the same question: either a claim is in the transcript or it is labelled as
not being. A confirmation code has to have been read out around the agreement, with
digits said as words counting, because that is how a reference number arrives. An
offered time no callee turn named is not read back in the next step, which says
"another time" instead. An accepted agreement with no `offered_datetime` used to
print "agreed for , which is outside the windows" and now reads `unconfirmed` with a
note. CALL-E's own note is printed with its name on it. Nothing in the report is
both unchecked and unlabelled now, except `declined_by_callee`, which says less than
the extraction did and is documented as the one that stands on it.

**4. The preview printed `reason_for_delegation` and the receipt did not hash it.**

Correct, so the claim that the receipt binds the exact preview was not true.
`receiptMaterial` is now the canonical JSON the receipt hashes
(`src/script.ts:253`, `previewReceipt` at `:274`) and it covers the errand file as
the app parsed it as well as the call that would be sent.

I did the inverse check you would have done next. It found a second one:
`callee.published_source` is printed in the preview and was not hashed either.
`receiptMaterial` is exported so the claim is checkable rather than asserted:
`test/script.test.ts:176` pulls the labelled lines out of a rendered preview,
asserts the list of them is exactly what the app prints, then asserts every value
behind them is inside the material and that editing each one moves the receipt.

`reason_for_delegation` is hashed and still never sent to the provider, asserted in
the same test, because that pair is the whole point of the field: it is part of the
consent and it is not part of the call.

**Two things of ours, not yours.**

You raised the untrusted base URL in round 1 and did not repeat it here, but this
app had the same weakness as the other two: https anywhere, which checks the
transport and says nothing about who answers. The key now goes to
`api.heycall-e.com`, to loopback for the local fake or to a host named exactly in
`CALLE_ALLOWED_HOSTS` or with `--allow-host`, which is repeatable. Hostnames are
compared whole and lowercased, so `localhost.attacker.example` and
`api.heycall-e.com.attacker.example` are both refused. Only loopback may skip https
(`src/calle.ts:97`, `:112`, tests `test/calle.test.ts:16` through `:47`).

The report said 0600 and only got it on create. `writeReport` opens with
`O_NOFOLLOW`, refuses a path that is not a regular file and fchmods the descriptor,
so a report written over a 0644 file comes back 0600 and a symlink is not followed
into somebody else's file (`test/report.test.ts:111`, `:121`). One thing fell out of
that: a report that could not be saved used to lose the report. The call has already
happened by then, so the text goes to stdout either way now and the write failure is
named on stderr with a line saying that is the only copy.

**Verification** (Node 22.22.2, from `apps/typescript/call-on-behalf`, repo root for
the validator):

```text
npm test                                113 tests, 113 pass, 0 fail, 0 skipped (87 before)
npm run demo                            exit 0, 6 cases
npm run preview                          exit 0, receipt printed
```

No real call was placed. Everything above is against the local fake CALL-E over the
real `@call-e/calle` client, with no credentials and nothing beyond localhost.

PR #44, opened 2026-07-29
feat(apps): add permit status clarifier

LubuSeb:
## Problem

Municipal permit desks often report only that an application is "under review," leaving applicants without a precise blocker or next action.

## Solution

Adds **Permit Status Clarifier**, a runnable Python CALL-E app with a local web UI and CLI. It prepares one authorised, information-only call to a public permit desk and returns a structured status, blocker, next action, deadline, resubmission channel, fee information, and public follow-up reference.

## Safety and side effects

- Masked preview is the default and creates no call.
- Live mode requires explicit authority, public-number, and live-call confirmations.
- One stable idempotency key prevents accidental duplicate calls.
- The task forbids personal-data collection, legal interpretation, scheduling, submissions, payments, commitments, retries, and recurring calls.
- Structured output redacts phone numbers and email addresses.
- No credentials, analytics, persistence, or request-body logging.

## Validation

- `python -m pytest -q apps/python/permit-status-clarifier` ? 8 passed
- `python scripts/validate_repository.py` ? passed
- Local preview endpoint and desktop UI verified

## Live verification status

Preview and mocked CALL-E paths are fully verified. One authenticated CALL-E test call is still pending because this environment does not contain a `CALLE_API_KEY`; that is why this PR is opened as a draft.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,7 @@ Runnable demo apps live under [`apps/`](apps/). They are not a CALL-E SDK and do
 | App | Language | Purpose |
 | --- | --- | --- |
 | [`apps/python/callback-window-coordinator`](apps/python/callback-window-coordinator/) | Python | Consent-first callback-window coordinator with masked preview, stable idempotency, and structured CALL-E results. |
+| [`apps/python/permit-status-clarifier`](apps/python/permit-status-clarifier/) | Python | Authorised municipal permit-status caller that returns a factual blocker-and-next-action brief after a masked preview. |
 | [`apps/python/batch-runner`](apps/python/batch-runner/) | Python | JSONL batch runner using CALL-E CLI auth state, FastMCP, Rich output, and MCP tool-call metadata. |
 | [`apps/python/broker-login-client`](apps/python/broker-login-client/) | Python | CALL-E brokered login client with local token cache and MCP HTTP calls. |
 | [`apps/typescript/broker-login-client`](apps/typescript/broker-login-client/) | TypeScript | CALL-E brokered login client using `@call-e/core`. |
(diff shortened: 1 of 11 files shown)


Ray-56:
This is a promising draft, but the current head has blockers before it is ready:

- The local `/api/call` endpoint accepts three self-asserted booleans with no authenticated session or single-use preview receipt bound to the canonical request, while the server can bind beyond loopback. Any reachable client can trigger a real call when a key is configured.
- Permit reference/project details are spoken before positively establishing the correct department, and `project_type`/organization/jurisdiction/department accept names, addresses, and other sensitive prose that the README says not to send.
- Completed structured fields are displayed even when the call failed, `task_completed` is false, confidence is low, or `reached_correct_department` is `no`/`unknown`. Fail closed instead of presenting an unverified blocker/action as fact.
- The idempotency key binds only `workflow_id`, not canonical content, and ambiguous create/wait outcomes have no reconciliation state.
- CLI/server base URLs can send `CALLE_API_KEY` to arbitrary plaintext or untrusted hosts; restrict live credentials to trusted HTTPS endpoints. `/api/health` should not report live readiness from key presence alone.

`diff --check` and repository validation pass. Runtime tests require Python 3.11, unavailable in this environment; no credentials or real calls were used.

Ray-56:

Remaining Must Fix only:

- Remote-bind call endpoint permits any reachable client to dial with configured key; add auth or enforce loopback-only scope.
- Provider base URL accepts arbitrary/insecure credential origins.
- Document or represent unverified result fields as advisory and correct live-readiness claim if key presence is the only check.

PR #45, opened 2026-07-30
feat(apps): add FreshChain delivery exception resolver

lsh2546:
## Summary

Adds `apps/python/freshchain-resolver`, a runnable CALL-E application for resolving delayed temperature-controlled deliveries with authorized receiving-site operations contacts. It provides preview, simulation, a SQLite audit dashboard, and protected CALL-E SDK execution.

## Type

- [x] New runnable app

## What changed

- Added a validated delivery-exception form and masked SQLite dashboard.
- Added deterministic no-call simulations that never mutate live execution state.
- Added strict fail-closed routing to `proceed`, `rebook`, `hold`, or `escalate`.
- Bound provider idempotency to a SHA-256 digest of the canonical call payload.
- Persisted accepted call IDs before polling and audit timeout/network ambiguity as `outcome_unknown`.
- Added atomic per-workflow SQLite claims and retry locks to both web and CLI execution.
- Restricted web live mode to exact loopback binds and require a strong Bearer token for every state-changing API request.
- Restricted credentialed CALL-E traffic to the official HTTPS origin, with exact loopback HTTP allowed only for tests.
- Added a loopback capture test proving the published SDK posts and polls at runtime.

## CALL-E verification boundary

CALL-E account authentication and API-key configuration were completed. An authorized South Korean test recipient could not be called because the platform did not support the `+82` recipient region, so this PR does not claim a successful cloud call.

The non-simulated SDK runtime test starts a loopback capture server, invokes the published SDK, and asserts `POST /v1/calls`, `GET /v1/calls/{id}`, authorization, payload-bound idempotency, and structured-result behavior.

## Validation

- App pytest: **16 passed, 14 subtests passed**
- `python scripts/validate_repository.py`: **passed**
- `git diff --check`: **passed**
- Root README diff: **one added resource row**
- PR: **one commit, mergeable, no base conflict**

## Checklist

- [x] Branch name, commit message, and PR title follow `docs/git-naming-conventions.md`.
- [x] No secrets, tokens, private phone numbers, recordings, or private transcripts are included.
- [x] Phone numbers are masked in documentation, API responses, and tests unless fictional.
- [x] Live web mutations require loopback binding and Bearer authorization.
- [x] CLI and web live paths share accepted-call auditing, execution claims, and ambiguity locks.
- [x] The app creates no recurrence or hidden follow-up call.
- [x] Runnable code has a no-call path by default.


Ray-56:
Blocking review findings for head 9ee07d9:

- Restore UTF-8/LF and limit the root README edit to the new resource row. The current diff corrupts existing punctuation/tree glyphs (`쨌`, `??`, mojibake), the new UI/docs also contain visible `??`, and `git diff --check` fails across the change.
- Make live result handling fail closed. `execute()` accepts any result object and `route_outcome()` ignores terminal status, `task_completed`, confidence, required schema fields, and `human_escalation_required`; a reproduced failed/low-confidence result that requests human escalation still routes to `proceed`.
- Persist the accepted call ID before polling, record timeout/network ambiguity, and atomically guard/reconcile per-case execution. Today a post-create polling failure returns 500 with no local audit record, while concurrent `/execute` requests can race.
- Mask phone numbers in create/list dashboard responses, and do not pass the live API key to arbitrary remote/plain-HTTP base URLs. Restrict production to the official HTTPS origin and allow HTTP only for exact loopback test servers.
- Align the repository artifacts and docs: remove the out-of-scope hackathon/checked-in PR copy, remove or include the claimed `sdk-runtime-test.log`, make the CLI live-gate documentation match the code, and rename the PR/commit to the required Conventional Commit format.

lsh2546:
@Ray-56 Thank you for the blocking review. I have addressed all of the reported items in commit `8d0bbcde`:

- Restored a clean UTF-8/LF diff and limited the root README change to one added resource row. `git diff --check` now passes.
- Made live-result routing fail closed for non-terminal/failed status, incomplete tasks, low confidence, malformed or missing required schema fields, and `human_escalation_required`.
- Persisted the accepted CALL-E call ID before polling, recorded timeout/network ambiguity as `outcome_unknown`, and added an atomic per-case execution guard that blocks unsafe retries until reconciliation.
- Masked phone numbers in dashboard create/list responses and restricted credentialed production traffic to the official CALL-E HTTPS origin; plain HTTP is allowed only for exact loopback test servers.
- Removed the out-of-scope hackathon/PR-copy artifacts and stale log claim, aligned the CLI live gate and documentation, and renamed the commit and PR using Conventional Commits.

Verification completed:

- `12 passed, 11 subtests passed`
- `python scripts/validate_repository.py` passed
- `git diff --check` passed
- Remote diff is 13 files, +1,742 / -0, with no conflicts

The PR is now marked Ready for review. Could you please re-review the updated head?

Ray-56:
Thanks for the update. Two current call-safety blockers remain:

- The simulate endpoint calls record_attempt, which can reset a live case from calling or outcome_unknown to resolved. That reopens live execution and can permit a second call while the first may still be active. Simulation must not mutate live execution state.
- The provider idempotency key uses only the workflow ID. Because phone, site, and ETA can change under that ID, an edited retry can replay an older call with different content. Bind the key to the canonical call payload.

lsh2546:
@Ray-56 Thanks for the follow-up. Both remaining call-safety blockers are addressed in the updated head `773a92b4`:

- Simulation now uses a dedicated `record_simulation_attempt()` path that appends only an audit attempt. It never updates case status or decision, and regression coverage verifies that simulation cannot clear either a `calling` or `outcome_unknown` lock or reopen live execution.
- The provider idempotency key is now a SHA-256 digest of the canonical call payload: task, recipients, result schema, and metadata. Exact retries retain the same key, while changes to the phone, receiving site, or ETA produce a different key. The published-SDK runtime test now verifies the generated payload-bound key.

Validation after rebasing onto the latest `main`:

- `13 passed, 14 subtests passed`
- `python scripts/validate_repository.py` passed
- `git diff --check` passed
- PR is mergeable with one commit, 13 changed files, and the root README remains a one-line addition

Could you please re-review the updated head?

Ray-56:
Two live-path blockers remain:
- The web API has no authentication or authorization. If `FRESHCHAIN_HOST` is bound off-loopback with live mode enabled, any reachable client can create an arbitrary E.164 case and trigger a real call using the public static confirmation header.
- The documented CLI path calls `execute()` without the accepted-call callback or SQLite execution claim. A timeout or crash can lose the call ID and ambiguity state, and concurrent CLI runs are not excluded despite the README claiming both protections.

lsh2546:
@Ray-56 Thanks for the operational-security review. Both remaining live-path blockers are addressed in updated head `75c67686`:

- Web live mode now fails closed unless `FRESHCHAIN_HOST` is an exact loopback host and `FRESHCHAIN_LIVE_TOKEN` is a whitespace-free secret of at least 32 characters. Every POST route requires constant-time Bearer authentication; the static live-confirmation header remains a separate side-effect confirmation, not authorization.
- CLI `--execute` now atomically claims the workflow in SQLite before provider execution, persists the accepted call ID before polling, completes the same audit attempt on success, records `outcome_unknown` on create/poll ambiguity, and blocks concurrent or unreconciled retries. The CLI documentation now includes the shared database path.

Regression coverage verifies non-loopback rejection, weak or missing token rejection, authenticated mutation, accepted-call persistence, ambiguity auditing, and retry locking.

Validation on latest `main`:
- `16 passed, 14 subtests passed`
- `python scripts/validate_repository.py` passed
- `git diff --check` passed
- PR is mergeable with one commit and 13 changed files; the root README remains a one-line addition

Could you please re-review the updated head?

PR #47, opened 2026-07-30
feat(apps): add consent-gate preflight

ILoveBuns:
## What changed

- add a Python ConsentGate app with deterministic call-plan validation
- add redacted audit manifests and a fully offline simulation path
- enforce a 24-hour retry suppression period after recipient rejection
- keep live CALL-E execution behind an explicit plan flag, exact human confirmation, and a separately supplied API key
- add eleven unit tests and an awesome-app list entry

## Why

Outbound AI calls can be technically valid while still missing explicit recipient consent, an AI disclosure, a local calling window, recording and retention choices, or a bounded retry policy. This app makes those decisions visible and testable before dispatch.

## Safety and side effects

`validate`, `manifest`, and `simulate` are offline and never place a call. `execute` can create one real outbound CALL-E call only after all policy gates and explicit human confirmation pass. Samples use a fictional reserved number, and audit output stores only a short phone fingerprint.

The entrant-controlled CN-region test was not dispatched because that region and Mandarin are not currently listed as supported. No call ID was created and no call was placed. The redacted platform feedback is tracked in CALLE-AI/call-e-integrations#68.

## Validation

- `python3 -m unittest discover -s apps/python/consent-gate/tests -v`

Both pass locally (11/11 policy tests).

## Demo

Public 1080p video with voiceover:
https://github.com/ILoveBuns/calle-consent-gate/releases/download/v0.1.0-demo/consentgate-demo-1080p-with-voice.mp4

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,7 @@ Runnable demo apps live under [`apps/`](apps/). They are not a CALL-E SDK and do
 | App | Language | Purpose |
 | --- | --- | --- |
 | [`apps/python/callback-window-coordinator`](apps/python/callback-window-coordinator/) | Python | Consent-first callback-window coordinator with masked preview, stable idempotency, and structured CALL-E results. |
+| [`apps/python/consent-gate`](apps/python/consent-gate/) | Python | Consent-first preflight, 24-hour rejection suppression, and redacted audit manifests for CALL-E calls. |
 | [`apps/python/batch-runner`](apps/python/batch-runner/) | Python | JSONL batch runner using CALL-E CLI auth state, FastMCP, Rich output, and MCP tool-call metadata. |
 | [`apps/python/broker-login-client`](apps/python/broker-login-client/) | Python | CALL-E brokered login client with local token cache and MCP HTTP calls. |
 | [`apps/typescript/broker-login-client`](apps/typescript/broker-login-client/) | TypeScript | CALL-E brokered login client using `@call-e/core`. |
--- a/apps/python/consent-gate/HUMAN_REVIEW.md
+++ b/apps/python/consent-gate/HUMAN_REVIEW.md
@@ -1,0 +1,21 @@
+# Human design review
+
+## 2026-07-29
+
+The entrant reviewed the ConsentGate proposal and required this concrete
+behavior:
+
+> After a call is rejected, do not call that user again within 24 hours.
+
+Implementation decision:
+
+- A rejection event is stored with only a redacted phone fingerprint, outcome,
+  and timezone-aware timestamp.
+- Validation and live execution both block the same recipient for 24 hours.
+- The error reports the earliest permitted retry time.
+- At exactly 24 hours, another attempt may be considered, subject to all other
+  consent and retry controls.
+
+The entrant also confirmed that this is a personal entry, uses no employer
+code, data, devices, or identity, and does not violate the entrant's employer
+policy.
--- a/apps/python/consent-gate/HUMAN_REVIEW.md
+++ b/apps/python/consent-gate/HUMAN_REVIEW.md
@@ -41,3 +41,9 @@ inappropriate when such a kill switch is required.
 Live tasks do not interpolate freeform purpose text. Version 0.1 allowlists
 only the fixed `accessibility_test` template; any new purpose kind requires a
 reviewed code change and regression coverage.
+
+The same rule applies to the opening disclosure: version 0.1 uses one fixed
+disclosure for `accessibility_test`, and live task construction reads the
+allowlist rather than caller-supplied prose. Durable attempt numbers are bound
+into provider idempotency keys so reconciliation remains same-key while an
+authorized second attempt after a terminal first result has a distinct key.
--- a/apps/python/consent-gate/HUMAN_REVIEW.md
+++ b/apps/python/consent-gate/HUMAN_REVIEW.md
@@ -26,3 +26,18 @@ The 24-hour rule applies only to a temporary rejection. A corroborated request
 never to call again creates a permanent do-not-call record. It may be cleared
 only after a new, independently verified opt-in; merely waiting 24 hours is not
 sufficient.
+
+A temporary request such as “end this call”, “hang up”, or “call me later” is
+accepted only when high-confidence task evidence is corroborated by the
+recipient transcript. It is persisted as `rejected` and starts the same 24-hour
+cooldown. Ambiguous extraction fails closed for reconciliation.
+
+The entrant also confirmed that ConsentGate will not accept medical, legal,
+financial, or emergency content and will not create recurring or hidden
+schedules. An accepted call can be ended verbally by the recipient; the pinned
+provider SDK has no operator-side cancellation endpoint, so live use is
+inappropriate when such a kill switch is required.
+
+Live tasks do not interpolate freeform purpose text. Version 0.1 allowlists
+only the fixed `accessibility_test` template; any new purpose kind requires a
+reviewed code change and regression coverage.
--- a/apps/python/consent-gate/HUMAN_REVIEW.md
+++ b/apps/python/consent-gate/HUMAN_REVIEW.md
@@ -19,3 +19,10 @@ Implementation decision:
 The entrant also confirmed that this is a personal entry, uses no employer
 code, data, devices, or identity, and does not violate the entrant's employer
 policy.
+
+## 2026-08-03 clarification
+
+The 24-hour rule applies only to a temporary rejection. A corroborated request
+never to call again creates a permanent do-not-call record. It may be cleared
+only after a new, independently verified opt-in; merely waiting 24 hours is not
+sufficient.
--- a/apps/python/consent-gate/README.md
+++ b/apps/python/consent-gate/README.md
@@ -8,7 +8,7 @@ ConsentGate blocks a CALL-E request until the caller has supplied:
 - an opening disclosure that identifies the call as AI-generated;
 - a recipient timezone and permitted calling window;
 - a bounded retry policy;
-- a 24-hour suppression period after the recipient rejects a call;
+- temporary rejection cooldowns plus permanent do-not-call suppression;
 - explicit recording and retention choices; and
 - a human-readable purpose that does not request secrets.
 
@@ -46,6 +46,7 @@ Live use is intentionally guarded:
 
 ```bash
 export CALLE_API_KEY_FILE="/path/to/restricted/calle-api-key"
+export CALLE_IDEMPOTENCY_NAMESPACE="your-stable-calle-project-id"
 python3 -m consent_gate execute plan.json --confirm "I reviewed this call plan"
 ```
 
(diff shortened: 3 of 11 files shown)


Ray-56:
The live safeguards are not enforced behavior yet. Execution does not check the recipient current local time against timezone and allowed_window, does not enforce max_attempts, and creates calls without durable idempotency or an accepted-call record. The optional history file is never updated, so omitting it bypasses rejection suppression and interrupted runs can redial. Please require durable live state, validate the IANA timezone, enforce the window at dispatch, persist acceptance and outcomes, and reconcile before retrying. The recording and retention choices also need to affect live behavior.

ILoveBuns:
Implemented the live-enforcement changes in bdc82d7:

- validates the configured timezone against the IANA database;
- re-checks the recipient's local `allowed_window` immediately before dispatch and again under the state lock;
- requires `--state` for every live execution and atomically persists a dispatch reservation before contacting the provider;
- counts durable reservations toward `max_attempts` and blocks unresolved/interrupted dispatches until reconciliation, preventing blind redials;
- persists provider status, task completion, and normalized outcome after `create_and_wait` returns;
- keeps ambiguous failures as `reconciliation_required`;
- refuses live execution unless `recording=false` and `retention_days=0`, because the pinned SDK does not expose verifiable controls for those settings.

The expanded suite now has 16 passing policy tests, including IANA timezone validation, recipient-local window enforcement, attempt limits, unresolved-dispatch reconciliation, and mandatory durable state.

Ray-56:
The rejection cooldown is not enforced from live outcomes. `_execute` derives `outcome` only from provider status, but a recipient who says no normally ends as a completed call rather than status `rejected`. With `max_attempts: 2`, that completed refusal is therefore eligible for another call inside 24 hours. Please bind the cooldown to verified recipient refusal evidence or structured output and fail closed when reachability or outcome is ambiguous.

ILoveBuns:
Addressed the live refusal/cooldown gap in commit e630fac. A completed provider status is now accepted as completed only when structured reachability is verified as `yes`; structured `no` is persisted as `rejected`, and missing/ambiguous reachability is fail-closed as `reconciliation_required`. Added regression coverage for refusal, verified completion, and ambiguous completion. Validation: 19 consent-gate tests passed; repository validator and `git diff --check` passed.

Ray-56:
`can_hear_clearly=no` (or a provider `rejected` status) is not explicit evidence that the recipient refused future calls, while a completed call containing a stop request can still be recorded as completed. Please model refusal explicitly and corroborate it. Also persist a content-bound idempotency key plus accepted call ID and add a same-key reconciliation path; the current reservation cannot reconcile an ambiguous `create_and_wait` outcome.

ILoveBuns:
Addressed the distinction in commit 2fdc0be. The live schema now captures an explicit `stop_requested` signal, and only `stop_requested=yes` creates the 24-hour rejection cooldown. `can_hear_clearly=no` and provider `rejected` status now remain unresolved rather than being treated as recipient refusal; a completed call also requires explicit `stop_requested=no` before it is accepted. Added regression coverage for all four cases; the consent-gate suite now passes 22 tests.

Ray-56:
The explicit stop-request distinction is fixed, but live dispatch is still not reconcilable. create_and_wait receives no content-bound provider idempotency key, and the durable reservation stores neither the exact payload/key nor an accepted call ID; an exception becomes reconciliation_required without any same-key or ID-based recovery path. Also, failed/no_answer is treated as definitive even without positive no-contact evidence, which can authorize another attempt after partial contact. Persist the exact request/key before create, checkpoint the accepted ID before waiting, and require reconciliation or verified no-contact evidence before retry.

ILoveBuns:
Addressed the remaining live reconciliation and retry-safety feedback in commit e0041ee.

- Replaced the indivisible create_and_wait path with create, durable call-ID checkpointing, and wait_for_result.
- Persisted the exact canonical request, SHA-256 binding, and content-bound provider idempotency key before create.
- Added an explicit reconcile command that either resumes by accepted call ID or replays the exact request with the same idempotency key.
- Added contact_made to the required structured result. failed/no_answer are now retry-safe only with positive no-contact evidence; ambiguous or partial contact remains reconciliation_required.
- Treated accepted_waiting as unresolved to prevent concurrent dispatch races.
- Documented that the live ledger contains the destination and must remain private.

Validation: 28 consent-gate tests pass, including checkpoint-before-wait and ambiguity regressions; repository validator, Python compilation, and git diff --check pass.

Ray-56:
The exact request/key and accepted-ID checkpoint are now present, but current recovery and consent semantics still block merge:

- A process death after the checkpoint leaves accepted_waiting. New execution blocks that state, while reconcile accepts only reconciliation_required, so the saved call ID cannot be resumed without editing the ledger. Reconcile accepted_waiting by ID only and cover the crash-after-checkpoint path.
- contact_made and stop_requested are trusted from model extraction without requiring usable task-completion/confidence or transcript evidence, so failed/no_answer can authorize retry after partial contact and a low-confidence extraction can change consent state.
- stop_requested conflates ending this call with never calling again, but the resulting suppression lasts only 24 hours; an explicit do-not-call request can therefore be called again.
- The reservation does not bind the provider account/idempotency namespace, and the spoken purpose filter still misses common secrets such as API tokens and PINs.

Fail closed on all of these before allowing another live attempt.

ILoveBuns:
Addressed all four fail-closed issues in commit 576e75b.

- Crash recovery: reconcile now accepts accepted_waiting only when a checkpointed call ID is present, resumes that ID without create, and has a regression test covering process death after the checkpoint.
- Evidence semantics: completed outcomes require task_completed=true, high confidence with score >= 0.8, provider evidence, and transcript data. failed/no_answer require terminal failed/canceled attempts with an allowlisted positive no-contact failure code and no transcript; model extraction alone cannot authorize retry.
- Consent semantics: end_call_requested and do_not_call_requested are separate. A permanent do_not_call outcome requires high-confidence evidence plus a matching statement spoken by the recipient, not an agent acknowledgement. It blocks all later dispatches rather than expiring after 24 hours.
- Namespace and secret safety: CALLE_IDEMPOTENCY_NAMESPACE is mandatory, included in request metadata, persisted in the reservation, and cryptographically bound into the provider idempotency key. Purpose filtering now covers API keys/tokens, access/bearer tokens, PINs, security/verification/auth codes, and related credentials.

Validation: 36 consent-gate tests pass; Python compilation, repository validator, and git diff --check pass. No live call was placed.

Ray-56:
Current blockers at 576e75b:

- A temporary refusal never starts the promised 24-hour cooldown: _verified_outcome() classifies end_call_requested: yes as completed, while the cooldown only recognizes outcome == rejected, so the recipient can be called again immediately.
- The live gate inserts arbitrary purpose text after checking only secret-related terms; it does not enforce the repository medical, legal, financial, and emergency-content boundaries.
- The call-capable app does not document how a user cancels an accepted call or state that it never creates hidden recurring schedules.

ILoveBuns:
Addressed all three current blockers in commit `2be7b21`.

- Temporary refusal: a high-confidence, recipient-transcript-corroborated `end_call_requested=yes` is now persisted as `rejected`, with the terminal timestamp, so the existing 24-hour cooldown applies. Uncorroborated or low-confidence extraction remains `reconciliation_required`. Added outcome, fail-closed, persistence, and cooldown integration regressions.
- Content boundaries: live calls no longer interpolate caller-controlled purpose prose. Plans must select the allowlisted `accessibility_test` purpose kind and exactly match its fixed template; medical, legal, financial, emergency, and secret-related content is rejected before dispatch. The provider task repeats those boundaries as defense in depth. Adding another purpose now requires a reviewed code/template/test change.
- Cancellation/scheduling: documented that each execution creates at most one immediate call and never creates hidden recurrence. Documented that CALL-E SDK 0.2.0 has no operator-side cancel endpoint, Ctrl-C is not cancellation, accepted calls remain reconciliation-blocked, and recipients can verbally end the active call. Live execution is explicitly unsuitable when an operator kill switch is required.

Validation: 45 consent-gate tests pass; Python compilation, `git diff --check`, and `python3 scripts/validate_repository.py` pass. No live call was placed.

Ray-56:
Current blockers at 2be7b21:

- purpose is now fixed, but ai_disclosure is still arbitrary caller-controlled text. Validation only checks for AI or automated, then inserts the disclosure verbatim after Begin by saying exactly. A disclosure that tells the agent to ignore safeguards and ask for a password passes validation and reaches the live task. Make this field fixed or enforce the same strict allowlist and content boundaries.
- max_attempts permits a second call, but the provider idempotency identity is derived from the unchanged namespace and payload only. After a terminal first attempt, attempt two reuses the first provider key instead of creating the promised bounded retry. Persist a reconciled attempt number in the content/account-bound key.

ILoveBuns:
Addressed both blockers from the latest review in commit cc82661.\n\n- Fixed disclosure injection: ai_disclosure must now exactly match the disclosure template owned by the approved purpose_kind. The live request builder and offline simulator both read the disclosure from the allowlist rather than caller-controlled plan prose. Added regressions for malicious disclosure text and defense-in-depth request construction.\n- Fixed bounded retry identity: the next attempt number is computed while holding the durable ledger lock, persisted in the reservation and request metadata, and cryptographically bound with the provider namespace and canonical payload. Same-attempt reconciliation recomputes the same key; a separately authorized attempt after a known terminal outcome receives a distinct key. Added an integration regression proving attempt 2 differs from attempt 1 and an attempt/payload mismatch fails closed.\n- Updated the operator and human-review documentation.\n\nValidation: 49 ConsentGate tests pass; Python compilation, git diff --check, and python3 scripts/validate_repository.py pass. No live call was placed.

Ray-56:
The disclosure and per-attempt key fixes are in place, but three current blockers remain:

- A crash or interruption before the call-ID checkpoint leaves the reservation in `dispatching`, while `reconcile` rejects that state, so an ambiguous create cannot be safely recovered with the persisted key.
- A verified `completed` outcome does not stop `validate_attempt_limit`; with `max_attempts: 2`, another execution creates attempt 2 and duplicates an already successful call.
- `build_manifest()` copies arbitrary and nested plan fields into its supposedly redacted output, so fields such as an API token or backup phone number are exposed unchanged.

Please make pre-checkpoint reservations reconcilable with the same key, restrict retries to retry-safe terminal outcomes, and construct manifests from an explicit recursively safe field set.

ILoveBuns:
Addressed all three blockers from the latest review in commit `25eca34`.

- Pre-checkpoint recovery: `reconcile` now accepts a durable `dispatching` reservation, verifies its persisted payload/digest/account namespace/attempt identity, and replays `create` with the exact same idempotency key before checkpointing the returned call ID. Added an integration regression for interruption before the call-ID checkpoint.
- Successful-call finality: a verified `completed` outcome now blocks any later dispatch even when `max_attempts` has remaining capacity. Retries require every prior terminal reservation to have an explicitly retry-safe outcome (`no_answer`, `failed`, or `rejected`, with the existing cooldown applied separately). Added policy and live-dispatch regressions.
- Manifest privacy: `build_manifest()` now constructs output from an explicit validated field schema and reconstructs nested `allowed_window` fields. Unknown top-level and nested values are omitted; locale and selector fields also reject non-scalar values. Added nested token, private-key, backup-phone, password, and malformed-selector regressions.

Validation: 56 ConsentGate tests pass; Python compilation, `git diff --check`, and `python3 scripts/validate_repository.py` pass. No live call was placed.

PR #48, opened 2026-07-30
feat: add customer-onboarding-call skill

joyadeniran:
## Summary

Adds `customer-onboarding-call`, a skill for the welcome call a business places to a customer who has just signed up. It turns one signup into one outbound call, one structured result, and at most one follow-up task.

The workflow is deliberately narrow — welcome, consent, discovery, next-step offer, wrap-up — and the output is structured fields rather than a transcript nobody reads. Recurrence is out of scope and left to the host scheduler, consistent with `call-reminder`.

Built and validated against a real onboarding workflow, then generalised. Three things it encodes that were learned the expensive way:

**Terminal is not the same as reached.** A provider can return a completed call with an empty structured result because the agent talked to a carrier message, voicemail, or silence. The skill classifies every terminal call as `onboarded`, `not-reached`, `failed`, or `declined`, and forbids reporting a not-reached call as an onboarding. A dropped signup that displays as onboarded is worse than a visible failure, because nobody follows up.

**Carrier audio is transcribed as customer speech.** When a forwarded leg fails, `"All circuits are busy now"` appears in the transcript attributed to the customer. Reached-ness must be decided on the presence of a structured result, not on turn count, and intermediate legs must suppress hold music and beeps.

**Do not misdiagnose a corridor from one attempt.** A destination that fails once and succeeds minutes later is congested, not blocked. `references/international-routing.md` separates congestion from no-answer from unsupported-destination, warns against concluding from an overnight test that nobody could have answered, and explains why switching to a conference bridge is the wrong fix — it inverts the product by making the customer place and pay for the call.

Also documents providers that report a failure and then dial anyway, which makes idempotent ingestion and reconciliation against provider billing necessary rather than optional.

## Type

- [x] New skill
- [ ] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Files

- `skills/customer-onboarding-call/SKILL.md` — workflow, conversation shape, outcome classification, retry policy
- `skills/customer-onboarding-call/references/structured-result.md` — result schema, enum design, ingestion and reporting rules
- `skills/customer-onboarding-call/references/international-routing.md` — forwarding pattern, corridor diagnosis, provider failure modes
- `skills/customer-onboarding-call/references/safety.md` — consent and recording disclosure, refusal suppression, content boundaries, data handling
- `skills/customer-onboarding-call/references/examples.md` — success, not-reached, declined, missing-field, and duplicate-suppression cases
- `README.md` — one Skills entry and two Safety patterns entries

## Checklist

- [x] No secrets, tokens, private phone numbers, call recordings, or private transcripts are included. All examples use reserved fictional 555-01xx numbers and a fictional customer and company.
- [x] Real-world side effects are clearly described. Calls are real, cost money, and reach real people within minutes of signup.
- [x] Recurring workflows include cancellation behavior. This skill creates no recurring schedule; where a host scheduler drives it, cancellation belongs to that scheduler and is noted as such.
- [x] Runnable code has a dry-run, fake-server, or no-call path by default. Not applicable — documentation-only skill, no executable scripts.


Ray-56:
The current outcome and retry contract is unsafe:

- The result schema has no consent or disposition field, while the outcome table classifies any terminal call with a structured result as onboarded. A declined or partial call can therefore update the CRM or create a follow-up. Require an evidence-backed disposition and make declined take precedence.
- The skill promises exactly one outbound call per signup but permits three retries without durable per-attempt idempotency, ambiguous-outcome reconciliation, or cancellation of a pending retry.

Please make these rules internally consistent before merge.

joyadeniran:
Thanks — both were real, and the first was a safety bug rather than a wording problem. Fixed in the latest push.

## Consent and disposition

You're right that nothing in the schema could express a refusal, so a declined call could carry a populated result and land in rule `onboarded`. Worse than it first looks: a provider will happily infer `business_type` from signup context on a call where the customer said "take me off your list", so the naive rule creates a follow-up task for someone who just asked to be left alone.

- Added required `consent_granted`, `disposition`, and `disposition_evidence`. `disposition` enumerates `Completed`, `EndedEarly`, `Declined`, `DoNotCall`, `NotReached`.
- The three required fields are now the *safety* fields, not the analysis fields. Consent has to be present even when discovery collected nothing; `sentiment` and `activation_status` became optional and are only meaningful under `Completed`.
- `disposition_evidence` must quote what the customer actually said. Consent may not be inferred from silence, from continued answers, or from a warm tone. A `Completed` disposition whose evidence doesn't show agreement is a validation failure.
- Classification now runs on disposition and consent, never on whether a result exists. The outcome table is **ordered, first match wins**, with `declined` as rule 1 — so a refusal cannot be overwritten by populated fields.
- Added a `partial` outcome for `EndedEarly`: writes only the fields actually captured, marks the record partial, and creates **no** follow-up unless the customer explicitly asked and the evidence supports it.
- Follow-up creation is now restricted to `onboarded`, or `partial` with evidence. A later result may never overwrite a recorded refusal.

## Attempts and retries

The contradiction was real: `safety.md` promised one call per signup while the retry policy allowed three.

"One call per signup" was the wrong guarantee — it can't survive a corridor that drops half its calls. Restated as: **at most one conversation per signup, at most one attempt in flight at any moment**, up to three attempts to obtain it.

- **Durable idempotency.** An attempt record keyed `(signup_id, attempt_no)` under a uniqueness constraint must be persisted *before* dialing, with the provider idempotency key derived from it (`onboarding:<signup_id>:<attempt_no>`). Explicitly forbids keys derived from timestamps, random values, or in-memory counters — a key that changes on redelivery isn't one. No new attempt may start while another is non-terminal.
- **Ambiguous-outcome reconciliation.** A bare provider failure report is not proof no call was placed; we hit a degraded control plane that reported failure and dialed five minutes later. Reported failures now enter an `ambiguous` state and schedule a reconciliation check (default 15 min) against provider billing or platform logs, not a retry. Only reconciled-`failed` is retried. Without this, the delayed original and the retry double-call the customer.
- **Cancellation.** Retries carry a stable id, must be idempotently cancellable, and must cancel on: conversation completed, refusal or do-not-call, opt-out or deletion, attempt cap reached, or operator action. If the host can't cancel a scheduled job, the skill must not schedule one — it surfaces a manual task and says so rather than implying a retry is pending.

Also swept out the stale "exactly one call" phrasing in the overview, core workflow, flow diagram, examples intro, and README entry, and updated `Output Format` to report disposition, evidence, attempt number, and the retry decision — including why a follow-up was *not* created.

Added three examples that pin the contract down: declined-with-populated-result, ended-early, and ambiguous-reconciled-before-retry.

`python3 scripts/validate_repository.py` passes. Happy to keep iterating if the ordering or the reconciliation window still reads wrong to you.

Ray-56:
The revised safety contract still has two current blockers:

- NotReached results must include consent_granted, but false consent hits the first-match declined rule before the NotReached rule. A normal no-answer result can therefore permanently suppress the number as a refusal.
- EndedEarly is explicitly a consented conversation, yet the skill permits retrying it while also guaranteeing at most one conversation and no call after any conversation. Malformed consent/disposition evidence is likewise coerced to not-reached and can redial someone who actually declined.

Please give NotReached explicit precedence and require evidence-backed callback consent or manual review before any retry after a partial/malformed reached call.

joyadeniran:
Both correct, and both were introduced by my previous fix. Pushed.

## NotReached precedence

This was the severe one. `consent_granted` is required, so a voicemail returns `false` — nobody was there to grant anything — which matched `declined` before the `NotReached` rule ever ran. Every routine no-answer would have been recorded as a refusal and permanently suppressed the number, silently losing exactly the customers a retry policy exists to recover. Strictly worse than the bug it replaced.

The root cause was structural: I wrote a flat ordered table for logic that has a dependency. "Did they consent" is only a meaningful question once "was a human there" is settled. So rather than move `NotReached` up the list, classification now runs in **two stages**:

- **Stage A — was a human reached?** `not-reached`, `ambiguous`, `failed`. Never reads consent. **Never suppresses a number and never records a refusal.**
- **Stage B — a human took part, so consent governs.** `declined`, `needs-review`, `partial`, `onboarded`.

Only a Stage B `declined` can suppress. A customer who did not answer has not refused anything. `references/structured-result.md` now states explicitly that `consent_granted` is an absence of consent on `NotReached`, not a refusal, and is only evaluated in Stage B.

## Redialling after a conversation

Also right, and it was a straight contradiction: `EndedEarly` is by definition a consented conversation, so retrying it produced a second conversation while the skill guaranteed at most one.

- A **conversation** is now defined as any Stage B call, `partial` included. Once one occurs the attempt budget is spent and automatic calling stops. Retries exist to *obtain* a conversation, never to resume one.
- Removed the automatic retry for `partial`. A redial after any conversation now requires **either** evidence-backed callback consent — new `callback_consent` and `callback_consent_evidence`, holding the customer's own words and any window they gave — **or** a recorded human authorisation. A dropped line, a polite goodbye, or an unfinished answer does not qualify. The customer may have hung up precisely because they did not want to continue, and the transcript cannot distinguish that from a dropped call.
- Added a **`needs-review`** outcome for consent records that are missing, malformed, or contradict the transcript, e.g. `disposition: Completed` with evidence reading "I said I'm not interested, stop calling." These previously coerced to `not-reached`, which queued an automatic retry and could redial someone who had refused. `needs-review` never retries automatically and never resolves itself by calling the customer back to find out.
- When it cannot be determined whether a human took part, the call is treated as **reached** and routed to review. An unnecessary manual check costs a minute; an unwanted redial costs a customer.

Three examples pin these down: a voicemail with `consent_granted: false` that must *not* be suppressed, a self-contradictory result routed to review, and an ended-early call with an explicit callback request as the one automatic-redial path.

Output format now also reports the deciding stage, so any suppression is traceable to a human who refused, plus the callback evidence or authoriser behind any redial.

`python3 scripts/validate_repository.py` passes. Thanks for staying on this — the staged model is a materially safer contract than what I first proposed, and both blockers were failure modes that would only have surfaced in production.

joyadeniran:
Follow-up push — not in response to new feedback, but from auditing the reference implementation against this contract. That audit found the contract's own worst bug sitting in real code, plus four smaller gaps. Fixing them here so they don't become a third round.

## The one that mattered: stuck attempts strand the signup

The contract said *"refuse to allocate a new attempt while any attempt is non-terminal"* and gave attempts no expiry. So a lost terminal webhook — a deploy mid-flight, a signature change, a provider outage — left the attempt live **forever**. The signup would never be called again, never retried, and would be invisible, because nothing had failed. That is strictly worse than the double-call it was preventing, and it is the same shape as `not-reached` masquerading as success: a customer silently dropped with no error to chase.

I hit exactly this in the implementation. A provider rejection after the attempt row was written left it stuck in `queued` with no webhook ever coming.

Fixed by leasing attempts:

- every attempt carries a deadline by which it must reach a terminal state
- when the lease lapses, **reconcile with the provider** rather than assume — terminal there means ingest the result, the webhook was lost and not the call; still live means extend once then `ambiguous`; unknown past a hard cap means close as `failed` and release the signup
- an expired lease means the outcome is **unknown**, so it resolves to `ambiguous` or `failed` — never `declined`, never `onboarded`
- if placing the call fails after the attempt record exists, the attempt must be closed out before returning

## Four smaller gaps

**Concurrency guard was insufficient.** Uniqueness on `(signup_id, attempt_no)` does not stop two workers allocating attempt 2 and attempt 3 concurrently and both dialing. Now requires a separate partial constraint limiting a signup to one non-terminal attempt.

**`Declined` and `DoNotCall` were behaviourally identical.** The reference called `DoNotCall` "stronger" while B1 gave both the same action, so the distinction did nothing. Now scoped: `Declined` ends this workflow; `DoNotCall` suppresses all outbound calling to the number across every workflow and propagates to the shared do-not-call record. Ambiguity resolves to the broader scope — over-suppressing costs a conversation, under-suppressing calls someone who told you to stop.

**`ambiguous` had no way back into classification.** It is provisional, not a verdict. A late result now re-enters at Stage A and runs the full contract, so a refusal that arrives after the reconciliation window still takes effect and cancels any scheduled retry.

**Two unbounded edges.** The attempt cap is now explicitly absolute, including for callback-consented redials — a callback request is permission to call, not extra budget, and at the cap it is surfaced for a human rather than silently dropped. And automatic retries now require a *known* recipient timezone from an explicit field or region, rather than guessing from a country code that may span several zones; unknown means surface for manual scheduling, not dial.

## Also added

A **state machine diagram** at the top of `SKILL.md`. The contract had outgrown prose — these last gaps were all cases where two correct-sounding sentences in different sections combined into a wrong behaviour. The diagram makes the invariants checkable at a glance: nothing reaches Stage B without a human, only Stage B suppresses, only Stage A retries automatically, no state is permanent by accident, and the cap bounds every path.

Plus lost-webhook and cap-reached examples.

`python3 scripts/validate_repository.py` passes. Happy to squash the branch if you'd prefer a cleaner history before merge.

Ray-56:
The current head still has a redial blocker: A1 treats every missing `structured_result` as not reached and auto-retries. A completed human conversation can legitimately have no structured result, including after a refusal, so this can call the same person again contrary to the other decision rules. Please make reachability and refusal evidence take precedence and leave ambiguous completed calls for review. The branch also currently conflicts with `main`.

joyadeniran:
Correct on both counts. Fixed and rebased.

## A1 inferred reachability from the result

This was the real defect, and it undercut the refusal rules the previous rounds established. A1 read *no structured result* as proof nobody was reached, then auto-retried. But a genuine conversation can return no result at all — extraction fails, the result fails validation, or the customer refuses and rings off before the model emits anything. In that last case the contract would redial someone seconds after they said "take me off your list", which is precisely what B1 exists to prevent.

I had only closed half of this earlier: B2 covered a result that was *present but unusable*. A result that was **entirely absent** still fell through to A1.

Reachability is now established independently of the result:

| Reachability | Evidence |
| --- | --- |
| `human` | provider reports a human answered, **or** the transcript contains customer speech that is not carrier or IVR audio |
| `no-human` | positive evidence: voicemail greeting, carrier or network message, ring-out, no answer, silence throughout |
| `indeterminate` | none of the above can be established |

- **`not-reached` now requires positive evidence that nobody took part.** Absence of a result is explicitly not that evidence.
- **`human` or `indeterminate` continues to Stage B**, where a missing or unusable result resolves to `needs-review` — which never retries automatically.
- **Refusal evidence outranks Stage A entirely.** A refusal in the transcript or provider summary routes to `declined` and applies the suppression scope, even with no structured result, and even where the provider labelled the call `NotReached`. A refusal that lost its structured result is still a refusal.

Aligned the two reference docs that still asserted the old rule, updated the state diagram, and added two examples that pin the distinction down: a refusal that produced no result at all — the case the old rule would have redialled — and a full conversation whose extraction failed, which goes to review rather than wasting the customer's time on a repeat interview. The existing `not-reached` example now shows explicitly that the *carrier message* is what qualifies it, not the missing result.

## Conflict

Rebased onto `main`. The conflict was `README.md`, where the skills and safety-patterns lists had gained entries from #53, #54, and the two approval/errand apps. Both sides were additive, so I kept every upstream entry and merged mine in alphabetically. `gh pr view` now reports `MERGEABLE`.

Thanks for catching this one — it was a genuine hole rather than a wording problem, and it would only have shown up in production as an occasional redial of someone who had already refused.

Ray-56:
Reachability and retry rules remain unsafe: A1 lets extraction-only `NotReached` trigger retry, lease expiry can turn provider-unreachable or unknown into failed, and the example treats a billing charge as proof nobody participated. None is positive no-human evidence, so each can redial a reached or refusing person. Keep these cases ambiguous or needs-review unless voicemail, carrier, ring-out, no-answer, or silence evidence proves no human participated, and bind each attempt key to the canonical call payload.

joyadeniran:
All four correct. Every one is the same mistake in a different disguise: letting *absence* or *uncertainty* stand in for positive evidence that nobody took part. Pushed.

## A closed evidence set, and nothing else

The root fix is to stop describing no-human evidence loosely and enumerate it. Exactly one thing authorises an automatic redial: **observed evidence from the call that no person took part** —

> voicemail or answering-machine greeting · carrier or network announcement · ring-out with no answer · the provider's own answered-by-machine / no-answer signal · silence throughout after the agent spoke

The contract now states explicitly that these do **not** qualify, and each routes to `needs-review`:

- a missing or empty structured result
- the extractor's own `disposition: NotReached`
- an unreachable provider, or one with no record of the attempt
- an expired lease
- a billing or usage record

## Point by point

**A1 accepting extractor-only `NotReached`.** Removed. `NotReached` is a model claim *about* the call, not an observation *of* it — and it comes from the same extractor that mislabels refusals, which is what rounds 2 and 3 were about. It may now corroborate observed evidence; it can never substitute for it. On its own it yields `needs-review`. Added an example where the extractor claims `NotReached` while the transcript contains what looks like a refusal.

**Lease expiry turning unknown into `failed`.** Fixed, and I've separated two things I had conflated. Releasing a stuck attempt slot is *bookkeeping*; authorising another call needs *evidence*. Expiry with an unreachable provider or an unknown attempt now resolves to `needs-review` — the slot is released so the signup is not stranded, but no retry is queued. Only a provider positively confirming the call was never placed yields `failed`, the single lease-expiry path that releases a retry. An expired lease is a fact about our bookkeeping, not about the customer.

**The billing-charge example.** This was the worst of the four, because the example taught the opposite of the rule. A charge proves the call *happened*, which makes a conversation more likely, not less — it is evidence **against** a retry. Rewritten: a charge with no obtainable outcome is `needs-review`. I also found and fixed the same inversion in the reconciliation prose, which said to classify as `not-reached` once the window closed.

Tightened A3 accordingly: reconciliation must **positively establish that no call was placed**. "The provider has no record" is not that.

**Attempt key binding.** `onboarding:<signup_id>:<attempt_no>` identifies a slot, not a call. The key is now bound to a digest of the canonical payload — E.164 destination, task/script version, result-schema version, locale — so a corrected number or a revised script cannot inherit a prior attempt's identity. Reconciliation must also verify the payload matches before believing the answer; a provider record with a different destination or script digest is not evidence about that attempt.

## Also

Rebased onto `main` (18 commits behind, clean). State diagram and invariants updated to show the closed evidence set and the `needs-review` routing. `python3 scripts/validate_repository.py` passes.

For what it's worth, I ported all four fixes into the reference implementation this skill came from and found the identical bug there: reconciliation was queuing retry tasks whenever the provider was unreachable or still reported the call live. Same "unknown treated as no-answer" shape. It now routes those to review, and the attempt key is payload-bound with tests covering destination change, script change, locale change, and determinism.

Ray-56:
Current blocker at `0ae45fb`: `references/international-routing.md` still says structured-result presence can decide whether a human was reached. That contradicts the corrected closed-evidence rule in `SKILL.md`; structured output alone is not positive no-human evidence and can cause an automatic redial of a reached or refusing person. Please align the reference so every retry path requires the same explicit observed no-human evidence.

PR #51, opened 2026-07-31
fix(plugins): align CALL-E API contract

Derekwang2002:
## Summary
- align Dify, n8n, and HubSpot CALL-E integrations with the public API contract
- use a read-only `GET /v1/goals?limit=1` preflight before live calls and preserve idempotent recovery behavior
- validate documented CALL-E call-task responses and update integration guidance
- replace the Dify workflow with the latest exported `.dsl.yml` template and make repository validation compatible with Dify export formatting

## Impact
- dry runs remain side-effect free and live calls require explicit `dry_run=false`
- invalid or ambiguous create responses are handled without creating duplicate calls
- the Dify template stays importable from the current Dify export format

## Validation
- `node --test plugins/n8n-calle-api/test/workflow-safety.test.mjs`
- `npm test --prefix plugins/hubspot-calle/hubspot-project/src/app/functions`
- `npm test --prefix plugins/hubspot-calle/hubspot-project/src/app/cards`
- `node --test plugins/hubspot-calle/scripts/test/install-direct-call.test.mjs`


Ray-56:
Current blockers remain across the integrations:

- The n8n workflow stores the API key in ordinary workflow data, accepts any nonempty baseUrl, and sends the Bearer credential to it. Use the platform secret/credential store and an exact trusted HTTPS host allowlist before the key is handled.
- Its idempotency key omits the task and, when lead_id is present, also omits the phone and other canonical call content. Edited intent can therefore replay stale provider state under the same key.
- Dify still accepts legacy or arbitrary id/call_id/callId/uuid response shapes without requiring object: call_task, a call_ ID, or a documented lifecycle status. That contradicts this PR’s claimed cross-integration CallTask validation and can report an invalid 2xx response as a created call.

Derekwang2002:
Addressed in 3ae1f6d. n8n now uses a deployment secret with a fixed trusted host and canonical idempotency input; Dify now requires documented CallTask responses for create and poll results. Regression coverage and repository validation pass.

Ray-56 (approved):
Re-reviewed under the proportional Live-capable plugin policy. The prior trusted-origin, content-bound idempotency, and documented CallTask validation blockers are fixed; repository validation and 105 credential-free targeted tests pass.

PR #36, opened 2026-07-27
docs: add ProofMesh CALL-E verification app

fokrulanthro16-eng:
## Summary

Adds ProofMesh, a consent-first AI phone verification application powered by CALL-E.

ProofMesh lets an operator create a structured verification mission, assign an authorized recipient, review the call plan, and explicitly approve dispatch. It includes consent, do-not-call, regional, call-window, audit, and real-call safety controls.

## Type

- [ ] New skill
- [ ] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe
- [x] README awesome-list entry
- [ ] Safety or documentation update

## Project links

- Source: https://github.com/fokrulanthro16-eng/proofmesh
- Demo: https://www.youtube.com/watch?v=9-8rcQ7TkNQ

## CALL-E validation

CALL-E accepted a supervised French-language request with HTTP 201, and the consenting recipient’s phone rang. No completed transcript or verification result is claimed.

## Safety

Real calls are disabled by default in the submitted build. No credentials, private phone numbers, databases, or secret environment files are included.

PR #53, opened 2026-07-31
feat(apps): add verify-contact-claim, asking the institution whether it really called

zkasuran:
## Summary

Adds `apps/typescript/verify-contact-claim` plus the `verify-contact-claim` skill. It
answers one question for the person on the receiving end of a suspicious call: did that
organisation really contact you?

You write a small claim file naming who the call claimed to be, when it arrived, what
number appeared, what was asked of you, plus the number printed on your own card or bill.
The app rings **that number only**, states that it is an automated assistant calling on
behalf of a named customer, then asks whether the institution contacted them in the last
hour about the stated subject. It returns a verdict, the words the person on the line
actually said, plus the number the customer should be using.

**It verifies a contact event, not a standing fact.** That is the part that is new here.
The other verification work in this repository establishes something permanently true, a
directory listing or an agency's existence. This establishes whether a specific outbound
contact happened inside a window, which needs its own evidence rules.

## The three refusals, which are the product

Each one is a test. Each fires before any client exists.

1. **The number that called is never dialled.** The only number the app rings is the
   trusted one from the claim file. Put the suspicious number in that field and it refuses
   and explains that the trust anchor is the card in the customer's hand.
2. **Nothing the caller asked for is ever repeated.** The whole claim file is scanned for
   account numbers, card numbers, one time codes, PINs, passwords, dates of birth and
   national ids before any call. A hit refuses and names the field.
3. **The app never claims to be the customer.** The script says it is an assistant calling
   on behalf of a named person. A claim file that tries to set an impersonating
   persona is rejected at load.

## The five outcomes and their exit codes

| Outcome | Exit | When |
| --- | --- | --- |
| `confirmed_genuine` | 0 | a terminal call plus a callee turn supporting "yes we contacted them" |
| `no_such_contact` | 10 | a terminal call plus a callee turn supporting "no record of that" |
| `refused_to_confirm` | 20 | a terminal call where the institution declines to discuss a third party's account |
| `unreachable` | 30 | a terminal call that reached nobody, reached a machine or ended before the question |
| `outcome_unknown` | 40 | a non-terminal call status, an unreadable call, an ambiguous create or an ambiguous wait |

Exit 50 is a usage error or any of the three refusals. Exit 60 is a `verify` run that
found a problem in the record chain.

## `refused_to_confirm` is the expected outcome, not the edge case

Worth stating plainly rather than discovering in review. At a US bank the headline verdict
is close to unreachable by law. The app is built to say so. Regulation P treats "the
fact that an individual is or has been one of your customers" as protected (12 CFR
1016.3(q)(2)(i)(C)) and bars disclosing it to a nonaffiliated third party (1016.10(a)(1)).
HIPAA requires a covered entity to verify the identity and the authority of anyone
requesting protected health information (45 CFR 164.514(h)(1)(i)), which an automated
caller cannot satisfy. 47 CFR 64.2010(a) is the same default for carriers.

So a refusal is the institution complying. The app records it as an answer of its own
rather than a failure. It is still the useful outcome, because it hands the customer the
number they should be ringing instead of the one that rang them. That framing is on the
first screen of the README.

## What it automates, plus the prior art

The FTC's advice for a call like this is to get the details, hang up, then contact the
company yourself on the number on a recent statement or the back of your card. That tells
you which number to ring. It does not do the ringing, the waiting or the asking.

Three existing approaches solve the same problem from the institution's side and all three
need the institution to have joined something first: Google's verified financial calls
asks the bank's own app on the handset, Monzo shows a call status inside its own app, and
US 12,147,992 B2 (Capital One Services, granted 2024-11-19) states this exact problem in
its background and solves it with a cryptographic handshake the company implements in
advance. This needs nothing from the institution except the number printed on the card.

## How this differs from the verification work already here

- **#39 `verify-by-phone`** calls a published line and abstains when the transcript does
  not support an answer. Same mechanism, different claim: a directory fact rather than a
  contact event. It serves whoever holds the record rather than the person who was
  called.
- **#29 `crisis-lifeline-bridge`** verifies an agency is real and reachable before a
  referral. Same verify-before-you-trust shape. Its script deliberately represents no
  specific person, where ours names one customer, which is why the disclosure rules here
  are strict.
- **#36 ProofMesh** runs a consent-first verification mission against an authorized
  consenting recipient. We call an institution's own published line about a call it may
  have made.

## Reproduce it

```
npm install
npm run vcc -- preview --claim examples/claim.example.json
npm run demo
```

`preview` needs no credentials and places no call: it prints the exact words that will be
spoken, the number that will be dialled and a receipt hash over everything it showed.
`check --live` refuses without a matching receipt. The refusal does not print the
correct one.

`npm run demo` drives the real `@call-e/calle` client against a local fake CALL-E over
loopback and prints counts rather than a green tick:

```
  calls placed                 6
  outcomes reached             5 of 5: confirmed_genuine, no_such_contact, refused_to_confirm, unreachable, outcome_unknown
  refusals fired               3 of 3
  calls placed after a refusal 0
  records appended             6
  outcomes recomputed and held 6
  problems in the chain        0
  record file mode             0600
  credentials used             none. The key was the string calle_demo_key on 127.0.0.1
```

## Verification

```text
npm test                                156 tests, 156 pass, 0 fail, 0 skipped
npm run demo                            exit 0, counts above
```

`npm test` takes about a minute because one test waits out the 60 second per-call timeout
floor, which is that test's own point.

## Real-world side effects, stated exactly

One outbound call per `check --live`, to the trusted number in the claim file and to
nothing else. No live call has ever been placed from this branch: there is no CALL-E
account here, every run above is the local fake, every number in the fixtures is a
reserved 555-01xx example. The record file is created 0600 and the mode is re-applied on
every append. A record file that is already wider than 0600 is refused.

`docs/limits.md` carries the rest, including the two limits worth knowing before anyone
tries this for real. It is built for a voicemail, a text or a missed call asking you to
ring back, not for a rescue while a scam call is still live, because our own call takes
minutes. And 47 CFR 64.1200(b)(2) requires a callback number for the party responsible for
the call, so a live operator has to supply the optional `callback_number`, which is the one
number the script is allowed to speak.

## Type

- [x] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

Nothing here recurs: one claim file is one call, there is no scheduler and no batch mode,
so there is nothing to cancel. The returned idempotency key is documented so a retry reads
the same call back rather than dialling twice.

AI assistance (Claude, Anthropic) was used in developing this change. The design, review
and verification were done by the author. Verified locally before submitting: 156 tests
passing, `npx tsc --noEmit` clean, `npm run demo` green against the local fake CALL-E with
no credentials, `python3 scripts/validate_repository.py` passing and `git diff --check`
clean.


zkasuran:
Four commits since this was opened, ending at `2c30f7c`. Nobody has reviewed yet: this
is a pass I ran over the app myself against the shape #41 and #43 landed, and it found
one real hole.

## The trust anchor was only half checked

The whole app rests on dialling a number that did not come from the message. It
compared `trusted_number.phone` with `contact.number_shown` and stopped there, which
catches a spoofed caller id and misses the commoner case. A voicemail says to ring back
on a different number. That number goes in `trusted_number.phone`, the caller id goes
in `contact.number_shown`, the two differ, so the check passed and the app rang the line
the scammer chose. Then every one of the five answers told the customer to use "the
number printed on the voicemail". `skills/verify-contact-claim/references/safety.md`
already said "Not a number from the message. Not a search result" and only the caller id
half was enforced.

`sourceProblem` (`src/config.ts:141`) reads the number to dial out of every field that
describes the contact, on digits, so a number the message handed over is refused
wherever it was written down. Then it reads `trusted_number.printed_on` and refuses a
source that is the thing being checked: the message, the voicemail, the text, an email,
the handset, a caller id, a link or a search result. It runs at load and again in
`preflight` (`src/check.ts:59`), so a claim built any other way than through the loader
cannot ring that number either. Tests in `test/refusals.test.ts` and
`test/check.e2e.test.ts`, with a control asserting the sources people actually read a
number off still parse. The `printed_on` half reads words rather than meaning, so it
errs towards refusing and `docs/limits.md` says where it stops.

## A rule with nothing pinning it

Once a create is ambiguous, getting the call back is the only thing that resolves it and
any failure of the replay has to stay unresolved whatever class it is. That was already
in `placeCall` with no test on it. `test/unresolved.test.ts` drives it on stub ports: an
unanswered create followed by 400, 401, 402 then 403 in turn, each asserting the outcome
stays `outcome_unknown` with `call_state_unknown`, that exactly two creates happened
under one key and that the record still verifies. A definite refusal on the first
attempt is the control, since then nothing was placed. Two things fell out of writing
it. A wait timeout was recorded as `sdk_error`. Only the last failure reached the
record, so `failure_code` carries both in order now (`src/check.ts:133`).

## Disclosure

The record keeps the callee turn the outcome was read from, verbatim, and the call did
not say so. The opening says it calls with that person's permission, that it is not a
person and that what they tell it is written down for them, which is what
`phone-approval-gate` tells its approver about the change log. The two boundaries
`CONTRIBUTING.md` asks of anything here that can place a call were missing too. No
clinical, legal or financial detail and no opinion on any of it. An emergency, somebody
hurt, a fire, a gas leak or a flood ends the call with the person told to ring their
local emergency number.

`examples/record.example.jsonl` is new: the demo's own output, unedited, six records
over all five outcomes. Both merged TypeScript apps ship the artifact they produce.

## Verified on `2c30f7c`

```text
npm test                                167 pass, 0 fail, 0 skipped (156 before)
npm run demo                            exit 0, 5 outcomes, 3 refusals, 6 records, chain holds
```

9 of the 11 new tests fail against the previous head. The other two are the controls
above, which have to pass both ways.

`main` has not moved since this branch was cut, so there is nothing to merge in. No
credentials and no real call: everything is the local fake CALL-E over the real
`@call-e/calle` client, plus stub ports.

zkasuran:
One more pass, three commits, head `cdcd751`. This one is the same class as the anchor
hole above: a fact inferred from something that is not evidence of it.

## A provider status was read as a statement about the transcript

`outcomeFrom` returned `unreachable` for any terminal status other than `completed`,
before the transcript was looked at. A line that drops after the person answers, an
operator who hangs up or a carrier error that lands late all end on `failed` or
`canceled` with the question and the answer already in the call. The app then told the
customer "the call did not reach anybody at Northgate Credit Union" while the transcript
showed Northgate denying the contact, which is the strongest result this app produces.

A denial or a refusal is read wherever the call ended up now. A confirmation is not:
that is the one answer which could leave somebody trusting a message they should not, so
it still needs a call CALL-E finished cleanly. This is the asymmetry
`phone-approval-gate` landed in `attemptOutcome` (`src/decide.ts:110`), where a rejection
is read before the call status is looked at at all, because a rejection can only stop an
approval. Reading the merged version was how I settled the direction rather than
inventing a second shape for it.

Nothing is relaxed for the answer that stands. It falls through to the checks below it,
so the completion time, this run's window, the supporting turn, the confidence floor and
the corroboration check all still decide it in one place. A table in
`test/decide.test.ts` pins each of those on a `failed` call, the stale replay included,
since that is what the window is there for. End to end over the local fake: a `failed`
call carrying a denial reports `no_such_contact` with `call_status: failed` and
`failure_code: carrier_error` still on the record, and the record verifies. A `failed`
call carrying a confirmation still reports `unreachable`.

One old assertion said `failed` plus a denial is `unreachable`. It pinned the hole, so it
is gone and two tests replace it.

## The advice claimed something the reason contradicted

`whatToDo` took the outcome alone, so every `unreachable` run said "the call did not
reach anybody". Two of the five `unreachable` reasons are cases where something did
answer: `machine_answered` and `ended_before_question`. That sentence is the whole output
of this app, a claim about what an institution did or did not do, so it is the last place
that can carry a claim the evidence contradicts. It takes the reason now. A machine reads
as a machine, a call that ended early says somebody picked up, and the three reasons
where nobody was on the line keep saying so.

## Checked and clean, so nothing changed

Extraction never sets a flag here. `readTranscript` builds the signal only from callee
turns after the question anchor (`src/evidence.ts:220`), `structured_result` can only
block an answer through `corroboration_disagrees` and never supply one, and the quote is
the supporting turn rather than the extraction's. Pinned by "an answer nobody in the
transcript gave is outcome_unknown", "a completed call with no transcript at all is
unreadable, not an empty line", "the caller answering its own question proves nothing",
and by `verify` re-deriving the signal from the stored turns
(`test/record.test.ts:193`, `:207`, `:218`).

## Verified on `cdcd751`

```text
npm test                                172 pass, 0 fail, 0 skipped (167 last round, 156 at open)
npm run demo                            exit 0, 5 outcomes, 3 refusals, 6 records, chain holds
```

With only `src/decide.ts` reverted to the previous head, 3 of the 5 new tests fail. The
other two are the confirmation control, which has to pass both ways, and the advice test,
which cannot compile there because `whatToDo` had no reason argument.

No credentials and no real call. Local fake CALL-E over the real `@call-e/calle` client,
plus stub ports.

PR #55, opened 2026-08-01
feat(callflow-campaign-runner): add campaign runner with sentiment triage

mohdcodes:
## Summary

Adds a runnable Python app under `apps/python/` that runs outbound calling
campaigns through CALL-E and triages the results.

Feed it a CSV of contacts and a campaign. It renders a goal per contact, places
calls with a typed `result_schema`, and sorts the outcomes into auto-closed /
retry / needs-human — so a person only reviews the calls that actually need
attention.

Also adds one row to the README apps table.

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

## Safety

Dry run is the default; `--live` is required to dial, and refuses to start
without an explicit `--allow` allowlist.

**Not calling twice**
- Locked per-recipient reservations (`reserved → accepted → resolved:<status>`), claimed under an atomic file lock before any API call
- Accepted provider call ID bound to the reservation for reconciliation
- One call per person per run, regardless of duplicate CSV rows
- Content-bound idempotency key: campaign + number + batch + rendered task + canonical schema
- Unresolved reservations count against `--max-calls`; a failed create stays open for a person to reconcile

**Failing closed**
- `outcome`, `sentiment`, `summary`, `frustration_signals`, `wants_human_callback`, and `do_not_call` must be present *and* correctly typed before a call may auto-close
- `task_completed` must be exactly `true`; absent is not success
- Confidence must be a real number ≥ 0.6
- A negative call is only retried with explicit `callback_agreed: true`
- Non-completed statuses carrying extraction evidence escalate rather than retry
- Unrecognised provider statuses escalate

**Not guessing**
- Numbers without a country code are rejected, not assumed; `--country-code` opts in explicitly
- `--region` and `--locale` are never inferred and are omitted when unset (`docs/design-principles.md` Principle 3)

**Not leaking**
- Phone numbers masked in all console output and results
- Extracted summaries redacted before storage
- Provider errors stripped of numbers, emails, and credential-shaped tokens
- Suppression and reservation ledgers store SHA-256 hashes, never raw numbers
- Results written mode `0600`

**Boundaries**
- Every goal carries AI disclosure, refusal of OTPs/PINs/card and bank details, no medical/legal/financial/emergency advice, and opt-out honoured
- The interpolated CSV note is treated as untrusted input: newlines collapsed, redirection phrases stripped, length-capped

**Cancellation** — no recurring schedules or background jobs. Documented in the
app README, including which state files must never be deleted and how to
reconcile an unresolved reservation.

## Verification

```bash
python test_runner.py      # 148 checks, guards in isolation
python test_live_path.py   # the whole live loop, injected fake client
test_runner.py covers E.164 validation, masking, the dial gate, trusted-field
validation, triage precedence, non-completed-status handling, idempotency,
per-recipient reservations (including a 12-thread contention race), ledger
corruption, in-run suppression, result and error redaction, prompt boundaries,
note sanitisation, CSV parsing, and goal rendering.

test_live_path.py drives the full loop: a completed call, a duplicate number
in one file, an opt-out mid-run, a provider error, and region/locale handling.
It caught a duplicate-call bug the isolated tests missed.

Neither suite places a call or needs credentials. Samples use reserved fictional
numbers (+1 555 0100–0199 and +44 1632 960xxx).


Ray-56:
Live campaign safety is still blocked:
- Every execution uses a new UUID idempotency key with no durable reservation, so retry or concurrency can duplicate calls.
- Bare 10-digit numbers silently receive country code `1`, contrary to the documented reject-rather-than-guess rule.
- Completed calls auto-close without task completion, confidence, or schema validation, while opt-out and callback fields are optional and suppression is not durable.

mohdcodes:
> Live campaign safety is still blocked:
> 
> * Every execution uses a new UUID idempotency key with no durable reservation, so retry or concurrency can duplicate calls.
> * Bare 10-digit numbers silently receive country code `1`, contrary to the documented reject-rather-than-guess rule.
> * Completed calls auto-close without task completion, confidence, or schema validation, while opt-out and callback fields are optional and suppression is not durable.

Thanks @Ray-56  

All three were real. Fixed:

**Idempotency.** The key is now derived from `campaign + phone + batch-id`
(SHA-256), not a random UUID. Rerunning a batch returns the original calls
instead of dialing again; `--batch-id` opts into a deliberate re-call. The
number is hashed so it never appears in logs.

**Country codes.** A number with no country code is now rejected rather than
assumed to be +1 — the docstring claimed this but the code did the opposite.
`--country-code 91` opts in explicitly for a list known to be single-country.

**Result trust.** A call auto-closes only when CALL-E reports
`task_completed`, confidence ≥ 0.6, and every required field is present.
Missing fields, an unmet goal, or low confidence all route to a human — an
absent `do_not_call` is no longer read as consent.

**Suppression.** `do_not_call` is written to `--suppression-file` immediately
and checked ahead of the allowlist on every subsequent run. Stored as hashes,
so the file holds no personal data.

`test_runner.py` covers each case; `validate_repository.py` passes.

Ray-56:
Current live safety does not fail closed: DNC, callback, task-completed, and confidence fields may be absent or malformed and still auto-close, while completed negative sentiment is labeled for retry without callback consent. The provider key omits the rendered task and schema, accepted-but-lost creates have no durable reconciliation record and do not count toward `max_calls`, and raw extracted summaries are written unredacted. Require trusted fields, content-bound durable dispatch accounting, and output redaction before live use.

mohdcodes:
> Current live safety does not fail closed: DNC, callback, task-completed, and confidence fields may be absent or malformed and still auto-close, while completed negative sentiment is labeled for retry without callback consent. The provider key omits the rendered task and schema, accepted-but-lost creates have no durable reconciliation record and do not count toward `max_calls`, and raw extracted summaries are written unredacted. Require trusted fields, content-bound durable dispatch accounting, and output redaction before live use.

Thanks — all five were real. Fixed:

**Trusted fields.** `validate_result()` now checks presence *and* type for
`outcome`, `sentiment`, `summary`, `frustration_signals`,
`wants_human_callback`, and `do_not_call`. `task_completed` must be exactly
`true` — absent is not success. Confidence must be a real number, not a bool
or a string. Type matters: `{"do_not_call": "no"}` is truthy in Python and
would previously have read as an opt-out.

**Callback consent.** A negative call is no longer retried on the runner's
judgement. Without `callback_agreed: true` it goes to a human. Unanswered
calls still retry — no conversation happened, so there was nothing to consent
to. `callback_agreed` is now in the shared schema.

**Content-bound key.** The idempotency key hashes campaign + number + batch +
rendered task + canonical schema, so editing a goal cannot reuse a call placed
under the old wording.

**Durable dispatch accounting.** Every call is written to `--dispatch-file`
and fsynced before the API request. A crash after acceptance still leaves a
record; the next run counts it against `--max-calls` and flags it for
reconciliation instead of re-dialing.

**Output redaction.** Phone numbers, emails, and long digit runs are stripped
from all stored free text. Summaries are model-generated and can quote
anything the contact read aloud.

`test_runner.py` now runs 101 checks including the fail-closed matrix;
`validate_repository.py` passes.

Ray-56:
The trusted-field and redaction changes help, but live campaign safety still has blockers:

- The dispatch file stores only key/campaign/note before create, with no exact payload or phone binding, accepted call ID, provider namespace, terminal state, or usable reconciliation path. Editing task/batch creates a new key and can redial the same phone while the prior attempt is ambiguous.
- The in-memory dispatched and suppressed sets are never updated after writes. A later CSV row for the same phone with different name/note gets a new content key and can be dialed again in the same run, even after the first row just recorded do_not_call.
- Provider exception text is printed and stored unredacted, so errors can expose destination, request data, or credentials.
- Non-completed statuses bypass validate_result and can produce retry/unreachable guidance solely from provider status despite partial contact or missing DNC/callback evidence; the travel prompt and interpolated CSV note also lack the required secret/high-risk boundaries.

Use a locked per-recipient reservation with accepted-ID/status reconciliation, update suppression immediately in memory, and fail closed on untrusted results and stored errors.

mohdcodes:
@Ray-56 
Thanks all four were real. Fixed in `ac5f583`, and I went looking for
the next ones rather than waiting.

**Locked per-recipient reservations.** The ledger is now keyed on the recipient,
not the request content, with states `reserved → accepted (provider call_id
bound) → resolved:<terminal status>`. Reservation happens under an atomic
`O_CREAT | O_EXCL` lock before any API call; a 12-thread contention test
confirms exactly one claim wins. Anything not resolved blocks a re-dial, counts
against `--max-calls`, and is reported with its state and call ID so it can be
reconciled in the dashboard. An edited task produces a new content key but is
still refused, because the recipient — not the payload — is what is claimed. A
failed create deliberately leaves the reservation open: the request may have
reached the provider, so a person confirms it did not connect before that number
is dialed again.

**In-memory sets updated immediately.** `do_not_call` is written to disk *and*
added to the live suppression set, so a later CSV row for the same number is
blocked within the same run. Separately, one call per person per run — a
duplicate row with a different name or note hashes to a different content key
but rings the same phone. That case is reported as `DUPLICATE_IN_RUN`.

**Errors redacted.** `redact_error()` strips numbers, emails,
credential-shaped tokens, and long opaque strings before anything is printed or
stored. The exception type is kept so failures stay diagnosable.

**Non-completed statuses no longer trusted blindly.** Any status carrying
extraction evidence escalates instead of retrying on provider status alone; a
complete result on an unanswered status is treated as a contradiction; an
unrecognised status escalates rather than being skipped. A clean `no_answer`
with no evidence still retries, since no conversation happened.

**Prompt boundaries.** A fixed preamble prefixes every goal, so a new campaign
cannot omit it: AI disclosure on request, refusal of OTPs/PINs/card and bank
details, no medical, legal, financial, or emergency advice, opt-outs honoured
without argument, and no promises of price or outcome. The `note` column is
operator-supplied and interpolated into the prompt, so it is treated as
untrusted — newlines collapsed, redirection phrases stripped, length-capped, and
the boundaries stated above it.

**Also fixed, unprompted:** `region` and `locale` were hardcoded on the campaign
dataclass, which contradicts `docs/design-principles.md` Principle 3. They are
now `--region` / `--locale` and omitted from the payload entirely when unset.
The results file is written `0600`.

**Verification.** `test_runner.py` — 148 checks. `test_live_path.py` — new,
drives the whole live loop against an injected fake client. That end-to-end test
caught a duplicate-call bug the isolated tests missed: the first call had already
resolved by the time the second row was read, freeing the reservation. It ships
for that reason. `validate_repository.py` passes; neither suite places a call or
needs credentials. Rebased onto current `main`.

Ray-56:
Current blockers at ac5f583:

- A polling timeout or non-terminal provider status is still recorded as resolved, which frees the recipient for a later redial while the original call remains ambiguous.
- Malformed reservation-ledger lines are silently skipped, so corruption can erase the only active claim and permit a duplicate call.
- Suppressions are loaded only once, and the reservation is marked resolved before the do-not-call record is durably written; concurrent runners or a crash can lose or bypass an opt-out.
- Live mode can bypass the exact recipient allowlist, and the prompt discloses that the caller is AI only if asked rather than at the start.

mohdcodes:
@Ray-56 
Follow-up on `ac5f583` — I audited the whole failure surface rather than
waiting for the next review, using `apps/python/callback-window-coordinator`
as the reference for what "merged" looks like. Seven more real bugs, now fixed:

**Redaction did not recurse.** A phone number inside a list or nested object in
`structured_result` was written to disk unredacted; only top-level strings were
covered. `redact()` now walks dicts, lists, and tuples at any depth.

**Confidence validation was one-sided.** `if confidence < 0.6` lets `NaN`
through, because every comparison against `NaN` is False — so a `NaN` score
auto-closed the call. Infinity and out-of-range values did too. The valid range
`0.0-1.0` is now asserted rather than the invalid one tested.

**An unusable call ID fell through.** `str(created["id"])` raised `KeyError`
into the generic handler. A create that returns no ID now raises explicitly:
a call that cannot be identified cannot be reconciled.

**Results files were overwritten.** `open("w")` meant a second run with the
same `--out` destroyed the first run's record, including any opt-out it
captured. Now writes a timestamped sibling, matching the `open("x")` pattern in
callback-window-coordinator.

**CSV cells were unbounded and uncleaned.** NUL bytes, ANSI escapes,
zero-width spaces, and bidi overrides flowed into the terminal, the logs, and
the agent's prompt; names were unbounded. All cells are cleaned and capped at
the CSV boundary. When cleaning alters a phone number the run says so on
stderr — an invisible character removed from a number changes who is dialed,
and that must not be silent. Verified an altered number still cannot pass an
allowlist it is not on.

`test_runner.py` is now 192 checks (was 161) covering hostile provider
payloads, recursive redaction, and hostile CSV cells. `test_live_path.py` and
`validate_repository.py` pass.

Ray-56:
Current live-safety blockers remain:

- Suppressions are still cached once, so a concurrent runner that started before an opt-out can acquire the recipient after the first runner records the opt-out and resolves its reservation, then call that opted-out person.
- A resolved recipient is released without preserving recipient+batch completion: editing the task/schema within the same batch creates a new provider key and permits another call, despite the documented requirement to change `--batch-id` for a deliberate re-call.
- Ledger validation accepts arbitrary state strings and treats every `resolved*` prefix as terminal, so a syntactically valid corrupt state can free an active claim.

Atomically re-check durable suppression when claiming, retain a recipient+batch completion guard, and validate exact ledger states/transitions before live use.

mohdcodes:
> Current live-safety blockers remain:
> 
> * Suppressions are still cached once, so a concurrent runner that started before an opt-out can acquire the recipient after the first runner records the opt-out and resolves its reservation, then call that opted-out person.
> * A resolved recipient is released without preserving recipient+batch completion: editing the task/schema within the same batch creates a new provider key and permits another call, despite the documented requirement to change `--batch-id` for a deliberate re-call.
> * Ledger validation accepts arbitrary state strings and treats every `resolved*` prefix as terminal, so a syntactically valid corrupt state can free an active claim.
> 
> Atomically re-check durable suppression when claiming, retain a recipient+batch completion guard, and validate exact ledger states/transitions before live use.

I have made the respective changes please review it.

Ray-56:
Current blockers at `713425a`:

- Opt-out state remains fail-open: suppression appends are not atomic, malformed or truncated records are silently skipped, and dispatch reads the file under a different lock. A crash or concurrent append can hide a real opt-out and allow another call.
- Result files are created and populated before they are changed to mode `0600`; a normal umask or crash during the write can leave identifiable call data readable at `0644`. Create the destination privately before writing.

mohdcodes:
> Current blockers at `713425a`:
> 
> * Opt-out state remains fail-open: suppression appends are not atomic, malformed or truncated records are silently skipped, and dispatch reads the file under a different lock. A crash or concurrent append can hide a real opt-out and allow another call.
> * Result files are created and populated before they are changed to mode `0600`; a normal umask or crash during the write can leave identifiable call data readable at `0644`. Create the destination privately before writing.

Both confirmed and fixed. I reproduced each one before changing anything.

Opt-out fail-open. You identified three defects and all three were real. reserve_recipient held the dispatch lock while reading the suppression file, but record_suppression locks the suppression file — different keys, so they never excluded each other. A test confirmed the worst case: a reservation was granted while an opt-out was being appended, meaning the runner would have dialled someone who had opted out. Claiming a recipient now holds both locks, always suppression-then-dispatch so the order cannot deadlock; verified with 20 interleaved claim/opt-out threads. The append was two write calls (header, then record), so a crash could leave a half-written digest — now a single os.write of both, fsynced, with the containing directory fsynced so a new file's entry is durable. And malformed lines were skipped with a warning, which is fail-open for exactly the reason you gave: the unreadable line may be the only record of a real opt-out. A malformed line, or a file not ending in a newline (an interrupted append), now raises SuppressionCorruptError and stops the run.

Result file permissions. Confirmed: with a 022 umask the file existed at 0644 while identifiable call data was written, and a crash during the write left it there permanently. Created via os.open(..., O_CREAT|O_EXCL, 0o600) now — owner-only from the first byte — on the timestamped-fallback path as well. The chmod is removed rather than moved, since it could only ever widen a window that no longer exists.

One thing worth flagging: a shipped test asserted "junk lines are not loaded as opt-outs" — it encoded the same wrong assumption as the code, so it could never have caught this. Replaced with fail-closed assertions that fail on the old implementation.

Verification is 263 unit checks and 42 live-path checks, plus standalone adversarial suites for the race, torn writes, and file modes. scripts/validate_repository.py passes.

Ray-56:
Current blockers at `b40ba63`:

- The documented `python test_live_path.py` suite still has 4 failures: an output path that is a directory is treated as a name collision and writes a sibling file with success, while the shipped test requires the post-call record to be dumped and a nonzero return. The required validation contract is not green.
- `reserve_recipient()` runs outside the per-contact `try`. Its new claim-time suppression/lock errors can unwind the campaign after earlier calls were placed but before the accumulated results are written, losing the local records the app promises never to discard. Preserve results in an outer failure/finally path while still failing closed before the next call.

PR #56, opened 2026-08-01
feat(apps): add DrillSignal business-continuity phone drill app

kenhuangus:
## Summary  Adds DrillSignal, a TypeScript demo app for consented business-continuity outage drills by phone. The app supports simulation (default, no network), fake-server (loopback CALL-E SDK contract), and opt-in live CALL-E mode with explicit safety preview, evidence-gated backup escalation, durable provider-call checkpoints, conservative reconciliation stops, mission control, and masked after-action reports.  ## Type  - [x] New runnable app - [ ] New skill - [ ] New workflow plugin - [ ] New provider adapter - [ ] New scheduler recipe - [x] README awesome-list entry - [ ] Safety or documentation update - [ ] Validation or tooling update  ## Checklist  - [x] Repository-facing content is written in English. - [x] Branch name, commit messages, and PR title follow `docs/git-naming-conventions.md`. - [x] No secrets, tokens, private phone numbers, call recordings, or private transcripts are included. - [x] Real-world side effects are clearly described. - [x] Phone numbers are masked in documentation and test fixtures unless they are clearly fictional. - [x] Recurring workflows include cancellation behavior (N/A - single-shot drills only; cancel documented). - [x] Runnable code has a dry-run, fake-server, or no-call path by default. - [x] `python3 scripts/validate_repository.py` passes.  ## Side effects  | Mode | Side effect | | --- | --- | | simulation (default) | None | | fake-server | Local HTTP to loopback fake only | | live (opt-in) | One or two outbound CALL-E calls per drill |  Default tests and `npm run demo` use simulation only. **No live call was placed during repository verification.**  ## Test plan  - [x] `cd apps/typescript/drill-signal && npm ci` - [x] `npm run verify` (check, test, build, test:dist, smoke:production, demo) - [x] `npm test` - 89 tests, 0 failures (plus 2 post-build dist static-serving tests in `npm run verify`) - [x] `npm run build` - [x] `npm run demo` - primary-unavailable-backup-success simulation completes - [x] `python3 scripts/validate_repository.py` from repository root - [x] Docker multi-stage build; `GET /api/health` and static UI (`GET /`, favicon) respond; operator token required for mutating routes on `0.0.0.0` - [x] Authorized live CALL-E runtime path verified separately; no credential, destination, call ID, recording, or transcript committed  ## Credentials  - `CALLE_API_KEY` - live mode only; server environment; never committed - `DRILL_SIGNAL_OPERATOR_TOKEN` - required when `DRILL_SIGNAL_BIND_HOST` is not loopback  ## Cancellation  Cancel is available in Mission Control. Pre-call cancel is immediate. After provider acceptance, the call ID is durably checkpointed. Cancel or ambiguous outcomes retain reconciliation state and block backup escalation; provider-side stop still cannot be guaranteed (documented in app README).  ## Docker  Multi-stage image: build with full `npm ci`, runtime with `npm ci --omit=dev` only. Non-root user, `/data` volume, port 3847, health check on `/api/health`.


Ray-56:
Two live-call blockers remain. The result-selection conditional discards valid top-level structured output when no recipient-level result exists, and it does not gate on task completion or confidence. More importantly, timeout or unknown is explicitly ambiguous but still advances to the backup, so the primary may still be active while a second party is called. Ambiguous attempts must stop and require reconciliation.

Ray-56:
`npm run verify` currently stops at TypeScript with TS6133 because `mock` is imported but unused, so the test, build, and demo pipeline does not run. Beyond that, accepted call IDs are not durably checkpointed before polling, and failed status is classified before transcript and result evidence, so a crash or ambiguous or late primary can be lost or incorrectly trigger the backup. Persist the accepted ID and reconciliation state and evidence-gate backup escalation.

Ray-56:
The accepted-ID checkpoint and validation pipeline are now fixed, but current recovery/read boundaries still block merge. A process death while createCall may have been accepted but before an ID returns is materialized as failed, so the unknown provider outcome is not retained for reconciliation; cancellation can also report that no provider call started while an accepted ID exists because callsPlaced is still zero. In addition, unauthenticated GET /api/drills/:id/preview returns the active drill object with full E.164 numbers, bypassing publicDrillView. Preserve create ambiguity, check the active provider ID before cancellation messaging, and protect or redact preview reads.

Ray-56:

Remaining Must Fix only:

- GET /api/drills/:id/preview exposes the active drill and full real numbers without authorization/redaction.
- Cancellation can claim no provider call started while an accepted provider ID exists; correct that statement.

PR #60, opened 2026-08-01
feat(apps): add CallNeuron scholarship outreach

ted0103:
## Summary

- add CallNeuron, a functional TypeScript operator prototype for consent-first scholarship and education-support outreach
- connect to CALL-E at runtime through brokered MCP authentication with separate plan, explicit confirmation, run, and status-monitoring steps
- support manual, CSV, XLSX, DOCX, and selectable-text PDF intake with per-recipient consent review, neutral voicemail, human dispositions, and privacy-minimal export

## Real-world use case

Scholarship teams can lose strong candidates between an approved shortlist and the first useful conversation. CallNeuron handles the repetitive first contact without ranking applicants or making award decisions. It introduces itself, explains why it is calling, asks permission, stays inside an approved brief, and hands interest or questions back to a human.

## Safety and side effects

- creating a CALL-E plan cannot ring a phone
- each real call requires a fresh operator checkbox and explicit confirmation
- one recipient is handled at a time; there is no batch dispatch, hidden scheduling, or automatic retry
- voicemail is off by default and excludes student and offer details
- tests use the shared fake MCP broker and reserved fictional numbers
- recipient data stays browser-local; exports exclude names, phone numbers, transcripts, and offer details

## Verification

- `npm run verify` — TypeScript, 12 focused tests, and Vite production build
- rendered QA at 1280px and 390px with no horizontal overflow or console errors

## Live operator prototype

https://call-neuron-preview.pages.dev/

No real call was placed during automated or rendered verification.


Ray-56:
Accepted run IDs and attempt state exist only in React memory while recipient dispositions persist. A refresh or second tab therefore makes the same recipient callable again, including recipients already marked opted out. A lost `run_call` response also leaves no durable unresolved record or same-run reconciliation path. Persist a content-bound dispatch reservation and accepted run ID before allowing another plan, and fail closed on ambiguous acceptance.

ted0103:
> Accepted run IDs and attempt state exist only in React memory while recipient dispositions persist. A refresh or second tab therefore makes the same recipient callable again, including recipients already marked opted out. A lost `run_call` response also leaves no durable unresolved record or same-run reconciliation path. Persist a content-bound dispatch reservation and accepted run ID before allowing another plan, and fail closed on ambiguous acceptance.

Hi Ray-56, 
Big Thanks on the comment, it really means a lot. Will move on to resolve it immediately 🫡

ted0103:
Implemented in 3342dd4 — thank you for catching this.

- A SHA-256 content-bound dispatch reservation is now written to IndexedDB before `plan_call`, with an atomic per-recipient key that also blocks concurrent tabs and re-imported rows.
- The durable phase moves to `dispatching` before `run_call`; the accepted run ID/status is persisted immediately on response and restored after refresh for same-run `get_call_run` reconciliation.
- A missing/invalid `run_call` response becomes a durable `acceptance_unknown` lock. It has no retry action and cannot be discarded; only reservations that never entered `run_call` can be removed.
- Opted-out recipients are excluded in the UI and rechecked from the durable draft inside the dispatch transaction. Reset preserves dispatch locks so it cannot reopen a recipient.
- Cross-tab updates use `BroadcastChannel` plus a visibility refresh. Confirmation secrets and transcripts remain memory-only.

Verification: `npm run verify` passes (TypeScript, 17/17 tests, production build), repository validation passes, and a headed browser smoke test confirmed opt-out and accepted-run blocking after reload with no browser errors. All broker tests use fake reserved data; no real call was placed.

Ray-56:
Durable protection is still keyed by phone plus mutable student and employee codes. Re-importing the same phone with changed codes, or under another student, creates a new key and allows another call; an opt-out without a dispatch attempt lives only in the draft and is deleted by Reset. Key call locks and suppression on a stable phone identity, retain opt-outs outside the resettable draft, and test reset and re-import bypasses.

ted0103:
Hi Ray56,

Thank you for your help with the project. Your suggestions and support were
really useful, and I genuinely appreciate your time

Yours sincerely
Ted

PR #62, opened 2026-08-01
feat: add Shopify cash-on-delivery confirmation gate plugin

luongs3:
## What this adds

`plugins/shopify-flow-cod-confirm/` — a Shopify Flow plugin that decides whether a cash-on-delivery order should be shipped, using a CALL-E confirmation call as the evidence.

Cash on delivery is the majority payment method across much of South-East Asia, and a well-known share of COD orders are fake, duplicated, or carry a wrong address. Merchants already pay humans to phone every order before dispatch. This plugin does that call and converts it into a **decision written back onto the order**, so a warehouse can act on a tag instead of reading a transcript.

The contribution is the shipping decision gate, not the phone call.

## Try it without placing a call

No credentials, no Shopify store, no phone:

```bash
cd plugins/shopify-flow-cod-confirm
node src/cli.mjs preview --order examples/order-cod.json --merchant "Demo Store"
npm test
```

`preview` prints the exact call script, the masked number and the idempotency key, and contacts nobody. `npm test` runs 31 tests against a local fake CALL-E server and a local fake Shopify Admin API, covering every branch of the decision table.

## Decision table

| Call outcome | Decision |
| --- | --- |
| Confirmed, address correct | `ship` |
| Confirmed, address wrong, correction captured | `ship` |
| Confirmed, address wrong, no correction | `hold` |
| Not confirmed / cancellation requested | `hold` |
| Answered, `structured_result` is null | `hold` |
| No answer, first attempt | one retry |
| No answer, final attempt / never terminal | `hold` |
| Ambiguous creation | `hold`, key preserved |
| Phone not E.164 | `hold`, no call placed |

## Notes on CALL-E behaviour that shaped the implementation

- **It polls `GET /v1/calls/{id}` and does not rely on `webhook_url`.** The field is accepted at call creation, but no webhook arrived for a completed call in field testing, so polling is the only reliable completion signal. Happy to update this if the behaviour is expected to change.
- **An answered call with a null `structured_result` is treated as a first-class state** and held, never read as a confirmation.
- Unanswered calls and failed routes are not billed, which is what makes the single retry cheap.

## Safety

- Identity check precedes any order disclosure, so a wrong number learns nothing about the order.
- A fixed, non-overridable instruction prohibits medical, legal, financial and emergency advice and forbids taking payment details.
- Stable idempotency key per (shop domain, order id, attempt) — a Shopify webhook retry cannot place a second call to a real customer.
- Ambiguity is never resolved by dialling again: a network failure mid-create, a `5xx`, a `429`, or a `2xx` with no call id all hold the order and preserve the key for reconciliation.
- Shopify webhook HMAC verified before a call can be created; the server warns loudly when the secret is unset.
- Phone numbers masked in logs, previews, run status responses and the order note.
- No recurring schedules. One order, at most two attempts.
- Sample orders use fictional reserved numbers in the `+1500555xxxx` range.

Cancellation and rollback are documented in the plugin README: disable the Flow action or webhook, or rotate `CALL_E_API_KEY`. A call already accepted by CALL-E cannot be cancelled through this plugin.

## Checklist

- [x] No secrets, tokens, private phone numbers or personal data
- [x] States the host and provider it supports
- [x] Describes side effects
- [x] Install and usage instructions
- [x] Masked or fictional phone numbers in samples
- [x] Cancellation / rollback behaviour documented
- [x] Dry-run, fake-server and no-call paths included
- [x] `python3 scripts/validate_repository.py` → `Repository validation passed.`
- [x] Branch name checked with `scripts/check_branch_name.py`
- [x] `README.md` resource list and `plugins/README.md` index updated in the same change

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,7 @@ Plugins should be explicit about inputs, outbound call side effects, credential 
 | [`plugins/n8n-nodes-calle`](plugins/n8n-nodes-calle/) | n8n | Documentation-only pointer to the standalone `@call-e/n8n-nodes-calle` community node package for native outbound-call nodes. |
 | [`plugins/dify-template`](plugins/dify-template/) | Dify | Importable Dify workflow DSL template for a one-shot outbound call tool with dry-run preview, API health gating, and masked results. |
 | [`plugins/hubspot-calle`](plugins/hubspot-calle/) | HubSpot | Static HubSpot Projects app for creating CALL-E call tasks from CRM records and workflow App Cards. |
+| [`plugins/shopify-flow-cod-confirm`](plugins/shopify-flow-cod-confirm/) | Shopify | Cash-on-delivery confirmation gate that turns a CALL-E call into a ship-or-hold decision written back onto the order as tags and a timeline note, with a no-call preview path. |
 
 ### Safety patterns
 
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -32,3 +32,4 @@ Each plugin should document:
 | [`n8n-calle-api`](n8n-calle-api/) | n8n | Importable CALL-E API workflow template for one-by-one outbound calls, metadata round trips, call status signals, transcripts, summaries, and structured results. |
 | [`@call-e/n8n-nodes-calle`](n8n-nodes-calle/) | n8n | Community node package for creating, waiting on, retrieving, and listing events for CALL-E AI-agent phone-call tasks. |
 | [`hubspot-calle`](hubspot-calle/) | HubSpot | Private static app with a direct-call workflow action and two explicitly confirmed CRM record App Cards; it does not write CALL-E state back to HubSpot. |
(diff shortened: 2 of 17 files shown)


Ray-56:
The COD gate is not safe to run live yet:
- Missing `SHOPIFY_WEBHOOK_SECRET` merely warns and then accepts arbitrary webhooks that can trigger calls.
- Every manual payment method is classified as COD, including bank transfer or check orders.
- Failed calls are retried without reachability or refusal evidence, while ship decisions ignore task completion, confidence, identity, and local schema validation.
- Address-correction outcomes mark orders ready to ship without updating the actual Shopify shipping address.

Ray-56:

Remaining Must Fix only:

- Missing SHOPIFY_WEBHOOK_SECRET accepts arbitrary live-triggering webhooks.
- All manual payment methods are treated as COD, causing unintended outbound calling for bank-transfer/check orders.
- Failed/ambiguous outcomes are retried; shipping mutations use insufficient result evidence.
- Address correction marks order ready without applying the address correction to Shopify.

PR #63, opened 2026-08-01
feat: add confidence-gated-results skill for null structured results

luongs3:
## The state nobody handles

A CALL-E call can reach `completed`, bill normally, and still hand back `structured_result: null`. The person answered. They talked for ninety seconds. The extraction produced nothing usable.

Most integrations read the result as a value and move on — so `null` quietly becomes "no" or an empty success, and a workflow ships an order, closes a ticket or marks a lead dead on the strength of a call that produced no evidence at all.

`skills/confidence-gated-results/` makes that state first class and gives it three exits.

## Three-tier outcome contract

| Tier | Meaning | Action |
| --- | --- | --- |
| **COMMIT** | Every blocking field present and actionable | Use it, record provenance |
| **REPAIR** | Call connected, extraction incomplete, gap is narrow | **One** narrowed follow-up call |
| **ESCALATE** | Ambiguous, contradictory, sensitive, exhausted | Hand to a human with the trail |

Default when unclassifiable is ESCALATE. `null` is not "no", and silence is not consent.

## Why the repair call decomposes the schema

A `null` usually means one composite schema asked for four things at once and the extractor could not satisfy all of them, so it emitted nothing rather than something partial.

So a repair call must **not** re-ask the original question — that reproduces the same failure and bills twice. It splits the failed schema into the smallest sub-schema covering only the *blocking* missing fields.

Rules enforced in code, not prose: one repair per original call, the schema is always a strict subset, a confidently-captured field is never re-asked, and the idempotency key is derived from the original call id.

## The decisive rule

The same `null` gets **opposite** treatment depending on `GET /v1/calls/{id}/events`:

- 96 seconds, 14 turns → extraction failure → **REPAIR**
- 5 seconds, 1 turn → soft refusal → **ESCALATE**, never call back

The terminal result alone cannot tell these apart. This is why the skill reads the event stream before choosing a tier, and it is the reason the classifier takes `turnCount`.

## Try it

```bash
cd skills/confidence-gated-results
node --test scripts/classify-result.test.mjs     # 19 tests, no credentials, no calls
node scripts/classify-result.mjs --call call.json --required a,b --turns 14
```

`scripts/classify-result.mjs` is pure functions plus a thin CLI. Exit codes `0` commit, `5` repair, `6` escalate. Tests cover the tier table, subset-schema construction, idempotency key derivation, sensitive-field refusal, contradiction detection, `false`-is-not-missing, and provenance merging.

## Safety

- A repair call is a **second call to someone already interrupted once** — justified only when the first call produced a genuine conversation *and* the missing field genuinely blocks the workflow.
- Never repair after a refusal, a hangup, a "call me back later", or for a sensitive field (payment, identifiers, health, OTP) — those escalate unconditionally.
- One repair per original call, enforced by idempotency key. No recurring schedules.
- Fictional reserved numbers in all samples; masking required in escalation payloads.

## Checklist

- [x] No secrets, tokens, private phone numbers or personal data
- [x] `references/safety.md` and `references/examples.md` present
- [x] Describes side effects and cancellation/rollback
- [x] No-call path: the classifier needs no credentials and places no calls
- [x] `python3 scripts/validate_repository.py` → `Repository validation passed.`
- [x] Branch name follows `docs/git-naming-conventions.md`
- [x] `README.md` resource list updated in the same change

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,7 @@ This project is an awesome list for AI-agent phone-call workflows. Add resources
 ### Skills
 
 - [`call-reminder`](skills/call-reminder/) - Scheduler wrapper skill for recurring CALL-E phone-call reminders.
+- [`confidence-gated-results`](skills/confidence-gated-results/) - Three-tier outcome contract for a phone call that succeeded but returned no usable structured result: commit, one narrowed repair call, or escalate with the evidence trail.
 - [`deployment-approval-call`](skills/deployment-approval-call/) - Spoken, code-verified human approval before an agent or pipeline does something irreversible.
 - [`google-form-callback`](skills/google-form-callback/) - Google Form response workflow for safe one-off callback calls with dry-runs, scheduling plans, and Sheets writeback. See the [workflow guide](docs/google-form-callback/).
 - [`outbound-call-skill-creator`](skills/outbound-call-skill-creator/) - Creator skill for generating focused outbound phone-call workflow skills from Google Forms, TikTok Ads, Notion, Airtable, local CSV files, or custom sources.
--- a/skills/confidence-gated-results/SKILL.md
+++ b/skills/confidence-gated-results/SKILL.md
@@ -1,0 +1,136 @@
+---
+name: confidence-gated-results
+description: Handle a CALL-E phone call that succeeded but returned no usable structured result. Use when a call reaches a terminal status and structured_result is null, partial, or fails schema validation, and an agent must decide between committing the result, placing one narrowed repair call, or escalating to a human with the evidence trail.
+license: MIT
+---
+
+# Confidence-Gated Results
+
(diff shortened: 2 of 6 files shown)


Ray-56:
The classifier does not yet provide the promised confidence gate. COMMIT checks only field presence, ignores task completion, confidence, schema types and enum values, and the default empty required-field list can commit a null result. Repair can recreate the full schema and generates a second-call task without inheriting identity, consent, disclosure, voicemail, or safety boundaries. Add an explicit per-call authorization and cancellation path before any repair call.

Ray-56:

Remaining Must Fix only:

- Classifier's advertised COMMIT gate accepts null/invalid outputs with default empty requirements and only presence checks.
- Repair call path needs explicit new-call intent and inherited recipient authorization/cancellation boundaries.

PR #66, opened 2026-08-02
feat(apps): add CallProof verification workflow

emezac:
## Summary

Adds CallProof, a runnable verification app that combines a Rails orchestration layer built with AgentKit Rails and an asynchronous Call Analyzer service.

The workflow preserves the verification contract, transcript-derived evidence, analyzer results, and human approval state. Rails submits analysis jobs, receives signed and idempotent callbacks, and uses an AgentKit human-in-the-loop checkpoint before the result can be approved.

The package includes:

- `apps/web/callproof/` with the Rails app, FastAPI analyzer, Redis/RQ worker, and Docker Compose stack
- a safe fake/no-call path enabled by default
- CALL-E provider integration guarded by explicit live-call controls
- signed HMAC callbacks and replay protection
- tests and documentation for the end-to-end verification workflow
- an awesome-list entry in the root README

The standalone project is also available at https://github.com/emezac/callproof.

## Type

- [ ] New skill
- [x] New runnable app
- [x] New workflow plugin
- [x] New provider adapter
- [ ] New scheduler recipe

## Verification

- [x] `python3 scripts/validate_repository.py`
- [x] Rails test suite: 19 tests, 118 assertions, 0 failures
- [x] Python test suite: 4 passed
- [x] RuboCop: 71 files inspected, no offenses
- [x] Zeitwerk autoloading check
- [x] Brakeman: 0 warnings
- [x] Docker end-to-end fake flow: Rails -> Analyzer/RQ -> signed webhook -> AgentKit HITL -> approval/resume -> approved
- [x] Real outbound CALL-E call

The real CALL-E call remains intentionally unverified, so this PR is opened as a draft. No real phone call was placed during validation.

## Safety and real-world side effects

The default flow is fake/no-call and uses the clearly fictional number `+1 202-555-0147`. A real outbound call requires all of the following: `provider=calle`, the live-call feature switch, per-request confirmation, the typed confirmation phrase, a CALL-E API key, and an HTTPS callback URL.

## Checklist

- [x] Recurring workflows include cancellation behavior or are not applicable.


Ray-56:
This draft has critical authorization and result-integrity blockers:
- The Rails live workflow has no authentication or authorization; any visitor can create a request, read its predictable confirmation phrase, and spend the server CALL-E account.
- An arbitrary, including HTTP, base URL receives the Bearer credential, and ambiguous network or 5xx create failures are not reconciled before a replacement call.
- Recipient status, phone, task completion, and confidence are not validated before marking a call complete.
- The analyzer always returns success at confidence 0.98 from weak or unrelated evidence, so unsupported calls can be auto-verified.

emezac:
@Ray-56 .  Thanks for the review — all four blockers are addressed in 986fe67:

1. **Auth/authz**: the live-call workflow (create/preview/confirm/cancel) and live
   request details are now behind fail-closed operator HTTP Basic auth. The
   confirmation phrase is derived from a random per-request token, not the
   sequential id, and is compared with a constant-time check.
2. **Credential exfiltration / ambiguous creates**: the CALL-E Bearer key is only
   sent over HTTPS to an allow-listed host (`api.heycall-e.com` by default).
   Timeouts and 5xx are treated as *ambiguous* — no replacement call is placed;
   the stable Idempotency-Key is reused on a deliberate retry.
3. **Completion integrity**: recipient status, recipient phone match, task
   completion and a minimum result confidence are validated before a call is
   marked complete. Any mismatch fails closed (call → failed, no analysis).
4. **Auto-verification**: the analyzer no longer returns a constant 0.98. Confidence
   is derived from evidence strength; unsupported/unrelated evidence yields low
   confidence, `needs_human_review=true`, and `goal_completion="unknown"`.

Tests, RuboCop, Zeitwerk, Brakeman and the Python suite pass; new regression tests
cover each fix. Still no real call is placed by the demo or tests.

Ray-56:
Current blockers:\n\n- Operator-created live previews are stored with live_mode=false, so the CallRequests authorization filter does not run. Their sequential URLs expose preview objectives and recipient metadata without login, and the page incorrectly labels them “Fake / no call.”\n- Ambiguous live-create failures are collapsed into failed and reported as having failed safely, even though the provider may have accepted the call. This needs an unresolved state and reconciliation path.\n- Result persistence still accepts missing recipient status, phone binding, and confidence, and it can infer task completion from terminal status/counts rather than explicit completion evidence.\n- Production Basic Auth is not protected by enforced HTTPS in the current Rails configuration.\n\nPlease keep the PR in draft until these live-path authorization, ambiguity, and result-integrity issues are closed.

emezac:
@Ray-56 .

Thanks again — all four are addressed in f8c54c3:

1. **Authz**: previews are operator-initiated but stored live_mode=false, so gating on
   live_mode leaked them. Added an `operator_initiated` flag; CallRequests auth now gates
   on it, operator requests are kept off the public home page, and the hardcoded
   "Fake / no call" label is replaced by an honest per-request mode.
2. **Ambiguity**: ambiguous create no longer reports "failed safely" — the request goes
   to a new `unresolved` state with a `reconcile` action that re-attempts using the stable
   Idempotency-Key (CALL-E dedups, so no second call).
3. **Result integrity**: persistence now requires recipient status, phone binding, and
   confidence, and an explicit `task_completed=true` — no inference from status/counts.
4. **Transport**: `config.force_ssl` enforced in production (excluding /up) so operator
   Basic Auth never traverses plaintext HTTP.

Regression tests cover each; Rails 35 green, RuboCop/Zeitwerk/Brakeman pass. PR kept in draft.

Ray-56:
Current blockers:\n\n- The new operator_initiated migration defaults every existing row to false and does not backfill pre-existing live/operator requests. The controller now authenticates only operator_initiated records, so an existing live_mode=true request becomes anonymously readable after migration. Backfill existing live rows and keep live_mode || operator_initiated as defense in depth.\n- The deterministic analyzer can auto-verify a failed objective. With delivery_changed=false, order_number_confirmed=true, surcharge_cents=0, and a transcript explicitly saying the delivery could not be moved, it returns goal_completion=complete, confidence=0.95, needs_human_review=false, and claims the goal was achieved. Completion must follow the declared success conditions and transcript evidence.\n- Reconciliation treats every 4xx as definitive no-call. A 409 idempotency conflict can mean the original call exists, but the controller marks the request failed and tells the operator no call was placed. Keep 409 unresolved and reconcile it as an ambiguous outcome.\n\nThe new preview rows, HTTPS enforcement, and result binding are improved, but these migration, analyzer, and reconciliation paths remain unsafe.

emezac:
@Ray-56 . Thanks. All three are addressed in 198bd8d:

1. **Migration downgrade** — the migration now backfills `operator_initiated = TRUE`
   for rows with `live_mode`, a `confirmed_at`, or status `awaiting_confirmation`.
   Defense in depth: authorization and the public home-page filter both key off
   `CallRequest#requires_operator_auth?` (`operator_initiated || live_mode ||
   confirmed_at`), so a row predating the column stays protected. Verified by migrating
   a database that held a pre-existing live row.
2. **Analyzer** — completion is now judged only against the contract's declared
   `success_conditions`, cross-checked with the transcript. `delivery_changed=false`
   with `order_number_confirmed=true` yields `goal_completion=failed`, low confidence,
   `needs_human_review=true`, and an `objective_not_met` finding citing the denying turn;
   the summary no longer claims the goal was achieved. A structured success denied by the
   transcript is recorded as a contradiction and never auto-verifies.
3. **Reconciliation** — 409 (idempotency conflict, so the original call likely exists)
   and 408 are ambiguous like 5xx: the request stays `unresolved` rather than being
   reported as "no call was placed". Other 4xx remain definitive.

Regression tests cover each. Rails 39 green, Python 10 passed, RuboCop/Zeitwerk/Brakeman
clean. PR kept in draft.

Ray-56:
Current analyzer blockers on `198bd8d`:

- Successful conditions are still not bound to affirmative transcript evidence. I reproduced `delivery_changed=true` with a transcript discussing only an unrelated $1.00 support fee; the verdict was `complete`, confidence `0.95`, risk `0.08`, and `needs_human_review=false`.
- `policy_adherence` only checks the surcharge. Declared `required_disclosures` and `forbidden_commitments` are never evaluated, so the same unsupported call reports `policy_adherence=true` and `missing_disclosures=[]`.

Please require positive transcript support for each successful condition and route unsupported or unevaluated policy claims to human review.

emezac:
Hi @Ray-56 .

Both are addressed in db5b885.

1. **Success needs transcript support.** A condition now counts only when it is declared,
   reported met by the provider, AND corroborated by a transcript turn. Your repro
   (`delivery_changed=true` over a call discussing only an unrelated $1.00 fee — the
   amount matched by coincidence) now returns `goal_completion=partial`, confidence
   `0.65`, risk `0.8`, `needs_human_review=true`, with the uncorroborated condition named
   in `negotiated_terms.unsupported_success_conditions` plus an explicit contradiction.
2. **Every declared policy term is evaluated.** `required_disclosures` with no transcript
   support land in `missing_disclosures`; a `forbidden_commitments` item discussed on the
   call sets `unauthorized_commitment`. Either drops `policy_adherence` and forces review.

Corroboration is topical (domain vocabulary, ES+EN), not semantic — documented as such,
and it fails toward human review; the LLM provider remains the path to semantic judgement.

Tests cover the exact repro, missing disclosure, forbidden commitment, and a fully
supported happy path that must still auto-verify. Python 14 passed, Rails 39 green.

Ray-56:
Thanks for the update. The success gate still accepts the agent's own question as corroboration. For example, with agent text “Can we move the delivery to Friday?”, recipient text only “There is a $1.00 support fee this month,” and `delivery_changed: true`, the evaluator returns `complete`, confidence `0.95`, risk `0.08`, and no human review. `_supporting_turns()` scans every speaker and treats the word “delivery” in the question as positive evidence. Please require affirmative recipient-side support (or route topical-only evidence to review).

emezac:
@Ray-56 . You're right, and thank you for staying on it — this was the fourth report of the same shape, so I stopped patching the reported case and changed the default the evaluator starts from. `a8b63fc`.

**What was actually wrong.** The evaluator assumed success and looked for reasons to doubt it. `_supporting_turns()` scanning every speaker was one symptom; the design was the cause. Anything nobody had enumerated produced no doubt, and therefore resolved as verified — an unlisted refusal ("Friday is a bad day for that"), a voicemail, a `system` turn, the agent's own question. That is why each fix held for exactly one case: the failure mode was the absence of a rule, and you cannot enumerate absences.

Completion now requires positive evidence, which inverts what an unanticipated case does — it lands in human review by construction.

**The three rules that follow.**

1. **Only the counterparty corroborates.** The agent is the party under audit, so its question ("Can we move the delivery?") or its claim ("I've moved it") can never prove the outcome. Agent turns are read only to establish what a following "yes" is answering. `system` and `unknown` prove nothing.
2. **Corroboration needs an explicit affirmation**, either on topic or answering the agent's on-topic turn. A topic raised without agreement is reported as `weakly_supported_success_conditions` and routed to review — your fallback ask. Completion no longer depends on `NEGATION_PATTERN` at all; that pattern now only reports a result/transcript disagreement, so an unlisted refusal phrasing can no longer read as consent.
3. **Agent-side obligations are read from agent turns.** A recipient asking "are you recording this?" is not the agent making a disclosure, and a recipient floating a forbidden option is not the agent offering it.

Evidence is also scoped per condition: word groups shared by every declared condition carry no discriminating power and are dropped. On a contract of `delivery_date_confirmed` + `delivery_time_confirmed`, "yes, the delivery date is fine" no longer confirms the *time* through the shared word "delivery". That one was found by the new suite, not by a report.

**How I'm trying to keep this from happening a fifth time.** `call_analyzer/tests/test_evaluator_adversarial.py` — 14 cases, each an attempt to force a false "verified", grouped by the hole it closes so a regression names itself rather than needing a bisect. Two of them pin the happy path, because a gate that never passes is not a gate. Full suite: 28 passing; Rails 39 runs / 181 assertions; RuboCop clean.

**What this still is not.** Corroboration is lexical, not semantic — it can tell that nothing in the call is about the delivery, but it cannot judge nuanced wording, and the affirmation list is closed by design. It fails toward human review, and the LLM provider is the path to semantic judgement. If you'd rather the topical-only tier not exist at all and everything short of an explicit affirmation go straight to review, that's a one-line change and I'll make it.

Ray-56:
Current blockers on de491e2:

- The positive-evidence gate still auto-verifies explicit denials. With recipient text “That is not correct; the delivery has not changed” and provider delivery_changed=true, the verdict is complete at 0.80 confidence with no human review. The narrow negation list misses this wording while the affirmation matcher accepts “correct.”
- Required disclosures are still satisfied by any one topical token. A confirm_order_number requirement is treated as present when the agent only says “confirm” about delivery date/time, yielding policy_adherence=true and no review although no order number was disclosed.
- CALL_ANALYZER_ENV defaults to development, so a deployment that omits both environment variables still signs callbacks with the public repository secret. The known default should require an explicit development/test mode.
- Reconciliation marks any later non-ambiguous provider error as proof that no call was placed. A 401 or other definitive rejection of the retry does not prove that the original timed-out request was not accepted; keep the original outcome unresolved unless the provider identifies it by the stable idempotency key.

Ray-56:
Current blocker on 18eb55a:

- The analyzer still treats a listed affirmation word as semantic agreement whenever the denial is not one of the enumerated particles. Using the app's actual delivery_date_confirmed and delivery_time_confirmed policy, an agent disclosure followed by recipient turns “I doubt the delivery date is correct; the fee is $0.00” and “I dispute that the delivery time is correct,” with provider date/time fields present, returns goal_completion=complete, confidence=0.95, risk=0.08, and needs_human_review=false. Both turns explicitly reject the claimed outcome, but “correct” unlocks auto-verification because “doubt” and “dispute” are not in NEGATION_PARTICLE. Please make auto-verification depend on evidence that can establish meaning, or route language outside a provably safe supported form to review; adding these two verbs to another deny-list would leave the same open-ended failure mode.

Ray-56:
Current blocker at `7f977c6`:

- The new allow-list still auto-verifies an explicit contradiction. `_accountable()` accepts a token when any contract word is a substring (`evaluator.py:323-336`), so with provider `delivery_changed=true` and recipient “Yes, the delivery is unchanged,” the verdict is `complete`, confidence `0.80`, risk `0.08`, and `needs_human_review=false`. Match whole supported terms or route morphology/polarity that cannot be interpreted to review.

Ray-56:
Blocking issue on current head `ed1596b`:

- The exact-protocol evaluator only examines the recipient turn immediately after each canonical statement and ignores later retractions (`evaluator.py:241-264`). I reproduced a fully canonical confirmation followed by “Actually, no. The date does not work.”; the verdict still returned `complete/compliant`, confidence `0.95`, risk `0.08`, `needs_human_review=false`, and no contradiction. A revoked agreement must not be presented as verified; detect subsequent contradictions or fail closed to human review.

Ray-56 (approved):
The frozen Must Fix item is resolved: later free-form turns now invalidate prior exact-protocol proofs and route the result to human review. Repository validation and all 41 analyzer tests pass. The stale PR-body wording and EOF whitespace are non-blocking under the proportional policy.

PR #67, opened 2026-08-02
feat(zapier-calle): add Zapier platform integration with fail-closed dispositions

TyrusRC:
## What this adds

Two contribution areas, which work together but stand alone:

- **`plugins/zapier-calle/`** - a Zapier Platform CLI integration for CALL-E, plus four example recipes and one safety pattern.
- **`skills/calle-script-advisor/`** - an Agent Skill with a dependency-free linter that checks a CALL-E task script and `result_schema` *before* a call is placed, so a call is not spent on a script that cannot produce a clean answer. It is the pre-flight half of the same idea the plugin enforces post-flight.

| Component | Purpose |
| --- | --- |
| `Place Call and Wait for Outcome` (create) | Places a call and resumes the Zap step with the terminal result |
| `Start Call (No Wait)` (create) | Fire-and-forget for volume, returns `call_id` |
| `Find Call Result` (search) | Reconciliation path for a stranded or missed call |
| `Call Completed` (trigger) | Static webhook, reacts to calls placed anywhere including CALL-E's CLI and MCP |
| `lib/disposition.js` + `lib/result-quality.js` | Fail-closed outcome classifier, ten dispositions |
| `lib/calling-window.js`, `lib/suppression.js`, `lib/retry-policy.js`, `lib/opt-out.js` | The four refusals: quiet hours, do-not-call, redial frequency, revoked consent |
| `lib/reconcile.js` | Authenticated re-read of any call named by an unsigned webhook, so no webhook body is ever classified |
| `lib/grounding.js` | Verifies an extracted answer against the transcript: a `<field>_quote` the recipient never said cannot be actionable, and a quote the schema demanded but the extraction omitted counts the same |
| `lib/notification-freshness.js` | Dates a webhook delivery by CALL-E's own `completed_at`, so a replayed old result cannot arrive as news |
| `skills/calle-script-advisor/SKILL.md` | The authoring rules: what a task script must state, and which result-schema shapes produce answers a workflow can act on |
| `skills/calle-script-advisor/scripts/check-call-script.mjs` | The linter itself - no dependencies, 25 tests, runs offline |

## Try it

The integration is deployed as a private Zapier app. Anyone can install it with
this invite link, connect a CALL-E API key, and run it:

https://zapier.com/developer/public-invite/244591/5987695714e8692147ff45fdcb3be684/

The fastest way to confirm it works end to end is
`examples/two-step-free-plan.md` - a deliberately minimal two-step Zap that runs
on Zapier's free plan and needs no third-party app connection. `Dry Run` starts
on, so the first run previews and places no call.

## Why the callback pattern, rather than polling

A Zapier action's `perform` must return in roughly 30 seconds. A phone call takes minutes. The existing n8n template in this repository polls `GET /v1/calls/{id}` for up to four minutes, which does not port.

Polling is not available here either: the Developer API (OpenAPI 0.6.0) exposes `POST /v1/calls`, `GET /v1/calls/{call_id}`, `GET /v1/calls/{call_id}/events`, and the goals routes. There is no list endpoint, so a Zapier polling trigger cannot be built at all.

`POST /v1/calls` accepts a per-request `webhook_url`. This integration passes it `z.generateCallbackUrl()`, so the action returns early, Zapier holds the step for up to 30 days, and `performResume` receives CALL-E's terminal `call.completed` event as `bundle.cleanedRequest`. `performResume` is CLI-only - the Zapier Platform UI cannot express this, which is why the integration is a CLI app.

## Safety properties

Each of these is enforced in code and covered by tests.

- **Fail closed.** `deriveDisposition` returns `is_actionable: true` only for a fully agreeing terminal success: `call.completed`, `status: completed`, `task_completed: true`, `completion_confidence.label: high`, its numeric `.score` at or above a configurable floor, and a `structured_result` in which every field the caller declared `required` carries a usable value. The default branch is `needs_human`. An unrecognized status or event type resolves to `needs_human`, never to success, because the API vocabulary can grow after this ships.
- **A present result is not a usable result.** This is the check worth reviewing first. CALL-E's own `result_schema` guidance says to include an `unknown` enum member for calls that cannot produce evidence - so a call where nobody answered returns `status: completed`, `task_completed: true`, a `high` label, and `{"confirmed": "unknown"}`. Every field says success. A required field carrying `unknown`, `null`, `""`, `[]` or `{}`, or absent entirely, is classified `review_required`. `false` and `0` are answers and stay `confirmed`. Observed live: a declined call returned `{"heard_clearly": "unknown"}`.
- **A revocation of consent overrides the business outcome.** When the recipient asks not to be called again, the call is forced to `needs_human`, never actionable, `lead_state: blocked_compliance`, whatever the extracted answer was, with the quote and its offset surfaced so the number can be added to a do-not-call list. Only the recipient's turns are scanned - the agent's own script contains the opt-out disclosure state bot laws encourage. Documented as an English phrase list with a stated ceiling, not as intent classification.
- **Three pre-flight refusals.** `outside_calling_window` (TCPA-grounded, callee's IANA timezone, DST-correct, raw UTC offsets rejected), `suppressed` (digits-only do-not-call matching), and `retry_policy_blocked` (default 2 per day, 4 hours apart). All three are stateless by design - a Zapier action has no durable storage, and inventing one would be worse than being honest there is none. Each fails closed on unreadable input, and none of them lets `Place Call and Wait` mint a callback URL for a call it refused.
- **`outcome_unknown` is distinct from `failed`.** A timeout or a non-terminal status does not mean the call did not happen. Treating ambiguity as failure is how a workflow dials the same person twice.
- **Idempotency keyed to the canonical payload.** `Idempotency-Key` is a SHA-256 digest over `{task, recipients, result_schema, recipient_result_schema, metadata}`, computed before `webhook_url` is attached. It is deliberately not derived from a Zap or run id: those are stable across edits, so an edited retry could otherwise replay an older call with a different number.
- **A webhook body is never classified.** Both webhook surfaces receive an unsigned body on an unauthenticated URL, and CALL-E publishes no signing secret. Matching the call id does not fix that - the id travels inside the same untrusted body, so an id an attacker knows is an id they can echo back, and everything behind it would be whatever they typed. The body is therefore treated as a notification naming a call; the call is then read back from CALL-E over the connection's own API key, and *that* response is what gets classified. `performResume` returns `needs_human` with `verified: false` when the read fails. The `Call Completed` trigger declines to fire at all when CALL-E has no such call on the connection (a forged or misrouted delivery), but still surfaces a call it simply could not reach CALL-E about, as `needs_human`/`verified: false` - a briefly unreachable API must not silently swallow a real outcome. `Find Call Result` reports `verified: true` unconditionally; it reads from CALL-E directly and never had a webhook to distrust.
- **An authentic record is not an authorized notification.** Re-reading the call proves it is real and belongs to the connection. It cannot prove the POST that named it was authorized or new: anyone who learns the static webhook URL and one valid call id can send it again, with a fresh envelope id each time, and each send re-reads the same genuine, already-finished result. Freshness *is* verifiable server-side, so it is checked - the outcome is dated by CALL-E's `completed_at`, which arrives in the authenticated response rather than in the sender's copy of the body, and a result published more than 15 minutes earlier comes through as `needs_human` with `notification_fresh: false`. At-most-once is not verifiable here: a duplicate inside that window is indistinguishable from a legitimate redelivery, and a Zapier trigger has no durable storage to remember a call id in (`z.cursor` belongs to polling triggers, and improvising dedup on an undocumented channel would read as a control while being decoration). So the `Call Completed` trigger reports the full authenticated outcome for routing, alerting and review, and **never sets `is_actionable: true`**; `Find Call Result`, an authenticated pull with no webhook to distrust, is the path for a step that writes. The resumable callback behind `Place Call and Wait` needs neither mitigation: Zapier mints its URL per waiting run, resumes that run once, and the step rejects any body naming a call other than the one it started.
- **Dry run cannot dial, and is the default.** Only an explicit negative (`false`, `'false'`, `0`, `'0'`) places a real call. Everything else previews - including absence: unset, `null`, or an empty string. A field nobody set, a field a Zap mapped from a blank upstream value, and a field that predates the input existing all arrive identically, and none of them is a person choosing to dial. Dry run makes zero network requests and never generates a callback URL, so it cannot strand a Zap.
- **Region and locale are never inferred** from the phone number, per this repository's design principles. They are optional inputs, omitted from the payload when not supplied.
- **Redaction.** Phone numbers are masked in previews, outputs, logs, and error messages, including numbers written in free text inside a call summary or transcript. Identifiers such as `call_id` are deliberately left intact so reconciliation still works.
- **Result schemas are validated against an allowlist**, not a blocklist, so anything outside CALL-E's documented supported subset is rejected in the Zap editor rather than after a call has been paid for.

## Testing

`npm install && npm test` runs 354 tests across 25 files against a bundled fake CALL-E server. No credentials, no network, no real calls, per this repository's Principle 7.

The skill's linter carries its own 25 tests, run with `node skills/calle-script-advisor/scripts/check-call-script.test.mjs` - no dependencies and no network either.

`test/fixtures/` holds three committed payloads - a clean success, a provider-reported success carrying no answer, and a call carrying a revocation - which `test/fixtures.test.js` runs through the real classifier, so they are executable evidence rather than documentation nobody runs.

`test/e2e-app.test.js` additionally drives the real app definition through `zapier-platform-core`'s `createAppTester`, exercising the actual `beforeRequest`/`afterResponse` middleware chain rather than calling `operation.perform` directly.

`npx zapier-platform-cli validate --without-style` reports the project structurally sound.

**Known limitation:** `z.generateCallbackUrl()` under `createAppTester` targets a fixed Zapier-hosted URL that cannot be redirected to the local fake server, so the live HTTP delivery leg of the webhook is exercised by Zapier in production, not by this test suite.

## Live verification

Three real outbound calls were placed against `https://api.heycall-e.com` on
2026-08-02, and all three operations were exercised against production.

| Scenario | Platform result | Integration verdict |
| --- | --- | --- |
| Recipient answered | terminal `completed`, confidence high 0.95 | `confirmed`, `is_actionable: true` |
| Recipient declined | terminal `failed`, `DECLINED (Hangup by: user)` | `failed`, `is_actionable: false` |
| Recipient declined, started via `Start Call (No Wait)` | terminal **`completed`** | `review_required`, `is_actionable: false` |
| Region implies a language requirement | HTTP 422 `call_not_ready` with `details.questions` | clarification surfaced to the user, no call spent |
| `Find Call Result` re-read of earlier calls | matched the live callback path exactly | same disposition, one classifier |

The third row is the one that matters most. **A call the recipient declined was
reported by the platform as `status: completed`.** An integration that branches
on `status === 'completed'` - the obvious thing to write - would have treated a
refused call as a success and let a workflow act on it. This one did not,
because `completed` alone is not sufficient: it also requires `task_completed`,
a high confidence label, and a structured result that validates.

The declined-call case in row two is informative for the same reason: CALL-E
returned an extracted `structured_result`, and the integration still refused to
mark the call actionable because `task_completed` was false.

No writeback recipe was run live, and no phone number, API key, or call id is recorded in this repository.

## Platform behavior observed live

Documented in the plugin README so other integrators do not have to rediscover it.

- A call task's top-level `status` read `queued` while the call was in flight, whereas `GET /v1/calls/{id}/events` reported the real progression (`call.started`, `Call is ringing`, `Call connected`, `Callee said`, `Call ended`). Use the events endpoint for live progress; do not infer failure from a non-terminal `status`.
- CALL-E retries after a declined call, so a declined call takes materially longer to reach a terminal state than an answered one.
- A call task for a Vietnamese number was rejected at creation with HTTP 422 and error code `call_not_ready`, carrying a `details.questions` array. CALL-E asks before dialing rather than guessing. This integration surfaces those questions to the user instead of reporting a generic failure, and the README explains that setting `Region` and `Locale` avoids the round trip.

## A note on the SDK

The integration calls the Developer API through Zapier's `z.request` rather than bundling `@call-e/calle`. Zapier's middleware owns authentication injection, error translation, throttling, and request logging; bundling a separate HTTP client inside a Zapier app bypasses all of it. Authentication and error handling are wired as `beforeRequest`/`afterResponse` middleware in `index.js` and covered by the end-to-end tests.

## Scope

Two new directories, `plugins/zapier-calle/` and `skills/calle-script-advisor/`. One row added to the table in `plugins/README.md`. Three lines added to the root `README.md` - one skill bullet, one plugin row, one safety-pattern bullet. No existing file is otherwise modified.


Ray-56:
Two P1 issues remain. Dry Run defaults to false and an undefined field is explicitly live, so a newly configured or tested Zap can place a call without a preview-to-live opt-in transition. The completion callback is also unauthenticated: matching an attacker-supplied call ID does not authenticate origin, yet forged fields can become `confirmed` and `is_actionable`. Verify signatures or refetch the authoritative call before producing an actionable result.

Ray-56:
Current blockers:\n\n- The integration defaults dry_run to false, and a missing value is interpreted as live, so a newly configured action can place a call without an explicit preview-to-live opt-in.\n- Both the static Call Completed trigger and the resumable callback trust unsigned request bodies. Matching a caller-supplied call ID is not authentication, yet forged payloads can become confirmed, qualified, and is_actionable=true.\n\nPlease default to no-call and require authenticated provider reconciliation before any webhook-derived outcome can become actionable.

TyrusRC:
Both blockers are fixed in v1.2.0. Reproducing each first, since the second one
was right for a reason I had not framed correctly in the code.

## 1. Dry Run defaulted to live

Confirmed. `dry_run` shipped with `default: 'false'`, and `isDryRun` returned
`false` for `undefined`, `null` and `''` - so an unset field, or one a Zap
mapped from an empty upstream value, dialled a real person with nobody having
chosen to leave preview mode.

Inverted. The field now defaults to `'true'`, and only an explicit negative
places a call: `false`, `'false'` in any case, `0`, `'0'`. Everything else
previews, absence included. Tests: `start-call > places no call when dry_run was
never set`, `> places no call when dry_run maps to an empty value`,
`isDryRun > treats an absent or blank value as a dry run rather than as
permission to dial`, and the same assertion through the real runtime in
`e2e-app.test.js`.

## 2. Unsigned webhook bodies became actionable

Confirmed, and your framing of the id check was the useful part. I had it in the
code as verification. It is not: the id travels inside the same untrusted body,
so an id an attacker already knows is an id they can echo back, and everything
behind it - `task_completed`, `completion_confidence`, `structured_result`, and
therefore `confirmed`/`is_actionable` - was whatever the sender typed.

Since the transport cannot be authenticated (CALL-E publishes no signing secret
and no subscription handshake), the content is authenticated instead. Neither
webhook surface classifies a delivered body any more. The body is treated as a
notification naming a call; that call is read back with `GET /v1/calls/{id}`
over the connection's own API key, and the authenticated response is what the
classifier sees. New module `lib/reconcile.js`, and a `verified` field on every
output.

The two lookup failure modes are handled differently, because they mean opposite
things:

| Situation | `Place Call and Wait` | `Call Completed` trigger |
| --- | --- | --- |
| Body does not name the call this step started | `needs_human`, no lookup spent | n/a |
| CALL-E has no such call on this connection | `needs_human` | Does not fire at all |
| CALL-E unreachable / erroring | `needs_human`, `verified: false` | `needs_human`, `verified: false` |
| CALL-E returns the call | classified from CALL-E's record | classified from CALL-E's record |

A call the connection cannot see is not the Zap's call - forged, stale, or wired
to another CALL-E project - so the trigger emits nothing rather than manufacture
a review item an attacker could flood. A lookup that merely failed is the
opposite problem: a real outcome must not vanish because the API blinked, so it
comes through unverified and non-actionable where a human sees it.

The id checks in `performResume` stayed, demoted to what they actually are - a
cheap filter that rejects a callback not even claiming to be about this call
before an API request is spent on it. The comment says so now.

Worth noting for the `Call Completed` trigger specifically: an attacker can
still choose the event *type* in the envelope. That is deliberate and safe -
the data being classified is the authenticated record, so a hostile type can
only push the verdict toward less actionable (an unrecognized type resolves to
`needs_human`). It cannot manufacture a success for a call CALL-E does not
report as one.

The transferable write-up in `docs/fail-closed-dispositions.md` §11 was rewritten
around this, since "match the id" was exactly the advice it was giving. It now
says why that is not enough and what to do instead, including the signature path
for providers that do offer one.

I have also asked CALL-E support whether webhook signing exists or is planned;
if it ships, the signature check replaces the extra read and the re-read stays
as the fallback.

Ray-56:
One blocker remains in the static webhook trigger: an authenticated re-fetch proves the call record is real, but it does not prove the unsigned notification is authorized or fresh. Anyone who learns the webhook URL and a valid call ID can POST that ID repeatedly (with fresh envelope IDs) and retrigger downstream Zap actions for an old authoritative result. Please make unsigned/replayed notifications non-actionable unless freshness/ownership can be verified server-side, or remove the actionable static-trigger path.

TyrusRC:
Fixed in v1.4.0. You are right, and the distinction you drew is the one I had
collapsed: the re-fetch authenticates the *record*, and I had been treating that
as if it also authenticated the *delivery*. It does not. A known URL plus a known
call id re-fetches the same genuine result every time it is posted, so nothing in
that loop tells the tenth delivery from the first, and the downstream step runs
again each time against an outcome that happened once.

Splitting the ask into its two properties is what decided the fix, because one of
them is reachable here and the other is not.

## Freshness - verifiable server-side, so it is now verified

`CallTask.completed_at` is required in the OpenAPI schema, typed `string|null`,
and documented as "when the complete terminal call result was published", null
until then. It arrives in the authenticated response, not in the sender's copy of
the body, so a replay cannot make it recent.

New module `lib/notification-freshness.js`. An outcome CALL-E published more than
15 minutes ago is reported as `needs_human` with `notification_fresh: false`, so a
replayed old success never arrives as `confirmed`. Absent, unparseable, or
future-dated (beyond two minutes of clock skew) all count as "not established" and
fail the same way.

Staleness only *downgrades a success*. `failed`, `review_required` and the rest
keep their own disposition: they are already non-actionable, they say something
more useful than `needs_human` does, and every legitimate late redelivery would
otherwise trip the rule. What must not survive a replay is `confirmed`, since
branching on it is exactly what the docs tell a Zap to do.

## At-most-once - not verifiable here, so the trigger no longer claims it

A duplicate landing *inside* the freshness window is indistinguishable from a
legitimate redelivery. Suppressing it means remembering call ids, and a Zapier
trigger has nowhere durable to remember them: `z.cursor` is a polling-trigger
channel, and Zapier's deduplication docs cover polling only and say nothing about
hook deliveries. I could have improvised dedup on `z.cursor` anyway, but a control
that silently no-ops is worse than an absent one, because it reads as a control.

So I took your second option for the part I cannot prove. **`Call Completed` never
sets `is_actionable: true`**, on any outcome. It still reports the full
authenticated disposition, summary, transcript and structured result - that is
genuinely useful for routing, alerting and review - but the field that gates side
effects is false, and the trigger's help text sends users to `Find Call Result`
(an authenticated pull, with no webhook to distrust) before a step that writes,
and tells them to key what they write to `call_id`.

Freshness is therefore a bound, not a fix, and the README says so: the trigger is
documented as at-least-once.

## Why `Place Call and Wait` did not need either change

I checked it rather than assuming, and it is structurally protected: Zapier mints
the callback URL per waiting run, resumes that run once, and `performResume`
already rejects any body naming a call other than the one that step started. The
worst a replay achieves there is resuming a run that was already waiting, with
whatever CALL-E truthfully reports at that moment. Applying a freshness window to
it would only add a way to drop legitimate late callbacks. The README now
contrasts the two surfaces explicitly, since "same transport, same missing
signature" hides quite different exposure.

This is all scaffolding for a missing signature. An HMAC over the raw body plus a
timestamp header would authenticate the notification directly and retire both the
freshness window and the actionability restriction; that is already question 2 in
the support request I have open with CALL-E, and if it ships I will replace both
with the signature check.

PR #69, opened 2026-08-02
fix(call-on-behalf): stop reading a status or a refusal as evidence

zkasuran:
## Summary

`apps/typescript/call-on-behalf` treats a definite error on a retried create as
proof that no call was ever created. It is not proof. This fixes that, plus two
more of the same shape in the same file.

When a create leaves the call unknown (no reply, a timeout, a 429, a 5xx) the app
re-issues the same idempotency key. That part is right: it returns the call CALL-E
already holds for that key and can never ring a second time. The bug is what
happened to a second failure. `src/errand.ts` branched on the class of that second
answer, and a definite 4xx was filed as a refusal, which the report renders as
outcome `api_error` with the next step "CALL-E refused to create the call, so no
call was placed and nothing was said on your behalf."

A refusal to the reconciliation can be decided before the idempotency lookup ever
happens. A rejected key answers 401, a spent balance answers 402, a payload the
gateway will not take answers 400, and none of those looked at whether a call
already exists under that key. So the app told somebody nothing was said on their
behalf while a call may have been live on a clinic's line, giving their name and
their date of birth. Any second failure now leaves the call `outcome_unknown` with
the whole chain recorded as `service_unavailable, then unauthorized`. The report
says the outcome is not known, keeps the commitment `unconfirmed`, claims neither
disclosure list and tells the person that running the same errand file again reads
that same call back rather than ringing anybody.

Found while hardening #42 against `Ray-56`'s round-four review, which rejected this
exact inference there. The same reasoning was already merged here in #43, so it is
in shipped code. The blast radius is smaller than #42's: one call, no release duty,
and nobody else gets called on the strength of the wrong answer. It is still a
report telling somebody nothing was said when something may have been.

### Two more of the same shape, in the same file

Found by sweeping `errand.ts` for anything that reads a provider status or an
extraction as an assertion about what happened.

**A call that ended early was read as a call nobody answered.** `failed` and
`canceled` are terminal, so there is a transcript to read, and a line that drops
after the slot was held carries the whole errand. The status was read first, so
that call came back `call_failed` or `not_reached` with "The call did not connect
to a person. Nothing was said on your behalf and nothing was arranged." next to a
transcript holding the questions, the answers and the booking, and next to
`reached_person: true` in the same report. A `no_answer` failure code on a call
somebody plainly answered did the same thing. The transcript is read first now.
The notes name the status as `call_failed` or `call_canceled` with the failure code
beside it, and a call that ended early is never `goal_met`, because it may have
been cut off partway through. A call with nobody on its transcript is still read
from its failure code, so `voicemail` and `not_reached` are unchanged for the case
they are for.

**A refusal CALL-E reported closed the errand.** `declined_by_callee` comes from
the extraction alone, which `docs/scope-and-limits.md` documents and defends as
saying less than the extraction did. It was also settling the commitment, so an
errand whose booking was refused came back `goal_met` with the next step
"Everything you asked was answered. Nothing was agreed, because this errand did
not ask for anything to be agreed.", which is false for every errand that asked
for a booking. A booking that was refused is not a goal met. It now reads
`partially_met` at best and the next step says plainly that nothing is booked.

Review then took that further, six times. It was right every time. A reported
refusal now needs a turn behind it rather than a label in the notes, because the next step is an
instruction and an instruction built on the extraction alone is the thing this app
exists to avoid. That turn has to be answering the arrangement: refusal words are
not specific to it, so "no, we do not take that plan" turns down a question and says
nothing about a booking the same call may have held. It also has to be about the time the
extraction reported, when it reported one, because two times proposed on one call is
otherwise a way for a no to the second to be read as a no to the first. That is the same
offered-time binding the agreement side uses, so a refused arrangement nobody can match to
the reported one fails closed to `unconfirmed` with the refusal quoted as what was actually
turned down. When one turn agrees and another
refuses the same arrangement the report stands behind neither and says a person has
to read the transcript and call. All three bindings look backwards only, because a turn
cannot be an answer to something nobody had said yet. And what counts as the caller
raising the arrangement is a genuine request or a proposal that names the offered time,
not a keyword and not a bare weekday in a statement, so "our appointment desk is open
Monday" cannot make an answer to another question look like a refused booking. The
request and its booking words also have to sit in the same clause of the turn, because a
turn is not one thing put to the callee: the question mark at the end of "I am calling to
book an appointment. Do you accept Aetna?" is an ask about insurance and nothing else.
Where one turn does put two things to them the answer belongs to whichever was asked
last.

### The shape is copied, not invented

`src/gate.ts` in the merged phone-approval-gate keeps a second create failure
unresolved whatever its class, and `placeCall` in #42 does the same and records the
chain `<first>, then <second>`. This app now matches both, including the chain and
the `call_<status>` reason name #42 uses in its notes.

## Verification

Node 22.22.2, from `apps/typescript/call-on-behalf`, no credentials and no
outbound calls:

```text
npx tsc --noEmit                        exit 0, no output
npm test                                142 pass, 0 fail, 0 skipped (113 before)
npm run demo                            exit 0, output byte identical to before
```

Twenty-nine new tests. The twelve report-level ones run end to end over the real
`@call-e/calle` client against the local fake server in `fake/calle-server.ts`.
Each one that changes a behaviour was run against the source it fixes and fails there:

| Test | Before | After |
| --- | --- | --- |
| a definite refusal on the reconcile is not proof the call was never made | `api_error` | `outcome_unknown` |
| a call that ended early still reports the conversation it carried | `call_failed` | `partially_met`, booking kept |
| the transcript says who was on the line, not the failure code beside it | `not_reached` | `voicemail` |
| a refusal CALL-E reported does not make the errand done | `goal_met` | `partially_met`, `unconfirmed` with nothing behind the claim |
| a refusal the callee actually voiced is reported as a refusal | not checked | `declined_by_callee`, the turn quoted |
| a refusal of one of the questions is not a refusal of the errand | `declined_by_callee` | `unconfirmed`, what was turned down quoted |
| an agreement and a refusal in the same call settle nothing | `committed` | `unconfirmed`, both turns quoted, no code |
| a refusal of another proposed time is not a refusal of the one reported | `declined_by_callee` | `unconfirmed`, the option they did turn down quoted |
| a statement about the appointment does not turn a refused question into a refused errand | `declined_by_callee` | `unconfirmed`, the plan refusal quoted |
| a question asked in the same turn as a booking line is what the answer answers | `declined_by_callee` | `unconfirmed`, the plan refusal quoted |
| a booking nobody asked for is not agreed and its reference number is dropped | `committed`, code 4471 | `unconfirmed`, the booking quoted, code dropped |

One report-level test passes on both trees on purpose. "a refusal of the time the
extraction reported is still a refusal" pins the other direction of the offered-time
binding, so nobody can later tighten it into dropping a refusal of the reported slot.

The first runs 401, 403, 400 and 402 through the reconcile path and asserts the
unknown outcome, the recorded chain, no call id, `unconfirmed`, a next step that
does not say nothing was said and no second call placed. No assertion in any of
them accepts either outcome. The fake server gained one knob, `createErrors`, so
one create attempt can be answered differently from the next. No existing test was
weakened, deleted or made tolerant and the demo output did not move. Seventeen unit tests
in `test/read.test.ts` pin the bindings themselves: which claim a turn is evidence for,
the time it has to be about, that nothing said after a turn can be the thing it was
answering and what counts as the caller raising the arrangement at all.

AI assistance (Claude, Anthropic) was used in developing this change. The design,
review and verification were done by the author. Verified locally before
submitting: the four commands above, plus each report-level test that changes a
behaviour run against the source it fixes to confirm it fails there.

## Type

- [ ] New skill
- [ ] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

No box says bug fix in a merged app, so that is the closest one. The change is
behavioral, in `src/errand.ts`, with the README and `docs/scope-and-limits.md`
updated to match.

## Checklist

### DIFF
--- a/apps/typescript/call-on-behalf/README.md
+++ b/apps/typescript/call-on-behalf/README.md
@@ -65,6 +65,12 @@ The extraction proposes. The transcript decides.
   the first. A refusal aimed elsewhere is quoted as what was actually turned
   down. When one turn agrees and another refuses the same arrangement, the report
   stands behind neither and says a person has to read the transcript and call.
+- Raising the arrangement means asking for it or proposing it. A statement that only
+  carries the word appointment or confirm raises nothing, so an answer to the
+  question before it stays an answer to that question. A time the caller proposes
+  without a question mark does count, because that is a time the callee can say yes
+  to. It is one rule and all three bindings use it: the agreement, the refusal and
+  the confirmation code that belongs to the agreement.
 - A time no callee turn named is not read back as an offer. CALL-E's own note
   is printed with its name on it. Nothing in the report is both unchecked and
   unlabelled.
@@ -158,7 +164,7 @@ Node 20 or later.
 cd apps/typescript/call-on-behalf
 npm install
 npm run check   # tsc --noEmit
-npm test        # 126 tests, no credentials, no outbound calls
+npm test        # 135 tests, no credentials, no outbound calls
 npm run demo    # the six cases against the local fake CALL-E
 ```
 
--- a/apps/typescript/call-on-behalf/README.md
+++ b/apps/typescript/call-on-behalf/README.md
@@ -56,6 +56,13 @@ The extraction proposes. The transcript decides.
   as booked and the note quotes whatever they did say. The confirmation code
   belongs to the agreement, so a code nobody read out is dropped even when the
   agreement stands.
+- A refusal is a claim of the same size, so it is checked the same way. A reported
+  `declined_by_callee` needs a turn that refuses the arrangement, which means a turn
+  answering the arrangement rather than one of the questions: "no, we do not take
+  that plan" turns down a question and says nothing about a booking the same call
+  may have held. A refusal aimed elsewhere is quoted as what was actually turned
+  down. When one turn agrees and another refuses the same arrangement, the report
+  stands behind neither and says a person has to read the transcript and call.
 - A time no callee turn named is not read back as an offer. CALL-E's own note
   is printed with its name on it. Nothing in the report is both unchecked and
   unlabelled.
@@ -149,7 +156,7 @@ Node 20 or later.
 cd apps/typescript/call-on-behalf
 npm install
 npm run check   # tsc --noEmit
-npm test        # 87 tests, no credentials, no outbound calls
+npm test        # 123 tests, no credentials, no outbound calls
 npm run demo    # the six cases against the local fake CALL-E
 ```
 
--- a/apps/typescript/call-on-behalf/README.md
+++ b/apps/typescript/call-on-behalf/README.md
@@ -65,12 +65,14 @@ The extraction proposes. The transcript decides.
   the first. A refusal aimed elsewhere is quoted as what was actually turned
   down. When one turn agrees and another refuses the same arrangement, the report
   stands behind neither and says a person has to read the transcript and call.
-- Raising the arrangement means asking for it or proposing it. A statement that only
-  carries the word appointment or confirm raises nothing, so an answer to the
-  question before it stays an answer to that question. A time the caller proposes
-  without a question mark does count, because that is a time the callee can say yes
-  to. It is one rule and all three bindings use it: the agreement, the refusal and
-  the confirmation code that belongs to the agreement.
+- Raising the arrangement means asking for it or proposing the time itself. A
+  statement that only carries the word appointment or confirm raises nothing, so an
+  answer to the question before it stays an answer to that question. That holds even
+  when the statement also names some unrelated weekday: "our appointment desk is open
+  Monday" offers no slot. A time the caller proposes without a question mark counts
+  only when it is the time the report would print, because that is a time the callee
+  can say yes to. It is one rule and all three bindings use it: the agreement, the
+  refusal and the confirmation code that belongs to the agreement.
 - A time no callee turn named is not read back as an offer. CALL-E's own note
   is printed with its name on it. Nothing in the report is both unchecked and
   unlabelled.
@@ -164,7 +166,7 @@ Node 20 or later.
 cd apps/typescript/call-on-behalf
 npm install
 npm run check   # tsc --noEmit
-npm test        # 135 tests, no credentials, no outbound calls
+npm test        # 138 tests, no credentials, no outbound calls
 npm run demo    # the six cases against the local fake CALL-E
 ```
 
--- a/apps/typescript/call-on-behalf/README.md
+++ b/apps/typescript/call-on-behalf/README.md
@@ -65,14 +65,18 @@ The extraction proposes. The transcript decides.
   the first. A refusal aimed elsewhere is quoted as what was actually turned
   down. When one turn agrees and another refuses the same arrangement, the report
   stands behind neither and says a person has to read the transcript and call.
-- Raising the arrangement means asking for it or proposing the time itself. A
-  statement that only carries the word appointment or confirm raises nothing, so an
-  answer to the question before it stays an answer to that question. That holds even
-  when the statement also names some unrelated weekday: "our appointment desk is open
-  Monday" offers no slot. A time the caller proposes without a question mark counts
-  only when it is the time the report would print, because that is a time the callee
-  can say yes to. It is one rule and all three bindings use it: the agreement, the
-  refusal and the confirmation code that belongs to the agreement.
+- Raising the arrangement means asking for it or proposing the time itself, in one
+  clause. A statement that only carries the word appointment or confirm raises
+  nothing, so an answer to the question before it stays an answer to that question.
+  That holds even when the statement also names some unrelated weekday: "our
+  appointment desk is open Monday" offers no slot. A time the caller proposes without
+  a question mark counts only when it is the time the report would print, because
+  that is a time the callee can say yes to. The request and the booking words have to
+  be in the same clause, so "I am calling to book an appointment. Do you accept
+  Aetna?" asks about insurance and nothing else. Where a turn puts two things to them
+  the answer belongs to the one asked last. It is one rule and all three
+  bindings use it: the agreement, the refusal and the confirmation code that belongs
+  to the agreement.
 - A time no callee turn named is not read back as an offer. CALL-E's own note
   is printed with its name on it. Nothing in the report is both unchecked and
   unlabelled.
@@ -166,7 +170,7 @@ Node 20 or later.
 cd apps/typescript/call-on-behalf
 npm install
 npm run check   # tsc --noEmit
-npm test        # 138 tests, no credentials, no outbound calls
+npm test        # 142 tests, no credentials, no outbound calls
 npm run demo    # the six cases against the local fake CALL-E
 ```
 
--- a/apps/typescript/call-on-behalf/README.md
+++ b/apps/typescript/call-on-behalf/README.md
@@ -64,7 +64,15 @@ The extraction proposes. The transcript decides.
   with: only a terminal status is read as a result. A call still `queued` or
   `in_progress` gets no verdict, no commitment and no privacy finding. Re-running the
   same errand file reads the same call back, because the idempotency key covers the
-  content of the call and has not changed.
+  content of the call and has not changed. A reconciliation that fails proves
+  nothing either, whatever class the second answer was: a definite refusal to it can
+  be decided before the idempotency lookup, so it is no evidence that the first
+  request never landed.
+- A status is not a claim about what was said. A call CALL-E ended as `failed` or
+  `canceled` can still have carried the questions, the answers and the slot, so the
+  report is read from the transcript and the notes name the status. It is not read
+  as everything done either, because a call that ended early may have been cut off
+  partway through.
 
 ## Try it without an account
 
--- a/apps/typescript/call-on-behalf/README.md
+++ b/apps/typescript/call-on-behalf/README.md
@@ -60,7 +60,9 @@ The extraction proposes. The transcript decides.
   `declined_by_callee` needs a turn that refuses the arrangement, which means a turn
   answering the arrangement rather than one of the questions: "no, we do not take
   that plan" turns down a question and says nothing about a booking the same call
-  may have held. A refusal aimed elsewhere is quoted as what was actually turned
+  may have held. It also has to be about the time the extraction reported, when it
+  reported one, so a no to the second slot proposed on a call is not read as a no to
+  the first. A refusal aimed elsewhere is quoted as what was actually turned
   down. When one turn agrees and another refuses the same arrangement, the report
   stands behind neither and says a person has to read the transcript and call.
 - A time no callee turn named is not read back as an offer. CALL-E's own note
@@ -156,7 +158,7 @@ Node 20 or later.
 cd apps/typescript/call-on-behalf
 npm install
 npm run check   # tsc --noEmit
-npm test        # 123 tests, no credentials, no outbound calls
+npm test        # 126 tests, no credentials, no outbound calls
 npm run demo    # the six cases against the local fake CALL-E
 ```
 
--- a/apps/typescript/call-on-behalf/docs/scope-and-limits.md
+++ b/apps/typescript/call-on-behalf/docs/scope-and-limits.md
@@ -90,6 +90,20 @@ this one. The cost is a call where the only precise form of the time or the refe
 in a later turn: that reads as not named, so the report gives less than the extraction
 claimed rather than more.
 
+Raising the arrangement takes two things at once and not one word. The turn has to be
+about the arrangement, which is booking language or the time the extraction reported. It
+also has to put that to the callee, which means an ask or an actual proposal: a question,
+a request or a time they could say yes to. A statement that only carries the word
+appointment, confirm or availability raises nothing. "This appointment is for next week",
+said between a question and its answer, would otherwise make that answer a reply to the
+arrangement, so a no to the insurance question would come back as a refused appointment.
+One rule, used by all three bindings, so it cannot hold on the agreement and not on the
+refusal or the confirmation code. A caller who proposes a time without asking a question
+is still proposing, so "Thursday the thirteenth at nine forty would suit her" is an anchor
+and the tightening does not cost real agreements. It fails closed the other way: an ask
+this does not catch leaves the evidence unbound, so the commitment reads
+`unconfirmed`.
+
 The time is matched as a wall clock, in the forms a transcript carries. When a turn
 names a day or a weekday it has to be the claimed one, so the same time on another
 authorized day is not reported as this one. A turn that names no day is taken as the
@@ -116,8 +130,9 @@ nothing on their own. "No, we do not take that plan" refuses a question and says
 nothing about the booking, which may have been held in the same call. So the turn has
 to be answering the arrangement: what the caller last put to the callee before it has
 to be the arrangement rather than one of the questions. Statements do not change the
-subject, so a caller reading out a date of birth between the two leaves the refusal
-answering the appointment it followed. Then the time is bound too. When the extraction
+subject. A caller reading out a date of birth between the two leaves the refusal
+answering the appointment it followed. So does a statement that only mentions the
+appointment. Then the time is bound too. When the extraction
 reports a datetime, that datetime has to have been named by the moment the callee spoke,
 in the refusing turn or in the caller's proposal before it, which is what the agreement
 side already requires. Two times
--- a/apps/typescript/call-on-behalf/docs/scope-and-limits.md
+++ b/apps/typescript/call-on-behalf/docs/scope-and-limits.md
@@ -92,17 +92,20 @@ claimed rather than more.
 
 Raising the arrangement takes two things at once and not one word. The turn has to be
 about the arrangement, which is booking language or the time the extraction reported. It
-also has to put that to the callee, which means an ask or an actual proposal: a question,
-a request or a time they could say yes to. A statement that only carries the word
-appointment, confirm or availability raises nothing. "This appointment is for next week",
-said between a question and its answer, would otherwise make that answer a reply to the
-arrangement, so a no to the insurance question would come back as a refused appointment.
-One rule, used by all three bindings, so it cannot hold on the agreement and not on the
-refusal or the confirmation code. A caller who proposes a time without asking a question
-is still proposing, so "Thursday the thirteenth at nine forty would suit her" is an anchor
-and the tightening does not cost real agreements. It fails closed the other way: an ask
-this does not catch leaves the evidence unbound, so the commitment reads
-`unconfirmed`.
+also has to put that to the callee, which is one of two things and not a third: a genuine
+request, meaning a booking turn phrased as a question or a request, or a concrete proposal,
+meaning a turn that names the offered time itself. A statement that only carries the word
+appointment, confirm or availability raises nothing, and neither does one that adds an
+unrelated weekday: "our appointment desk is open Monday", said between a question and its
+answer, offers no slot, so it must not make that answer a reply to the arrangement. That is
+why the proposal half is the offered time and not any weekday, and why a first-person
+purpose statement such as "I am calling to book an appointment" is not an ask: it says why
+the caller rang and puts no question to the callee. A time the caller states rather than
+asks still counts when it is the offered time, so "Thursday the thirteenth at nine forty
+would suit her" is an anchor and the tightening does not cost real proposals. One rule, used
+by all three bindings, so it cannot hold on the agreement and not on the refusal or the
+confirmation code. It fails closed the other way: an ask this does not catch leaves the
+evidence unbound, so the commitment reads `unconfirmed`.
 
 The time is matched as a wall clock, in the forms a transcript carries. When a turn
 names a day or a weekday it has to be the claimed one, so the same time on another
--- a/apps/typescript/call-on-behalf/docs/scope-and-limits.md
+++ b/apps/typescript/call-on-behalf/docs/scope-and-limits.md
@@ -93,7 +93,10 @@ next step says "another time" rather than a time nobody said. An agreement with 
 time at all cannot be checked against the windows, so it reads `unconfirmed` too.
 CALL-E's own note is printed with its name on it, because nothing here checked that
 sentence. The one claim left standing on the extraction alone is
-`declined_by_callee`, which says less than the extraction did rather than more.
+`declined_by_callee`, which says less than the extraction did rather than more. It
+is labelled in the notes as CALL-E's reading and it does not settle the errand: a
+booking that was refused is not a goal met, so the outcome is `partially_met` at
+best and the next step says plainly that nothing is booked.
 
 When nothing supports the claim, the report says not answered or `unconfirmed` and
 notes that CALL-E claimed otherwise. It will sometimes be too strict, in two ways.
@@ -110,11 +113,14 @@ lost connection, a timeout, a server error. The call may be ringing right now.
 
 In that case the app reconciles under the same idempotency key first, which returns
 the existing call rather than placing a second one. If it still cannot read the
-call, the outcome is `outcome_unknown` with exit code 40. The report says the
-outcome is not known instead of saying nothing was said. Running the same errand
-file again reads the same call back, because the key covers the content of the
-call. Editing the file first makes it a different call, so the next step says not
-to.
+call, the outcome is `outcome_unknown` with exit code 40. Getting the call back is
+the only thing that settles it. A refusal to the reconciliation, a definite one such
+as 401 or 402 included, can be decided before the idempotency lookup ever happens,
+so it says nothing about the request that went unanswered and the call stays
+unknown. The report says the outcome is not known instead of saying nothing was
+said. Running the same errand file again reads the same call back, because the key
+covers the content of the call. Editing the file first makes it a different call, so
+the next step says not to.
 
 A call CALL-E has not finished with lands in the same place. Only a terminal status
 is read as a result: `completed`, `failed` or `canceled`, which is every terminal
@@ -127,8 +133,18 @@ verdict, no commitment and no privacy finding. Reading a call in flight as a res
 is how a call still ringing gets reported as an errand that is done. The next step
 is the same: run the same errand file again and it reads that same call back.
 
-A refusal is different. When CALL-E declines to create the call at all, nothing was
-placed, the outcome is `api_error` and the report can say so plainly.
+A refusal on the first attempt is different. CALL-E declined to create the call and
+nothing was placed, so the outcome is `api_error` and the report can say so plainly.
+
+A call that ended before the conversation did is the other way round. `failed` and
+`canceled` are terminal, so there is a transcript to read. A line that dropped after
+the slot was held carries the whole errand: the questions, the answers and the
+agreement. That status is not read as nobody having answered. What was said comes
+from the transcript and the notes name the status as `call_failed` or
+`call_canceled` with the failure code beside it. The outcome is never `goal_met`,
+because a call that ended early may have been cut off partway through. A call with
+nobody on its transcript is still read from its failure code, which is where
+`voicemail` and `not_reached` come from.
 
 ## Who you may call
 
--- a/apps/typescript/call-on-behalf/docs/scope-and-limits.md
+++ b/apps/typescript/call-on-behalf/docs/scope-and-limits.md
@@ -90,22 +90,39 @@ this one. The cost is a call where the only precise form of the time or the refe
 in a later turn: that reads as not named, so the report gives less than the extraction
 claimed rather than more.
 
-Raising the arrangement takes two things at once and not one word. The turn has to be
-about the arrangement, which is booking language or the time the extraction reported. It
-also has to put that to the callee, which is one of two things and not a third: a genuine
-request, meaning a booking turn phrased as a question or a request, or a concrete proposal,
-meaning a turn that names the offered time itself. A statement that only carries the word
-appointment, confirm or availability raises nothing, and neither does one that adds an
-unrelated weekday: "our appointment desk is open Monday", said between a question and its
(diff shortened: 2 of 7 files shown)


Ray-56:
The fix still leaves one extraction-only refusal path. When `commitment_made` is `declined_by_callee`, `errand.ts` sets the commitment from structured extraction without transcript evidence, and `nextStep()` then states as fact that the callee would not arrange it. The later note labels this as the model reading, but the operational instruction has already made the unsupported claim. Treat extraction-only refusal as unconfirmed unless recipient turns support it.

zkasuran:
Fixed at `04190bd`. You were right and the gap was worse than a wording problem: the whole point of the last round was that a claim needs a turn behind it, and the refusal path was the one place still exempt.

**What it does now.** `declined_by_callee` needs a turn that plainly refuses the errand, or the business refusing to deal with an automated caller at all. Without one the commitment reads `unconfirmed` and the next step says nothing is settled either way, rather than picking a side the transcript does not support. With one, the note quotes the turn instead of admitting nothing was checked.

**One thing I changed beyond the fix.** `next_step` now takes which claim went unsupported. Folding a refusal into the existing `unconfirmed` message would have meant weakening "treat nothing as booked" into something vague enough to cover both, and that message is precise for a reason. So the agreement case keeps its wording and the refusal case gets its own: "CALL-E reported that they would not arrange it and the transcript does not show anybody refusing, so nothing is settled either way."

**The refusal patterns are deliberately narrow.** Missing a real refusal costs a softer report that says the outcome is unknown. Inventing one costs the person a fact about their own errand, so the checks are built to be wrong in the first direction, same as the existing answer support checks.

**A test I had to change rather than add.** `a refusal CALL-E reported does not make the errand done` was asserting the old behaviour on a transcript where the callee offers a Thursday slot and refuses nothing at all. That fixture is exactly the case you describe, so it now pins the safe reading, and a companion test covers a refusal the callee actually voiced ("I am afraid we cannot book that over the phone for somebody else") landing as `declined_by_callee` with the quote in the notes.

`docs/scope-and-limits.md` no longer claims `declined_by_callee` is the one thing standing on the extraction alone, because it is not any more.

Verified: `tsc --noEmit` clean, tests 117 -> 118, all passing.

emezac:
## 🤖 MaaS PR Reviewer (Paritok GPU Compressed Context)

📌 **Resumen Arquitectónico del PR**

Este Pull Request se presenta como una auditoría de código Ruby. Sin embargo, al examinar los archivos modificados, se observa que el código fuente corresponde a una aplicación **TypeScript** (archivos `.ts`) y documentación en Markdown (`.md`). Los cambios se centran en refinar la interpretación de estados y rechazos dentro de un sistema de gestión de llamadas o "errands", buscando mayor precisión en la evidencia extraída de las interacciones.

🚨 **Code Smells y Anti-Patrones Detectados**

Como MaaS PR Reviewer Bot, mi especialización se centra exclusivamente en el análisis de código **Ruby**, aplicando las Principios de Diseño Orientado a Objetos de Sandi Metz (POODR, 99 Bottles of OOP) y las mejores prácticas de Ruby en producción.

He detectado que el código proporcionado en este Pull Request es **TypeScript** y no Ruby. Las reglas de refactorización de Sandi Metz (límites de líneas para clases y métodos, número de parámetros, patrones de controladores, Template Method, Squint Test) son heurísticas y principios profundamente arraigados en el contexto y las convenciones del lenguaje Ruby.

Debido a esta discrepancia fundamental, **no puedo aplicar de manera efectiva ni generar una auditoría significativa de este Pull Request utilizando mi conjunto de reglas y expertise para Ruby.** No sería apropiado ni preciso intentar forzar métricas o patrones específicos de Ruby a un codebase de TypeScript, ya que las convenciones y los patrones idiomáticos varían entre lenguajes.

💡 **Propuesta de Refactor Específica con Código Ruby Clean**

Dado que el código fuente no es Ruby, no es posible proporcionar una propuesta de refactorización relevante en Ruby.

💬 **Comentario Inline Sugerido para GitHub**

No es posible sugerir comentarios inline específicos basados en las reglas de Sandi Metz para Ruby, dado que el código es TypeScript.

---

**Sugerencia Adicional:** Si necesitas una revisión de código Ruby, por favor, proporciona un Pull Request que contenga código Ruby. Estaré encantado de aplicar mi experiencia en Sandi Metz y Ruby para ayudarte a mejorar la calidad de tu código.

---
*Revisado automáticamente por MaaS v2.0 + Paritok GPU Gateway*

Ray-56:
Refusal evidence is still not bound to the requested commitment. `readTranscript()` accepts any broad refusal phrase anywhere in a callee turn, so refusal of an unrelated question can corroborate extraction-only `declined_by_callee` even when the requested booking was accepted. Anchor refusal to the bot request and commitment, and handle conflicting acceptance and refusal evidence as unconfirmed or manual review.

zkasuran:
Fixed at `0e4067a`. Confirmed. It is the same mistake one level in: the last round
bound the refusal to the transcript and left it unbound to the thing the errand
asked for.

## What was wrong

`readTranscript()` returned the first callee turn matching a refusal pattern,
wherever it sat in the call. Those phrases are not specific to an arrangement. "No,
we do not take that plan" matches. It is an answer to one of the errand's own
questions. So an extraction-only `declined_by_callee` could be corroborated by a
refusal of the insurance question in a call where the booking was accepted and read
back with a reference number. `nextStep()` would then tell the person their
appointment was turned down.

## Refusal is anchored the way agreement already was

`refusalEvidence()` in `src/read.ts` replaces the unanchored lookup. It asks what the
caller had last put to the callee before the turn:

- the arrangement, so the turn is refusing the arrangement,
- one of the errand's questions, so the turn is refusing that question and is not
  evidence about the arrangement,
- nothing, when the caller never raised an arrangement at all, which is the same
  refusal the agreement side already applies through `commitmentPromptAt`.

Two details worth naming. Statements are skipped when walking back, because a caller
reading out a date of birth between the question and the answer does not change what
the callee's next turn is answering. And a caller turn that raises the arrangement
counts as the arrangement even when it is also one of the questions, which is the
usual case: "what is the earliest appointment you have" is both. A refusal aimed
somewhere else comes back as `otherQuote` and the report quotes it as what was
actually turned down, so `unconfirmed` does not read as nothing having been said.

`readTranscript()` no longer carries `refusalQuote` at all. An unanchored refusal is
exactly the thing that must not be evidence, so it is not there to be picked up by
whoever touches this next.

## Conflicting evidence settles nothing

When one turn agrees to the arrangement and another refuses it, the report stands
behind neither: the commitment reads `unconfirmed`, the note quotes both turns and the
next step says a person has to read the transcript and call. Reporting the agreement
would book something the callee took back. Reporting the refusal would drop a slot
they may be holding. The confirmation code belongs to an agreement that stands, so it
is dropped there too. This runs before the extraction's own value is read, so it
applies whichever way the extraction leaned.

## Tests

Two at the report level, both failing against `04190bd`:

- "a refusal of one of the questions is not a refusal of the errand" is your case.
  The slot is held and read back with a reference number, the callee turns down Blue
  Shield PPO, the extraction says `declined_by_callee`. The commitment is
  `unconfirmed`, the next step no longer says they would not arrange it, the note
  quotes the insurance refusal as what was turned down and that question is still
  answered `no` out of the same turn.
- "an agreement and a refusal in the same call settle nothing" holds the slot, then
  refuses it after the caller asks again. Extraction says `accepted`. Commitment
  `unconfirmed`, no committed datetime, no confirmation code, both turns quoted.

Three unit tests in `test/read.test.ts` pin the anchor itself: a refusal answering a
question is not evidence, a refusal after a statement still answers the arrangement
before it and a refusal in a call with no arrangement refuses no arrangement.

One existing assertion changed wording only. The note for a supported refusal now
reads "a turn in the transcript refuses it" rather than "the transcript agrees",
which was ambiguous next to an actual agreement.

`docs/scope-and-limits.md` and the README say what the anchor is and what happens when
the evidence conflicts.

Verified from `apps/typescript/call-on-behalf`: `npx tsc --noEmit` clean, `npm test`
123 pass 0 fail 0 skipped (118 before), `npm run demo` exit 0 with all six cases,
`python3 scripts/validate_repository.py` passes, `git diff --check` clean. No
credentials and no real calls were used.

Ray-56:
The latest anchoring fix binds a refusal to the last commitment-shaped prompt, but it still does not bind that refusal to the specific arrangement or datetime reported by the extraction. lastAskBefore returns commitment whenever the bot turn contains generic booking language, even if offered_datetime names a different slot; a refusal of one proposed option can therefore corroborate declined_by_callee for another option or for the overall errand. Apply the same offered-time binding used for agreement evidence, or otherwise fail closed when the refused arrangement cannot be matched.

zkasuran:
Fixed at `e07eda0`. Confirmed. You named the binding I left off. `refusalEvidence()`
had one of the two bindings the agreement side has.

## The asymmetry

`agreementEvidence()` binds twice. The caller has to have raised the arrangement
(`commitmentPromptAt`). Then, when the extraction reports a datetime, that datetime has
to be named around the agreeing turn: `src/read.ts:446`, through `namedAround` at
`src/read.ts:588`, which reads the turn plus one either side. `refusalEvidence()` from
last round carried the first binding only. `lastAskBefore()` (`src/read.ts:488`) returns
`commitment` for any caller turn carrying booking language, so a refusal that followed a
second proposal was read as evidence for whatever `offered_datetime` named. That is your
case exactly: a no to one option corroborating `declined_by_callee` for another option or
for the errand.

## What the binding is now

`src/read.ts:567`, the same branch as `src/read.ts:446`. When `offered` is non-empty and
the refusing turn does not have that datetime named around it, the turn is not evidence
about the reported arrangement. It becomes `otherQuote` and the search carries on, so the
report still quotes what was actually turned down. Reusing `namedAround` rather than
writing a second rule was deliberate: one definition of "named around" means the two
paths cannot drift apart later.

What that does to the report is at `src/errand.ts:387`. With no refusal quote, a claimed
`declined_by_callee` reads `unconfirmed` with reason `refusal`, the next step says nothing
is settled either way and the note names what was turned down instead.

## When the extraction reports no time

`offered` empty means there is nothing to bind to, so the prompt anchor stands alone.
That is the same asymmetry the agreement path already has: `src/errand.ts:370` refuses to
call an agreement `committed` with no time for it, because there is nothing to hold
against the authorized windows, so that reads `unconfirmed` too. Keeping the refusal side
identical beat inventing a third rule for it.

The binding fails closed. It costs something in one place.
`offered_datetime` is contracted as "the date and time that was agreed or offered"
(`src/script.ts:168`), not as the time that was refused. On a refused errand the
extraction can put a time in there that was discussed rather than the one turned down.
Where it does, the refusal reads as being about something else and the commitment comes
back `unconfirmed` with the refusal quoted, which says less than the call held. That is
the direction I would rather be wrong in: a real refusal reported as unsettled costs a
phone call to check, a refusal of one option reported as a refusal of another costs
somebody a slot that still stands. The ordinary single-time call is unaffected, because
the caller's own proposal is one of the turns `namedAround` reads, so a refusal right
after "could you hold Thursday the thirteenth at nine forty" is still bound to it.

## Tests

126 now, 123 before. Two of the three fail against `0e4067a`, checked by putting the old
`src/read.ts` back and running the files:

| Test | At `0e4067a` | At `e07eda0` |
| --- | --- | --- |
| a refusal of another proposed time is not a refusal of the one reported (report level) | `declined_by_callee`, next step "Bayview Family Clinic would not arrange it on this call, so nothing is booked." | `unconfirmed`, "nothing is settled either way", the Wednesday refusal quoted as what was turned down |
| a refusal has to be about the time the extraction claims (unit) | `quote` is "No, we are fully booked on Wednesday. I cannot fit that in." | `quote` is empty, that turn comes back as `otherQuote` |
| a refusal of the time the extraction reported is still a refusal (report level) | `declined_by_callee` | `declined_by_callee` |

The first is your case at report level. The caller asks for Thursday the thirteenth at
nine forty, the callee goes away to check it, the caller then offers Wednesday the twelfth
at two in the afternoon and that is the one the callee turns down. The extraction reports
`declined_by_callee` with Thursday in `offered_datetime`. Every question is still answered
out of that same call, so the binding costs no answers.

The unit test pins both directions plus the empty case. The same turn is evidence for
`2026-08-12T14:00:00-07:00`, the option it does answer. It is evidence again when no
datetime is reported at all.

The third passes on both trees on purpose. It is there so nobody can later tighten this
into dropping a refusal of the reported time: the callee refuses Thursday at nine forty by
name and the report still reads `declined_by_callee` with the turn quoted.

Last round's three unit tests and two report-level tests pass unchanged. So does "a
refusal the callee actually voiced is reported as a refusal", which reports
`offered_datetime: ""` and still lands `declined_by_callee` through the empty-offered
path.

## Verified

From `apps/typescript/call-on-behalf` on Node v22.22.2, no credentials and no real calls:
`npx tsc --noEmit` clean, `npm test` 126 pass 0 fail 0 skipped (123 before), `npm run
demo` exit 0 with output byte identical to `0e4067a`, checked by running the demo on both
trees and diffing, `python3 scripts/validate_repository.py` prints "Repository validation
passed.", `git diff --check` clean. `docs/scope-and-limits.md` and the README describe
both bindings now, including where the time binding fails closed.

Ray-56:
The new offered-time check still has a reverse-order binding bug. refusalEvidence reuses namedAround, which accepts the turn on either side. If the bot proposes Wednesday, the callee refuses Wednesday, and the next bot turn newly proposes Thursday, that following Thursday mention makes the earlier refusal pass as evidence for extracted offered_datetime=Thursday. The report can therefore still say Thursday was refused before the callee answered that proposal. Bind refusal to its own text and the preceding prompt only, or explicitly distinguish a later readback from a new proposal, and add this reverse-order regression case.

zkasuran:
Fixed at `fda7dd2`. Confirmed exactly as you describe it, including which mention
makes it pass.

## The reverse order

`refusalEvidence` bound the time with `namedAround`, which took the turn before, the
turn itself and the turn after. The turn after is the hole. The caller proposes
Wednesday, the callee refuses, the caller then proposes Thursday. That Thursday
mention one turn later made the refusal of Wednesday pass as evidence for
`offered_datetime` = Thursday. The report could say Thursday was turned down before
anybody had put Thursday to them.

## Why the first option and not the second

I tried to keep the read back and tell it apart from a new proposal, because a read
back is sometimes the first precise form of a time in the call. Every rule I could
write for that was a guess. The best one was structural, "a later mention only counts
for a time that was already on the table", and it still fails:

```
bot   Could you hold Thursday the thirteenth at nine forty?
user  Let me look.
bot   If not, could you do Wednesday the twelfth at two in the afternoon?
user  No, we are fully booked on Wednesday. I cannot fit that in.
bot   All right. So Thursday the thirteenth at nine forty then?
```

Thursday was on the table at the first turn, so the refusal of Wednesday binds to
Thursday again. The lexical version, matching read back language against proposal
language, is the unanchored pattern match this module is supposed to be free of.

So the binding looks backwards only: the refusing turn and the caller's turn before
it, which are the only two places the thing a turn was answering can have been said.
That is also what the module header already claimed it did.

## The same window was in two more places

Closing only the refusal would have left the class open, so all three are backward
only now.

`agreementEvidence` had it too, in the same shape:

```
bot   Could you hold Wednesday the twelfth at two in the afternoon?
user  Yes, I have booked her in.
bot   And could you do Thursday the thirteenth at nine forty as well?
```

That agreement to Wednesday passed as an agreement to Thursday.

`codeNamedAround` had it for the reference number. Two bookings in one call and the
second reference read out after the first agreement printed the wrong number against
the first appointment. It is `codeNamedBefore` now, because the old name would be
describing something the function no longer does.

One helper, `upTo`, is all three bindings.

## The cost, so it is on the record

A call where the only precise form of the time or the reference is in a later turn
now reads as not named. The commitment comes back `unconfirmed` and the code is
dropped, so the report says less than the extraction claimed rather than more. The
usual transcript is unaffected: the callee names the time in the turn they agree in,
or the caller names it in the proposal they are answering.

## Tests

Three regression cases, each failing on the previous head and passing on this one:

- `a refusal is not evidence about a time the caller raised after it`
- `an agreement is not evidence about a time the caller raised after it`
- `a confirmation code counts where it was read out, not in a later booking`

Each also asserts the other direction, so the same turn is still evidence for the
option it does answer.

Verified locally on `fda7dd2`: 129 tests pass (126 before), `tsc --noEmit` clean,

Ray-56:
Current blocker at fda7dd2:

- Refusal evidence is still misbound when an intervening bot statement contains a broad commitment keyword. I reproduced: Do you accept Aetna?; This appointment is for next week.; No, we do not take that plan. With no offered datetime, lastAskBefore classifies the statement as a commitment prompt and reports the insurance answer as an appointment refusal. Require an actual arrangement proposal or interrogative anchor, or retain the last real question instead of treating any statement containing appointment, confirm, or availability as a commitment.

zkasuran:
Fixed at `1715445`. Confirmed. The mention that made it pass is the word appointment
sitting inside a statement.

## What the old code did with your three turns

`lastAskBefore` walks back for the last thing the caller put to the callee and it
accepted any bot turn matching the commitment vocabulary. So "This appointment is for
next week." became the anchor one turn back and the refusal after it was read as
answering the arrangement. With no `offered_datetime` there was no second binding to
catch it, so `declined_by_callee` stood and the next step said Bayview would not
arrange it. The question the callee was actually answering, "Do you accept Aetna?",
sat one turn further back and was never reached.

## The rule now

`raisesArrangement` takes two things at once instead of one word:

- the turn is about the arrangement: booking language or the time the extraction reported
- the turn puts it to them: an ask or an actual proposal, which is a time they could say yes to

A statement carrying only `appointment`, `confirm` or `availability` fails the second
half, so it raises nothing and the walk carries on to the last real ask. On your three
turns that is the insurance question, so the refusal comes back as `otherQuote`, the
commitment reads `unconfirmed` and the note quotes what they did turn down.

Both halves are needed. A question mark alone would drop the calls where the caller
proposes rather than asks. A transcript of speech does not always carry the punctuation
either, so the ask forms cover the modal and request shapes a script uses as well as the
question mark. A time is a weekday, a day of the month or a clock time. "Next week" is
not a time anybody can hold, which is why your statement proposes nothing.

## One helper, so it is the class and not the instance

All three bindings go through `raisesArrangement`. The same statement was live on the
other two:

```
bot   Do you accept Aetna?
bot   This appointment is for next week.
user  No, we do not take that plan. I have put her in for Thursday the thirteenth at
      nine forty, reference four four seven one.
```

That read as a real agreement with the caller, `committed`, with 4471 printed against
it, in a call where nobody had been asked to hold anything. It reads `unconfirmed` now,
the booking is quoted as what they did say and the reference is dropped because there is
no agreement for it to belong to.

Neither direction can be displaced. A statement between the arrangement and the answer
leaves the arrangement as the anchor. A statement between a question and its answer
leaves the question as the anchor. Both are pinned.

## Where it fails, out loud

It fails closed: an ask the forms do not catch leaves the evidence unbound, so the
commitment reads `unconfirmed` and the report says less than the extraction claimed.
One boundary is worth naming rather than leaving to be found. A declarative that names
the arrangement and also puts a concrete time in it does anchor, whether it is a fresh
proposal or a read back, because a time on the table is a time they can answer. The
`confirm_existing` fixture is that shape already, "I am calling to confirm the
appointment on Thursday.", with no question mark in it.

## Tests

Six, 129 to 135. Five fail on `fda7dd2`, run against the unmodified source before the
fix:

| Test | At `fda7dd2` | At `1715445` |
| --- | --- | --- |
| a statement about the appointment does not turn a refused question into a refused errand (report) | `declined_by_callee`, "would not arrange it on this call" | `unconfirmed`, "nothing is settled either way", the plan refusal quoted |
| a booking nobody asked for is not agreed and its reference number is dropped (report) | `committed`, code 4471 | `unconfirmed`, the booking quoted, code dropped |
| a statement that mentions the arrangement is not the caller raising it (unit, your three turns) | the refusal bound to the arrangement | `otherQuote`, nothing bound |
| a statement about the arrangement does not displace the last real ask (unit) | bound to the arrangement | bound to the question, still bound to the arrangement where that was the last ask |
| booking language after a statement about the arrangement is not an agreement (unit) | agreement bound at index 2, the code counted | index -1, the code dropped |
| a time the caller states rather than asks is still an arrangement proposal (unit) | bound | bound, on purpose |

The last one passes on both heads. It is there so this tightening cannot quietly lose a
real proposal.

`docs/scope-and-limits.md` carries the rule in its own paragraph and the README says it
once for all three bindings. The README test count was stale at 126 and reads 135 now,
with the suite.

Verified on `1715445`: `npx tsc --noEmit` clean, `npm test` 135 pass 0 fail 0 skipped
(129 before), `npm run demo` exit 0 with output byte identical to `fda7dd2`,
`python3 scripts/validate_repository.py` prints "Repository validation passed."

Ray-56:
Current blocker at `1715445`: the new arrangement binding still treats ordinary statements as proposals. Any `I am calling to ...` line qualifies as an ask, and any booking-keyword statement containing a weekday/time qualifies via `namesATime`. For example, after `Do you accept Aetna?`, `Our appointment desk is open Monday`, followed by `No, we do not take that plan` can still be reported as an appointment refusal even though no slot was offered. Bind only to a genuine request or concrete proposal.

zkasuran:
Fixed at `e05ebca`. Confirmed, on your exact three turns.

## Your example

```
bot   Do you accept Aetna?
bot   Our appointment desk is open Monday.
user  No, we do not take that plan.
```

The middle turn matched the commitment vocabulary on the word appointment and matched
the proposal arm on the weekday Monday, so `lastAskBefore` read it as the arrangement.
With no offered time there was no second binding to catch it, so the insurance refusal
bound to the arrangement and came back as a refused appointment. It reads `unconfirmed`
now, the refusal is quoted as `otherQuote` and the walk carries on to the insurance
question, which is what the no was actually answering.

## Both arms were too loose, both are tightened

You named both. They are the two arms of `raisesArrangement`.

The ask arm accepted `i'?m calling to` and `would like to`. Those are purpose
statements: they say why the caller rang and put no question to the callee, so an
`I am calling to ...` line qualified as an ask when it asks nothing. Both are gone. The
ask forms are interrogative and request shapes now, nothing else. A question mark works.
A modal or auxiliary before a pronoun works (can/could/would/do/does/is/are/have/has
you). So does a `please` or an `any chance`.

The proposal arm was bare `namesATime`, true for any weekday, day of the month or clock.
So a declarative carrying a booking word and any time counted as a proposal. It is
`mentionsDatetime(text, offered)` now: a statement is a proposal only when it names the
time the report would print, not when it mentions some unrelated day. "Our appointment
desk is open Monday" names no offered time, so it proposes nothing. `namesATime` and its
`CLOCK_TIME` table were only ever used here, so they are removed rather than left dead.

So a turn raises the arrangement only when it is about the arrangement (booking language
or the offered time) AND either asks a genuine question or names the offered time
itself. A keyword in a declarative aside is neither.

## Why this is a shift, said plainly

Round six left a residual on the record: "a declarative that names the arrangement and
also puts a concrete time in it does anchor". That was too generous. Your example is
why. Any time was enough, so a booking-keyword statement plus a bare weekday passed. The
rule now is that the concrete time has to be the offered one, the time the extraction
reported and the report would print. A proposal of the slot under discussion still
anchors. A statement that merely mentions a day does not.

## The confirm_existing fixture still holds

The merged fixture is "I am calling to confirm the appointment on Thursday." with
`offered` set to Thursday. It no longer anchors on `i'm calling to`, which is gone. It
anchors on the proposal arm instead, because it names Thursday, which is the offered
time. So it still reports the agreement. That is not luck: a caller who states the
offered time is proposing it, question mark or not, which is the same reason
"Thursday the thirteenth at nine forty would suit her" has always anchored.

## Fails closed

Unchanged in direction. An ask these forms miss leaves the evidence unbound, so the
commitment reads `unconfirmed` and the report says less than the extraction claimed. A
missed real ask costs a phone call to check. Inventing one costs somebody a fact about
their own errand, so the miss is the direction to be wrong in.

## Tests

Three added in `test/read.test.ts`, run against the previous head `1715445`:

| Test | At `1715445` | At `e05ebca` |
| --- | --- | --- |
| a booking-keyword statement with an unrelated weekday is not a proposal (your example) | the refusal bound to the arrangement | `otherQuote`, nothing bound |
| a caller purpose statement is not a question put to the callee | bound via `i'm calling to` | `otherQuote`, nothing bound |
| a concrete proposal of the offered time anchors and so does a real question | passes | passes |

The first two fail on `1715445`. The third pins the other direction, a proposal that
names the offered time and a genuine question with no time, so the tightening cannot
quietly drop a real anchor. The confirm_existing fixture is now called with `offered`
set to Thursday, so it exercises the proposal arm rather than the dropped ask form.

Verified on `e05ebca`, from `apps/typescript/call-on-behalf`, no credentials and no
outbound calls: `npx tsc --noEmit` clean, `npm test` 138 pass 0 fail 0 skipped (135
before), `npm run demo` exit 0 with output byte identical to `1715445`. And
`python3 scripts/validate_repository.py` prints "Repository validation passed." The
README count and the rule paragraph in `docs/scope-and-limits.md` are updated to match.

Ray-56:
Current blocker at e05ebca: arrangement detection still binds request syntax from an unrelated clause to booking words elsewhere in the same bot turn. I reproduced I am calling to book an appointment. Do you accept Aetna? followed by No, we do not take that plan; refusalEvidence reports the insurance answer as an appointment refusal. Bind the interrogative/request form to the booking clause itself, or split turns into clauses before classifying the anchor.

zkasuran:
Fixed at `2178ced`. Confirmed, on your exact two turns.

## Your example

```
bot   I am calling to book an appointment. Do you accept Aetna?
user  No, we do not take that plan.
```

One turn, two things said. The first says why the caller rang and the second asks about
insurance. `raisesArrangement` tested the whole turn, so the booking words in the first
sentence borrowed the question mark from the second and the turn read as the caller
putting the arrangement to them. The insurance refusal then bound to the arrangement and
came back as a refused appointment, in a call where no slot had been put to anybody. It
reads `unconfirmed` now, the refusal is quoted as `otherQuote` and the walk carries on to
the insurance question, which is what the no was answering.

## The turn is split into clauses first

I took both halves of what you asked: the split plus the same-clause binding it makes
possible. `clausesOf` is the only thing that decides what a clause is. Both classifiers
read clauses.

`clauseRaises(clause, text, offered)` needs both halves inside one clause. The clause has
to be about the arrangement, which is booking language or the time the extraction
reported. It also has to put that to them, which is a question or request form in that
same clause or the offered time named in that same clause. Nothing carries across a
break, so a question mark can only ever be the ask for the thing it was asked about.

The break is where one thing said ends and the next begins: a sentence end, a semicolon,
a colon, a comma, a dash or a line break. The comma is in there because speech splices
two utterances with one as often as it writes a full stop. Your example with a comma
instead of a full stop is the same bug. A coordinator counts as a break when a fresh
question opens right after it, "do you accept Aetna and can we book Thursday?", because
that is a second thing put to them inside one sentence. There is a test for the comma
form and one for the coordinator form.

## Where inside the turn also matters

Splitting alone would have left a near neighbour of your example open, so I closed that
too rather than wait for the round. A turn can raise the arrangement and then ask
something else:

```
bot   Could you hold Thursday the thirteenth at nine forty? Do you accept Aetna?
user  No, we do not take that plan.
```

Every clause here is honest. The first is a real request for the slot. Asking only
whether the turn contains one would bind the no to the booking. The no is about insurance
because insurance is what was asked last. `lastAskBefore` now compares the last
clause that put the arrangement to them against the last clause that asked one of the
errand's questions. The later one wins. That is the same rule that already picks the
last turn, applied inside a turn.

A clause that is both counts as the arrangement. "What is the earliest appointment you
have for a routine check-up?" is a question on the errand's list and the arrangement put
to them at the same time. It is the reason the ordinary fixture reports anything at all.

## Two things kept on purpose

The proposal arm asks the whole turn as well as the clause. `mentionsDatetime` refuses a
turn that names a different day. The clause holding the clock often names no day at all,
so reading a clause alone would take "Wednesday, at nine forty" for Thursday at nine
forty. Both have to hold, which keeps the day check the round before last put in.

A question whose words run across a break still counts, ending where the turn ends, so it
is the last thing said. Without that, a question the split happened to cut in half would
go missing and the turn would fall back to the arrangement, which is the reading this
whole rule exists to stop.

## What it costs and the residual

It fails closed the same way. A request whose booking word sits on the far side of a
comma from its question form, "about the appointment, could you do Thursday?", anchors
nothing now, so the commitment reads `unconfirmed` and the report says less than the
extraction claimed. A turn that names the reported time still anchors on the proposal arm
whatever its punctuation, so real proposals are unaffected.

One residual, stated rather than hidden: a single clause that puts two things to the
callee at once reads as the arrangement, because inside one clause there is no order to
read. That is the same tie that makes "what is the earliest appointment you have" work,
so it is deliberate and not an oversight.

## Tests

Four added, run against the previous head `e05ebca`:

| Test | At `e05ebca` | At `2178ced` |
| --- | --- | --- |
| a question in one clause does not make booking words in another an ask (your example plus the comma splice of it) | the refusal bound to the arrangement | `otherQuote`, nothing bound |
| a question asked in the same turn as a booking line is what the answer answers (report level) | `declined_by_callee` | `unconfirmed`, the plan refusal quoted |
| the last thing a turn puts to them is what the next turn answers | the refusal bound to the slot | `otherQuote`, with the reverse order still binding |
| a request and its booking words in one clause still anchor | passes | passes |

The first three fail on `e05ebca`. The fourth pins a genuine request inside a
multi-clause turn, so the tightening cannot quietly disable the ask arm.

Verified on `2178ced`, from `apps/typescript/call-on-behalf`, no credentials and no
outbound calls: `npx tsc --noEmit` clean, `npm test` 142 pass 0 fail 0 skipped (138
before), `npm run demo` exit 0 with output byte identical to `e05ebca`. And
`python3 scripts/validate_repository.py` prints "Repository validation passed." The
README count and the rule paragraphs in the README and `docs/scope-and-limits.md` are
updated to match.

PR #70, opened 2026-08-03
feat(apps): add cloud steward incident callback

mayowaoladosu:
## Summary

- add a preview-first Python app for one consented on-call incident notification
- keep Cloud Steward infrastructure actions pending regardless of phone response
- plan, inspect, run, and query CALL-E through the documented CLI flow
- mask phone numbers and keep confirmation tokens out of output and audit records
- add fake-provider tests and explicit uncertainty/cancellation guidance

## Validation

- python -m unittest discover -s apps/python/cloud-steward-incident-callback/tests -v
- python scripts/validate_repository.py

No live call was placed. Examples use the reserved fictional +1 555 number range.


Ray-56:
The preview/fake-CLI path is mechanically sound, but live execution is not duplicate-safe. The request-ID check and audit append are unlocked and happen only after call run plus status, so concurrent runs or a crash/timeout after provider acceptance can leave no reservation and redial the incident. Plan validation accepts an omitted to_phone, so the promised target inspection is not enforced, and the credential filter misses common secret forms such as Authorization: Bearer values and private-key blocks. Persist an atomic content-bound reservation and exact provider key/plan before run, retain all post-run ambiguity for reconciliation with no retry, require an exact target match, and reject common credential formats before any call.

Ray-56:
Superseding review under the community policy adopted on 2026-09-11: The explicitly validated destination is supplied in the plan request; a missing echoed destination is not evidence of misrouting. Cross-process crash checkpoints and perfect text filters remain follow-ups without a concrete automatic duplicate or credential leak. Maintainer cleanup: Resolved additive README conflict while preserving both catalogs.

PR #73, opened 2026-08-03
feat(apps): add sequential call cascade

lukisch:
## Summary

A single call has one target. This app is a search across several: candidates that must
first be ordered, criteria that are only settled during the conversation, a verdict per
call, and a state that carries across calls — who was asked, and what it failed on.

Four kinds of condition stay apart:

| Kind | Meaning |
| --- | --- |
| `must` | without it there is no deal |
| `boundary` | up to here and no further |
| `concession` | available, but staged |
| `wish` | improves the ranking, blocks nothing |

**A concession is an authorisation, not a hint.** A result that used a tier nobody granted
is rejected exactly like an offer above the boundary — an agent that bought the deal with
authority it was never given exceeded its mandate, and its yes does not count. A later tier
played before an earlier one is rejected too, because an agent that plays every card at
once gives away more than a person would.

An unknown outcome halts the cascade rather than being read as a refusal, an acceptance or
an unreachable line. The next call would otherwise be built on a guess, and a cascade that
guesses spends the user's remaining candidates on a mistake.

The example is dinner; the pattern is not. A dentist appointment within 30 km, a tradesman
with a free slot, a spare part in stock — same mechanism, different subject.

Full application: <https://github.com/lukisch/hungrycall>

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Verification

```
cd apps/python/hungrycall-cascade && python -m pytest -q
........................                                                 [100%]
24 passed in 0.17s

Repository validation passed.
```

Both were run on this branch. The validator was also run on the unmodified upstream
checkout first, so a green result here is a comparison and not just an assertion.

## Real-world side effects

None. There is no live transport and no `--live` flag in this edition — it cannot dial even
if configured to. No account, no API key, no network request of any kind. Every response
comes from the fixture file, and the run is labelled `FIXTURE RUN -- SIMULATED, NO CALL
PLACED` on its first line.

## Cancellation

One pass, in order. There is no scheduler and no retry, so nothing runs later or repeats by
itself. Stopping the process stops it before the next candidate; nothing was dispatched, so
nothing needs recalling.

## Data

Fictional numbers only, from the `+1 555-0100…555-0199` range reserved for fiction. Numbers
are validated as E.164 before processing and masked in every output path — report,
structured result and error message alike. A regression asserts that no full number appears
anywhere in either output.

## Limits

Not for medical, legal, financial or emergency workflows.

## Checklist

- [x] Real-world side effects are clearly described — there are none in this edition.
- [x] Recurring workflows include cancellation behavior — there is no recurrence; the single pass is described above.
- [x] Runnable code has a dry-run, fake-server, or no-call path by default — no-call is the only path.


Ray-56:
The current head `da0d107` is clean at P0/P1 and its 24 offline tests pass, but it now conflicts with the latest `main` after another apps-list contribution merged. Please merge or rebase `origin/main` and resolve the README/apps index conflicts; no code changes are otherwise requested in this review.

PR #75, opened 2026-08-03
Add fs-sponsor-outreach: batch sponsor-outreach caller for student racing teams

EmeraldKingg:
Adds a Python CLI app under apps/python/fs-sponsor-outreach/ that reads a CSV of prospective sponsor contacts and places one outbound call per contact via the CALL-E CLI, returning structured interest-level and follow-up results.

Status: script and README complete; validated with `scripts/validate_repository.py`. Live-call testing pending a consenting test contact in a CALL-E-supported region — will update before marking ready for review.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ callback-writeback.json
 skills-lock.json
 docs/codex-implementation-plan.md
 docs/superpowers/
+apps/python/fs-sponsor-outreach/results.csv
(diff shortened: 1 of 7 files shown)


Ray-56:
Current blockers at `4c1ff69`:

- The default invocation places live calls; no explicit live confirmation/allowlist is required, phone values are not E.164-validated, and full numbers are printed and persisted.
- There is no durable idempotency or accepted-run checkpoint, so a restart or ambiguous provider response can redial contacts.
- CSV/provider fields are inserted into `innerHTML` without escaping, creating stored XSS in the results dashboard.

Ray-56:

Remaining Must Fix only:

- Default execution makes live calls without explicit intent or E.164/authorized-recipient checks.
- Full real phone output is printed/persisted.
- CSV/provider fields flow to innerHTML without escaping, creating stored XSS.

PR #78, opened 2026-08-03
feat(apps): add standardized survey runner

lukisch:
## Summary

A survey that dials is easy. A survey whose numbers survive scrutiny is not, and almost
every way of getting it wrong flatters the result. Three of those ways are structural in
this app rather than configurable.

**The denominator.** Completion is measured against everyone drawn, never against the
people who picked up. The shipped study prints both, and labels which one is the result:

```text
Drawn:          12
Withdrew:       1  (removed from the data and every denominator)
Included drawn: 11
Completed:      4

Completion yield: 36.4%  (4 of 11 included drawn)
  For contrast, not the result: 57.1% against the 7 who picked up.
```

Same four interviews, two very different numbers. The distance between them *is* the
non-response, and a reader should not have to compute it.

**Coding stays checkable.** A category stored without the words it came from is
unfalsifiable, so a coded answer with an empty raw text is refused at construction. An
uncoded raw answer is fine; coded-without-source is not.

**Withdrawal removes rather than flags.** It drops the identifier, the number and the
answers, and the record leaves every later denominator. In the shipped study one respondent
completed the interview and then withdrew: their completion stops counting and
`included_drawn` falls to 11, while `drawn` stays at 12. A regression asserts neither their
id nor their number appears in either output.

Four ethics rules are locked — consent before questions, the right to stop immediately, the
right to withdraw afterwards, and refusal of high-risk subjects. A study file may **add**
rules; the load path raises if it tries to overwrite one. A consent requirement that
configuration can switch off is not a consent requirement.

Full application: <https://github.com/lukisch/researchcall>

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Verification

```
cd apps/python/researchcall-survey && python -m pytest -q
....................................                                     [100%]
36 passed in 0.12s

Repository validation passed.
```

Both were run on this branch. The validator was also run on the unmodified upstream
checkout first, so a green result here is a comparison and not just an assertion.

## Real-world side effects

None. There is no live transport and no `--live` flag in this edition. No account, no API
key, no network request of any kind. Every outcome comes from the study file, and the run
is labelled `FIXTURE RUN -- SIMULATED, NO CALL PLACED` on its first line.

## Cancellation

One attempt per person. There is no scheduler, no retry and no automatic second round; a
person drawn twice is refused outright. Stopping the process stops it, and nothing was
dispatched.

## Data

Fictional numbers only, from the `+1 555-0100…555-0199` range reserved for fiction — the
frame stands in for a public directory listing filtered by postcode. Numbers are validated
as E.164 before processing and masked in every output path. Regressions assert that no full
number reaches either output, and that a withdrawn respondent appears in neither.

## Limits

High-risk subjects are refused at load time, before a sample is even drawn: medical, legal,
financial and emergency topics raise rather than being handled carefully.

## Checklist

- [x] Real-world side effects are clearly described — there are none in this edition.
- [x] Recurring workflows include cancellation behavior — there is no recurrence; one attempt per person, described above.
- [x] Runnable code has a dry-run, fake-server, or no-call path by default — no-call is the only path.


Ray-56:
Current blockers at `69be620`:

- Withdrawal is implemented as an `anonymised` flag and clears only responses; the original ID/phone remain in `Record.drawn`, and `Interview.person_id`/`note` remain in memory, contrary to the promised deletion semantics.
- `BROKE_OFF` responses are accepted and emitted even though the documented rule says partial answers after stopping are deleted.
- The high-risk check examines only the study subject, so a benign subject can still contain medical/legal/financial/emergency questions that are stored and output.

PR #95, opened 2026-08-04
feat(apps): add Genesis CALL-E orchestrator

bodycount82:
## Summary

Adds `apps/python/genesis-call-e-orchestrator`, a runnable Python app with three structured business phone workflows:

- appointment booking → calendar-draft action;
- consent-aware lead qualification → CRM or do-not-contact action;
- service/vendor coordination → human-reviewed comparison action.

## Safety and default behavior

- Default execution is a masked preview and places no call.
- `--simulate` demonstrates the complete flow without credentials or network access.
- Live mode requires `authorized_recipient: true`, `CALLE_LIVE_CALLS_ENABLED=true`, `CALLE_API_KEY`, and `--confirm-authorized-recipient`.
- It makes one call with deterministic idempotency and no recurrence.
- Credentials are environment-only; output masks phone numbers.
- Prompts require AI/organization disclosure, permission to continue, and opt-out handling.
- The app never modifies calendars, CRMs, contracts, quotes, or payments; follow-ups are drafts requiring review.

## Validation

- `python -m pytest -q` — 10 passed
- `python scripts/validate_repository.py` — passed
- `python scripts/check_branch_name.py --branch feat/genesis-call-e-orchestrator` — passed

No real phone call is required to review or test this contribution.


Ray-56:
Current blockers on df32581:

- `CALLE_BASE_URL` is accepted verbatim and receives `CALLE_API_KEY` without HTTPS or trusted-origin validation. The deterministic key is also not bound to provider origin/account, so the same request can be created in another idempotency namespace while an earlier outcome remains ambiguous.
- Live execution uses an indivisible `create_and_wait` call with no durable pre-dispatch reservation, accepted-call-ID checkpoint, or explicit same-attempt reconciliation state. The lead opt-out result is returned as a draft action but is not durably enforced against a changed request to the same recipient.
- Any `status == completed` response with a nonempty structured object is treated as successful without requiring `task_completed` or validating the scenario schema/evidence. Malformed or partial output can therefore create a calendar or CRM follow-up draft.
- Caller-controlled context is interpolated directly into the live task, allowing the reviewed scenario boundaries to be overridden by injected instructions.
- The output claim is inaccurate: only the destination field is masked, while free-form provider results can contain phone numbers or sensitive details and are written with default file permissions.

Please bind and persist the approved provider request, fail closed on provider/result ambiguity, enforce opt-outs, constrain task inputs, validate terminal evidence, and recursively sanitize private output.

Ray-56:

Remaining Must Fix only:

- Unrestricted CALLE_BASE_URL receives CALLE_API_KEY.
- Raw free-form provider output can disclose real phone/transcript data despite masked-output claim; sanitize normal output.

PR #93, opened 2026-08-04
feat(apps): add hardened FieldClose HVAC closeout workflow

dingtao416:
## Summary

FieldClose turns one human-approved commercial HVAC closeout call into a trustworthy, structured result and an explicit human next step. The contribution includes a runnable fake/no-call path, protected live-call controls, durable audit history, and an end-to-end operator workflow under `apps/web/fieldclose/`.

This PR supersedes #72 and incorporates the blocking review feedback received there.

## Type

- [x] New runnable app
- [x] Safety or documentation update

## Review findings addressed

- Persist refusal and do-not-call suppression by recipient phone lookup hash across cases.
- Reconcile accepted CALL-E call creation after ambiguous responses or process interruption, and support pre-creation cancellation.
- Treat failed and canceled provider states as terminal outcomes instead of polling forever.
- Reject insecure production database, CALL-E, authentication, and SMTP transport configuration.
- Add explicit emergency/local-services and medical, legal, and financial escalation boundaries to the live prompt.
- Align deployment and operations documentation with the current public demo and protected staging state.

## Safety model

- Fake/no-call behavior remains the default.
- Every live attempt requires approval bound to the exact recipient and approved brief.
- Stable idempotency prevents duplicate creation for the same approved attempt.
- Uncertain provider results remain uncertain and require human disposition.
- The agent cannot diagnose, quote, negotiate, authorize payment, promise arrival times, or handle emergencies.

## Validation

FieldClose source:

- `pnpm validate`
- 95 unit tests passed
- 62 integration tests passed
- 15 Playwright end-to-end tests passed
- TypeScript, ESLint, Drizzle schema check, and production build passed
- Secret/private identifier scans returned zero findings

Upstream contribution:

- `python scripts/validate_repository.py`
- Nested `pnpm install --frozen-lockfile` and `pnpm typecheck`
- Repository secret/private identifier/workspace-marker scans returned zero findings

## Checklist

- [x] Repository-facing content is written in English.
- [x] Branch name, commit messages, and PR title follow `docs/git-naming-conventions.md`.
- [x] No secrets, tokens, private phone numbers, call recordings, or private transcripts are included.
- [x] Real-world side effects are clearly described.
- [x] Phone numbers are masked in documentation and test fixtures unless they are clearly fictional.
- [x] Recurring workflows include cancellation behavior.
- [x] Runnable code has a dry-run, fake-server, or no-call path by default.
- [x] `python scripts/validate_repository.py` passes.

## Links

[...]

PR #96, opened 2026-08-05
feat(apps): add hardened FieldClose HVAC closeout workflow

dingtao416:
## Summary

FieldClose turns one human-approved commercial HVAC closeout call into a trustworthy structured result and an explicit human next step. This contribution adds the runnable fake/no-call path, protected live-call controls, durable audit history, and end-to-end operator workflow under `apps/web/fieldclose/`.

This PR supersedes #93, rebases the contribution onto the latest upstream `main`, synchronizes the latest FieldClose source, and addresses the blocking review on `664e1ae`.

## Blocking findings addressed

- Fail closed when a non-loopback production PostgreSQL URL does not contain exactly one `sslmode=verify-full`.
- Pass `ssl: "verify-full"` explicitly to postgres.js for remote databases while retaining the loopback development exception.
- Apply the same certificate-verification requirement to the public-demo build gate and document the deployment requirement.
- Add regression coverage for missing TLS, `sslmode=require`, conflicting modes, loopback URLs, environment parsing, public-demo verification, and effective postgres.js client options.
- Align readiness claims with inspectable evidence: only the fake-only public deployment is claimed; protected staging deployment, isolation, production access, and deployed CALL-E/SMTP remain pending verification.
- Synchronize the latest source changes for US-only live-recipient validation and explicit Shanghai calling-window coverage.

## Safety model

- Fake/no-call behavior remains the default.
- Every live attempt requires approval bound to the exact recipient and approved brief.
- Stable idempotency prevents duplicate creation for the same approved attempt.
- Uncertain provider results remain uncertain and require human disposition.
- The agent cannot diagnose, quote, negotiate, authorize payment, promise arrival times, or handle emergencies.

## Validation

Upstream contribution:

- `python scripts/check_branch_name.py --branch feat/fieldclose-hvac-closeout-tls`
- `python scripts/validate_repository.py`
- nested `pnpm install --frozen-lockfile`
- nested `pnpm typecheck`
- 27 focused database/environment unit tests passed

FieldClose source:

- TypeScript, ESLint, 103 unit tests, Drizzle schema validation, 63 PostgreSQL integration tests, and production build passed
- the first publish-time Playwright run had two timing flakes; an unchanged immediate `pnpm test:e2e` retry passed all 15 tests
- no live CALL-E, SMTP, or external database action was performed

## Checklist

- [x] Repository-facing content is English.
- [x] The branch and commit names follow upstream conventions.
- [x] No secrets, private phone numbers, recordings, or private transcripts are included.
- [x] Real-world side effects and credential handling are documented.
- [x] Fake/no-call behavior is the default.
- [x] Upstream repository validation passes.

## Links

- Public fake-only demo: https://fieldclose.dramaforge.icu
- Source repository: https://github.com/dingtao416/FieldClose


Ray-56:
Current blockers at 9a72bf1:\n\n- setup-local-demo.mjs writes secrets with mode 0600 only at creation time. Overwriting an existing 0644 .env.local preserves 0644, and the path is not protected against non-regular or symlink targets, so the generated auth and phone-protection keys can be exposed.\n- Concurrent live execute requests can both finish prepareLiveProviderRequest and invoke CALL-E after the transaction lock is released. The shared idempotency key helps provider deduplication, but the two responses can still race: accepted, failed, or ambiguous persistence updates overwrite one another and leave the attempt/case/tasks inconsistent. Claim provider creation durably before releasing the transaction, and make every outcome update conditional on the claimed state.

dingtao416:
Addressed both blockers in `95a2810`.

- Secret writer now rejects symlink/non-regular targets, hardens existing files to `0600`, verifies the result, and has regression tests.
- Live creation now uses a durable claim with a 60-second recovery lease, reuses the same idempotency key, makes terminal writes conditional, and the regression test asserts exactly one provider call.

Validation passed: FieldClose `pnpm validate` (107 unit, 64 integration, 15 Playwright, type/lint/schema/build), focused upstream tests, branch-name check, and `scripts/validate_repository.py`. Private phone/email scans found zero matches; local redacted live-test evidence remains gitignored.

Ray-56:
One ambiguous-outcome blocker remains on `95a2810`: `classifyCreationError` maps every `CalleAPIError` with `status < 500` to `failed_before_acceptance`, including HTTP 408 and 409. Neither response proves that CALL-E did not create the call, so this can permit an unsafe follow-up attempt. Please classify 408/409 as `ambiguous_requires_reconciliation` and add focused adapter tests.

dingtao416:
Addressed the remaining ambiguous-outcome blocker in `e742691` (PR head `3f8ea66` after merging current `upstream/main`).

The adapter no longer treats every `<500` response as proof of pre-acceptance failure. `failed_before_acceptance` now uses an explicit allowlist (400, 401, 403, 404, 405, 413, 415, 422); every other API error defaults to `ambiguous_requires_reconciliation`, including 408, 409, 425, 429, all 5xx responses, transport exceptions, and missing success bodies.

Focused adapter coverage now exercises 18 cases, including the requested 408 and 409 paths. Documentation also clarifies that lease recovery can repeat the HTTP request only with the same durable idempotency key, so the guarantee is one logical provider creation.

Validation: FieldClose `pnpm validate` passed (120 unit, 1 skipped, 64 PostgreSQL integration, 15 Playwright, type/lint/schema/build); merged-tree upstream typecheck and all 18 focused adapter tests passed; branch-name and repository validation passed. Added-content scans found no private phone, email, or credential values.

Ray-56 (approved):
The frozen ambiguous-creation blocker is resolved: HTTP 408/409 and all non-definitive provider or transport errors now require reconciliation. Repository validation, typecheck, and 121 credential-free unit tests pass.

PR #100, opened 2026-08-06
feat(apps): add voice-preflight, hear a task script before the callee does

zkasuran:
## Summary

CALL-E speaks the `task` text to a real person and there is no way to hear it
first. I checked before building: `CreateCallRequest` carries `task`, `recipients`,
`result_schema`, `recipient_result_schema`, `metadata` and `webhook_url` with
`additionalProperties: false`, so there is no field for a voice or a provider and
an unknown one is rejected rather than ignored.

So this renders the same text through a text-to-speech API the operator already
pays for, then gates the call on what the audio and the text actually show. It
places no calls and needs no CALL-E account.

Three failures it catches, all invisible in review:

**A line disappears.** `phone-approval-gate` reports approval only when a person
reads back a six digit code, so the sentence asking for it is the whole mechanism.
Somebody shortens the wording, the sentence goes, then the gate cannot be
approved by anyone. The suite still passes, because no test asserts prose. That is
case 2 of the demo.

**The voice cannot speak the recipient's language.** A recipient carrying
`locale: hi-IN` with an English-only voice burns the one attempt you had.

**The script is longer spoken than anyone thought.** Characters do not tell you
seconds. A budget only means something measured against real audio.

## Any provider, described rather than coded against

One HTTP client, no vendor code. A descriptor supplies the endpoint, method, auth
header and prefix, static headers, body template, where the audio sits, format,
character limit and the languages the voice speaks. `audio.kind` covers the three
shapes providers use: `body` for raw bytes, `base64Field` for base64 inside JSON,
`urlField` for a link that gets followed.

`examples/` carries descriptors for ElevenLabs and Fish Audio plus the local fake.
The ElevenLabs one is built from the published `POST /v1/text-to-speech/{voice_id}`
reference. The Fish Audio one is built from a working implementation, including its
static `model` header, which is what the `headers` field exists for. `maxChars` on
both is this app's own conservative ceiling rather than a documented provider limit,
and `docs/providers.md` says so.

## No finding predicts pronunciation

This is the design decision worth reviewing first. A tool that claims a string
will be misread is making a guess nobody measured. The audio is the ground truth
and a person listening to it is the real check.

| Finding | Blocks | Basis |
| --- | --- | --- |
| `locked_line_missing` | yes | substring absent from the task |
| `voice_language_mismatch` | yes | language subtag not in the descriptor's list |
| `text_over_provider_limit` | yes | character count against the descriptor |
| `spoken_too_long` | yes | duration measured from the rendered container |
| `digit_run_unseparated` | no | a run of four or more digits is present |

The last row reports the run and says to listen. Duration is parsed from the WAV
header or read from ffprobe. With neither it is reported unknown and the length
check is skipped rather than estimated from character count.

## Credential handling

The descriptor names the environment variable holding a credential and never holds
one. A descriptor is the file people commit and paste to each other, so the loader
refuses key material in every field it reads rather than in a static header alone:
`authPrefix`, the endpoint path and query, a header name as well as a header value,
the body template, the JSON path and the language list. `authEnv` has to be shaped
like a variable name. An unknown field is refused rather than ignored, so there is
no spare field to paste a key into. The same check runs again in `render`, because a
descriptor built in code never passes the loader.

There is no default trusted host, because this talks to no fixed vendor. The
operator names it once with `--allow-host` or `VOICE_ALLOWED_HOSTS` or nothing is
sent. Plain http reaches loopback only, which is what the fake and the tests use.
https alone says the transport is encrypted and nothing about who is on the other
end.

Two things a provider decides at runtime get their own rule. Redirects are followed
by hand, up to three hops. The request carrying the credential may only be
redirected inside the endpoint origin the operator wrote, so a provider can move one
of its own paths while a `Location` pointing anywhere else is refused. That is
narrower than the allowlist on purpose: a host named so `urlField` audio can be
fetched is not a host that may receive the key. An audio URL a provider returns
carries no credential, so it gets the allowlist plus https, with no loopback,
link-local or private address.

Order matters and `docs/providers.md` states it: the descriptor contract, then the
character limit, then the host, then the cache, then the credential. A script too
long to send never reads a secret. An endpoint nobody allowed never receives one.
Rendered audio is cached under a digest of provider, voice and text, written `0600`.

## Relationship to #67

`skills/calle-script-advisor` in #67 lints the `task` and the `result_schema` as
text. This renders audio and checks what only exists once text becomes speech.
Different axes. I have kept this one clear of that one: no score, no schema
linting, no drafting advice, no safety-phrase rules.

## Type

- [x] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Verification

Run at `9e52629`:

- `npm run check` clean, `tsc --noEmit` strict
- `npm test` 84 passed, 0 failed, 0 skipped
- `npm run demo` four cases, three refusals, real WAV over loopback

No lockfile and no `node_modules` are committed. 30 files, all new except the two
index tables.

Two rounds of review findings from @Ray-56 landed on this branch after the first
push. `f0ea4e3` checks every redirect hop plus every provider-returned URL.
`9e52629` keeps the credential on the endpoint origin so the fetch allowlist cannot
authorize a key drop. It also refuses inline key material in every descriptor field.
The count above grew from 52 with the tests those findings needed. Each thread comment
carries the reproduction and the failure it prints against the unfixed code.

## Checklist

## Real-world side effects

`preview` contacts nothing and reads no credential. `render` calls the provider the
descriptor names, which bills characters against the operator's own account. No
phone call is placed by any path. The app holds no phone numbers. The only
files written are cached audio under the chosen cache directory, mode `0600`.

## Not tested live

No render has run against ElevenLabs or Fish Audio from this branch, because I hold
no account with either. The whole suite and the demo run against the local fake,
which returns a real container so the measurements are real. The two vendor
descriptors are transcriptions of a published reference and of a working
implementation. `docs/providers.md` tells the reader to check either against the
provider's current reference before relying on it.

## AI assistance

AI assistance (Claude, Anthropic) was used in developing this change. The design,
review and verification were done by the author. Verified locally before
submitting: `tsc --noEmit`, `npm test` at 84 passed with 0 failed and 0 skipped,
`npm run demo` end to end against the local fake provider plus
`python3 scripts/validate_repository.py`.


Ray-56:
Current blockers:\n\n- The credential-bearing provider fetch follows redirects automatically. Node strips Authorization on a cross-origin redirect but preserves descriptor-selected headers such as xi-api-key; a local reproduction confirmed the custom API key reached the redirect target. Disable automatic redirects or validate every hop and rebuild headers only for an approved host.\n- audio.kind=urlField follows the provider-returned URL with no scheme or host validation. A provider response can therefore make the app issue arbitrary GET requests, including to loopback or link-local services. Apply the same explicit host and protocol policy to the audio URL before fetching it.\n\nThe 52 tests and demo pass, but they do not cover these redirect and returned-URL trust boundaries.

zkasuran:
Both are real. Fixed in f0ea4e3 on this branch, with the tests that were missing.

## 1. The credentialed request followed redirects

**What the reproduction showed.** Two local fake providers on separate loopback
ports. Hop one answered `302` with an absolute `Location` at hop two. Hop two
recorded `GET /speak` carrying `xi-api-key: fake-key-value`. Nothing was stripped,
exactly as you said: the descriptor names the header its key travels in, so
`Authorization` is the only one that would have been dropped.

**What the fix does.** I took the second option you named rather than refusing
redirects outright. The reason is in a comment on `sendChecked`: a blanket refusal
breaks a provider that normalises a host or moves a path, which an operator cannot
fix from a descriptor. So `redirect: "manual"` is set on every request now. Hops
are followed by hand, up to three. Each `Location` goes through the same
`assertTrustedUrl` as the endpoint before anything is sent to it. The headers are
rebuilt per hop from a function rather than carried across one, so a credential
only ever attaches to a URL that has just passed. A hop on the origin the request
started from is the same destination so it passes. Anything else has to be https on
a host the operator named, which also means a loopback fake cannot redirect the key
to a second local port.

**Which test pins it.** `refuses a cross-origin hop and the target sees no request
at all`. It asserts hop two saw nothing before it asserts the refusal, so against
the old code the failure prints the request with the key in it rather than a bare
missing rejection:

```
the redirect target was never contacted
+ actual - expected
+ [ { auth: 'fake-key-value', body: '', method: 'GET', path: '/speak' } ]
- []
```

Four more cover the rest of that surface: `refuses a hop to a remote host the
operator never named`, `follows a hop to an allowed host and rebuilds the
credential there`, `stops at the hop cap rather than following a loop` (the old
client made 21 requests on a same-origin loop where this one stops at 4) plus
`follows a same-origin 307 and re-attaches the credential with the body`, which is
the path that had to keep working. That last one passes before and after by design.

## 2. `audio.kind=urlField` followed the returned URL unchecked

**What the reproduction showed.** The fake answered `/speak-url` with a URL on a
second loopback server standing in for an internal service. The service was asked
for `/latest/meta-data/iam`. The 21 bytes it replied with were written into the
audio cache as a `.wav`:

```
the internal service was asked for: [ '/latest/meta-data/iam' ]
render returned: { provider: 'local-fake', voiceId: 'v', bytes: 21, ... }
cache holds local-fake-ed53fbb68a8f.wav, 21 bytes: internal-service-body
```

**What the fix does.** One function, `assertTrustedUrl` in `src/hosts.ts`, is the
policy for all three call sites now: the endpoint, every redirect hop and any audio
URL a provider hands back. https, a host on the operator's allowlist, no
credentials in the URL itself and no literal address in `127.0.0.0/8`, `::1`,
`169.254.0.0/16` including `169.254.169.254`, `10.0.0.0/8`, `172.16.0.0/12`,
`192.168.0.0/16` or `0.0.0.0`. Mapped and abbreviated forms are normalised first,
since `new URL()` rewrites `[::ffff:127.0.0.1]` as `[::ffff:7f00:1]`. A URL that
came out of a response gets the strictest form: loopback is refused there even
though the operator may point the endpoint at a local fake, because a response must
not be able to steer this app at a service on the machine it runs on. The link is
fetched with no credential attached. The same script against the fix:

```
refused: ConfigError: The audio URL local-fake returned: 127.0.0.1 is on this
machine, which a URL from a provider may not point at. Name the host with
--allow-host or VOICE_ALLOWED_HOSTS. ... LOCAL_FAKE_KEY was not sent to that host
and no audio was written.
the internal service was asked for: []
```

**Which tests pin it.** `refuses a URL on this machine and the internal service
sees nothing`, `refuses a URL that is not https`, `refuses a literal address in
loopback, link-local or private space` (`169.254.169.254`, `10.0.0.7`,
`172.16.4.9`, `192.168.1.10`, `0.0.0.0`, `127.0.0.1`, `[::1]`), `refuses a URL
carrying credentials in the URL itself`, `refuses a URL on a host the operator
never named` plus `checks a redirect on the audio link with the same policy`, which
is there because the hop rule has to hold on this path too. `fetches a URL on an
allowed https host, without the credential` is the working path. It asserts the
link fetch carried no auth header.

## The counts, since that was the real gap

`test/trust-boundaries.test.ts` is 12 cases. Every case asserts what the far side
saw before it asserts the refusal, which is why the failures above print the
request that should never have been made.

With `src/gateway.ts` plus `src/hosts.ts` reverted to 6b8c6a1 and everything else
left as shipped, that file runs 12, 2 pass, 10 fail. The two that pass either way
are the working paths named above. On this branch all 12 pass.

The suite is 64 tests, up from 52. `npm run check` clean, `npm test` 64 passing 0
failing, `npm run demo` unchanged, `python3 scripts/validate_repository.py` passes.
Node 22 locally.

Docs moved with the behaviour, because a descriptor author sees it: `README.md` plus
`docs/providers.md` now state the hop rule and the fact that a `urlField` provider
whose audio sits on a separate CDN host needs that host allowed too.

One note on `SECURITY.md`. It asks for a private advisory for a vulnerability plus
no working exploit code or live credentials. This is unmerged code under review so
the thread seemed the right place. The reproduction runs entirely against the
local fake with a fake key value, no third-party host and no real credential. Happy
to move any of it if you would rather.

Ray-56:
Current credential-safety blockers on `f0ea4e3`:

- One `allowedHosts` set authorizes both credential-free audio URLs and credential-bearing redirects. A CDN host added for `urlField` audio can therefore receive the TTS key when the provider endpoint redirects to it; the current test explicitly expects the credential to be reattached to any allowed redirect host. Keep credentials on the original approved origin or use a separate credential allowlist.
- The descriptor contract still permits inline secrets outside static headers. I verified that long key material in `authPrefix` and in an endpoint query parameter is accepted and sent. Please structurally reject inline credentials across all descriptor fields and add regression tests.

zkasuran:
Both confirmed, both fixed in `9e52629` on this branch. Each was reproduced first,
against the local fake with a fake key value.

## 1. One allowlist was answering two questions

**What the reproduction showed.** A `urlField` descriptor whose audio sits on
`cdn.acme.example`, with the operator naming that host for exactly that reason.
The endpoint answered `307` pointing at the CDN. Hop one went to the CDN carrying
the key:

```
POST https://api.acme.example/v1/tts
  xi-api-key: fake-tts-key-value
POST https://cdn.acme.example/v1/tts
  xi-api-key: fake-tts-key-value
GET https://cdn.acme.example/render.wav
  xi-api-key: (absent)
LEAK: cdn.acme.example was named for credential-free audio and received ACME_TTS_KEY on 1 request(s).
```

**Our own test asserted the wrong thing, which is worth saying plainly.**
`follows a hop to an allowed host and rebuilds the credential there` shipped in
that file last round and I listed it in my previous comment as one of the cases
covering the surface. It encoded the finding as the expected behaviour: the
credential rebuilt for any allowed host. It is gone. In its place is `refuses a
hop to a host that was only allowed for audio`, the same scenario with the
correct expectation. The file header now says so too, so nobody reads the history
and wonders which version was intended.

**What the fix does.** I took the first option you named, credentials on the
original approved origin, rather than a second allowlist. A separate credential
allowlist is one more thing to misconfigure and it can be widened. The endpoint
origin is a value the operator has already written and cannot widen by accident,
and a provider that genuinely serves from another host is one endpoint edit away,
which is a decision the operator makes rather than one a `Location` header makes
for them.

So `assertTrustedUrl` is two functions in `src/hosts.ts` now. `assertFetchable`
answers "may this app fetch it": https, a host on the allowlist, no literal
loopback, link-local or private address. The endpoint, an audio URL and every hop
an audio URL takes go through it. `assertCredentialTarget` answers "may the
credential be attached". The answer is the endpoint origin only: same scheme,
same host, same port. A hop off that origin is refused rather than retried without
the key, because continuing would hand the script body to a host the provider
chose. A silent unauthenticated retry also reads as a provider outage rather than
as the policy decision it is.

The same script against the fix:

```
render threw: The endpoint redirected to hop 1: https://cdn.acme.example is not the
origin the operator approved. ACME_TTS_KEY travels to https://api.acme.example only,
which is the endpoint you wrote, so a host you allowed for audio never receives it.
Point the descriptor endpoint at that origin if the credential belongs there.
ACME_TTS_KEY was not sent to that host.
POST https://api.acme.example/v1/tts
  xi-api-key: fake-tts-key-value
no credential reached the audio host
```

**Which tests pin it.** `refuses a hop to a host that was only allowed for audio`
asserts the CDN was never contacted and that every request carrying the key went
to the endpoint, before it asserts the refusal. Against `src/` at `f0ea4e3` it
fails with the request that should never have been made:

```
the CDN host was never contacted
+ actual - expected
  [
    'https://api.acme.example/v1/tts',
+   'https://cdn.acme.example/v1/tts'
  ]
```

`refuses a hop to the endpoint host on another port, because a port is part of an
origin` is new alongside it. Both working paths stay green. `follows a same-origin
307 and re-attaches the credential with the body` is untouched, so a provider
moving one of its own paths still works. `follows an audio link across two hosts
the operator named, with no credential` is new, because tightening the credential
rule must not become a refusal of every hop on the path that carries no key.

## 2. Inline secrets outside static headers

**What the reproduction showed.** Eleven descriptors through `loadDescriptor` then
`render`: ten probes, one per field that can carry a string, plus a legitimate one
as the control. Nine of the ten probes were accepted by the loader. Five of those
put a fake 33 character value on the wire, including both fields you named:

```
authPrefix: ACCEPTED and sent. LEAK: the inline value travelled.
  headers:   {"authorization":"Bearer FAKEKEY0123456789abcdefghijklmnop fake-tts-key-value","content-type":"application/json"}
endpoint query parameter: ACCEPTED and sent. LEAK: the inline value travelled.
  requested: https://api.acme.example/v1/tts?api_key=FAKEKEY0123456789abcdefghijklmnop
authHeader: ACCEPTED and sent. LEAK: the inline value travelled.
headers field name: ACCEPTED and sent. LEAK: the inline value travelled.
bodyTemplate: ACCEPTED and sent. LEAK: the inline value travelled.
```

The remaining five matter as well. Credentials in front of the host loaded and
were then refused by `render` rather than at load. Three more loaded with the
value sitting in `audio.path`, in `languages` and in an unknown field. Those three
strings never reach a URL or a header, but they are still key material in a file
people commit and paste. The one probe the old guard did catch was the static
header value.

**What the fix does.** `assertNoInlineSecret` in `src/config.ts` is now a
structural check over the whole field set rather than two spot checks. It runs
before every other validation. The field set is closed, so an unknown field is
refused rather than ignored. That was deliberate: without it, "every field" is a
list that a future field or a typo walks straight out of.

Where a field has a grammar the grammar is the guard. `authPrefix` is one auth
scheme name, at most 16 characters plus an optional trailing space, which is what
`Bearer ` and `AWS4-HMAC-SHA256 ` are. `authHeader` and every static header name
have to be RFC 7230 field names. The endpoint is parsed, userinfo is refused
outright and the path, the query and the fragment are scanned separately so the
refusal names the part.

What is left is free text, so it is scanned for the shape of a key: 20 characters
unbroken or 24 counting `_` and `-`. `bodyTemplate`, static header names and
values, `audio.path` and `languages` all go through it. Two fields do not, on
purpose. The host is governed by the allowlist and a long hostname of ordinary
words is legitimate. `name` and `authEnv` run past the thresholds by design and
neither string reaches a URL or a header. The second threshold is 24 rather than
20 because `examples/provider.elevenlabs.json` carries `eleven_multilingual_v2`,
22 characters of exactly the alphabet a key is written in. It is a model name.
A false positive costs one rename. A false negative commits somebody's key.

The check also runs at the top of `render`, because a descriptor built in code
never passes the loader.

**Which tests pin it.** `test/descriptor-contract.test.ts`, 18 cases. One per
field, starting with the two you probed, plus the three grammar cases and one for
the unknown field. Each asserts the refusal names the field and does not echo the
value back.

Four are the other half of the contract, since a guard that refuses everything
would pass every case above: `accepts every descriptor shipped in examples` loads
all three, `accepts a long model name, which is not a credential` pins the
ElevenLabs body template, `refuses key material in a static header value` is the
one the old guard already caught and `renders a legitimate one, so the gate in
render is not in the way` renders a valid descriptor through a stub and asserts
the header arrived as `Bearer fake-key-value`.

## Counts

Against `src/config.ts`, `src/gateway.ts` and `src/hosts.ts` at `f0ea4e3`, with
both test files as they stand now, those two files run 32 cases: 15 pass, 17 fail.
In the contract file 14 of the 18 fail, which is every case except the static
header value and the three that prove a legitimate descriptor still loads. In
trust-boundaries three fail: the CDN hop, the port case and `refuses a hop to a
remote host the operator never named`, which refused before this change too but
with the allowlist message rather than the origin one.

The suite is 84 tests, up from 64 last round. `npm run check` clean, `npm test` 84
passing 0 failing, `npm run demo` unchanged, `python3 scripts/validate_repository.py`
passes. Node 22 locally.

Docs moved with the behaviour, since a descriptor author sees both rules.
`docs/providers.md` gains the credential contract field by field and the order of
operations now starts with it. `README.md` states that naming a host for audio
does not authorize the key there.

On `SECURITY.md` again: both reproductions ran locally against a stubbed fetch and
the local fake with fake key values, no third-party host and no real credential.
The script is not in the diff.

PR #102, opened 2026-08-06
feat(apps): add linecanary phone-line monitoring with companion skill

mark24680617:
## What this adds

**LineCanary** — synthetic uptime monitoring for business phone lines and deployed AI voice agents, plus the `linecanary-monitor` agent skill.

Websites have had uptime monitoring for decades; phone lines never got theirs. LineCanary places scheduled CALL-E calls to lines the operator **owns**, walks the caller journey (menus, key presses, questions), asserts on the schema-validated results, diffs them against the line's own history, and alerts on regressions — with a dashboard that shows the full transcript of what the canary heard on every call.

- `apps/typescript/linecanary/` — runnable app: config-as-code checks, ownership verification (greeting-code call or recorded attestation), call-window guard, baseline regression diffing, operator dashboard + per-client public status pages with uptime, per-check call logs, Slack alerts with transcript excerpts, GitHub Action for CI gating, AI incident notes and one-call line discovery (optional, Anthropic key)
- `skills/linecanary-monitor/` — companion skill: when/when-not to use, config reference, worked examples, safety boundaries

## Side effects & safety

- Live runs place **real phone calls** — at most one per check per invocation; the host scheduler owns recurrence
- **Dry-run is the default**; `--live` is explicit
- Calls only go to **verified lines** (greeting-code proof of control, or recorded written authorization), only **inside the configured call window**, and every call opens with an AI disclosure; unexpected live persons are apologized to and released
- Transcripts are treated as untrusted data end to end (assertions compare, never obey; HTML-escaped; kept inside data tags in AI prompts) — see `docs/threat-model.md` and `docs/compliance.md`

## How to test (no credentials, no calls)

```bash
cd apps/typescript/linecanary
npm install
npm test        # 102 tests drive the real @call-e/calle SDK against a local fake server
npm run demo    # healthy day → silent IVR breakage → regression alert, zero network
```

With a CALL-E key: `init → verify → run --live → serve` (full quickstart in the README).

## Live demo

A production instance monitors a real Twilio IVR ("Sample Dental") on a real phone number, on a live schedule:
- Public status page: https://linecanary-production.up.railway.app/status/dental-ivr
- Operator dashboard: https://linecanary-production.up.railway.app/ (credentials in the Devpost submission's testing instructions)

Prior art inside this repo is scoped in the README (the n8n `calle-ivr-quality` example is a one-shot probe; LineCanary is continuous monitoring with baselines, alerting and a product surface).


Ray-56:
Current blockers:\n\n- verify and discover create real calls without requiring --live. discover also bypasses the stored ownership-verification gate entirely, so any configured E.164 target can be called despite the claimed ownership-only and dry-run-default posture.\n- A first failing check is reported as healthy: an exact in-memory run produced outcomeStatus=fail with reportOk=true and no regressions, so the CLI exits 0. Baseline parsing also converts any corrupt history file to an empty history, recreating this fail-open state.\n- Create ambiguity and timeouts are not durably reconciled before later scheduled runs can use a new minute-based key and call again. The key also omits line and phone, so separate configs reusing a check ID can collide within a provider account.\n- Operator pages and /api/state expose full transcripts whenever the server is bound externally without LINECANARY_DASHBOARD_PASSWORD; the code does not refuse that configuration.\n\nPlease make all call-producing commands explicitly live, fail closed on first-run failures and state corruption, add durable ambiguous-call recovery, and require protection for non-loopback operator surfaces.

mark24680617:
Thank you for the review — all four blockers were real, and they're fixed in 7a92558. Point by point:

**1. `verify`/`discover` placed calls without `--live`; `discover` bypassed the ownership gate.**
Both commands now require `--live` to place a call; without it they print a dry-run preview of the exact call they would place. `discover` additionally requires a stored ownership verification matching the line's phone — the same gate `run --live` enforces — and exits 2 with guidance otherwise. (For `attestation` lines, `verify` records the statement without a call, so it stays available without `--live`; the greeting-code path is gated.) Docs and the skill's examples updated to match.

**2. First failing check reported healthy; corrupt history silently reset.**
`runChecks` now fails closed: any non-pass outcome fails the report (exit 1) regardless of whether a baseline exists to diff against — "first run is baseline" now only applies to the *diff*, never to `ok`. Corrupt state files are no longer swallowed into an empty history: they're renamed to `<file>.corrupt-<ts>` with a stderr warning, so the evidence survives and the reset is loud. New tests cover both (`a failing first run fails the report even with no baseline`, `a corrupt history file is quarantined loudly`).

**3. Ambiguous creates/timeouts not durably reconciled; key omitted line/phone.**
Added a persisted `PendingAttempt` per check (`<check>.pending.json`): recorded before every create, upgraded with the call id after a successful create, and cleared only once an outcome is appended (or on a definitive non-ambiguous refusal). The next run reconciles instead of dialing: a pending call id is polled directly; a pending key from an ambiguous create is *reused*, so the provider's idempotency replay returns the existing call rather than ringing the line again. Keys are now `linecanary:<check>:<phone>:<minute>`, so two configs reusing a check id under one account can't collide. Tests drive both paths end-to-end through the real SDK against the wire-faithful fake and assert the line rang exactly once.

**4. Operator surfaces exposed unauthenticated on external binds.**
`startDashboard` now refuses to bind to any non-loopback host unless a password is set, with an error naming the fix (`LINECANARY_DASHBOARD_PASSWORD` or bind loopback). The public `/status` pages remain the only unauthenticated surface.

Suite is 118/118 in the packaged copy; `scripts/validate_repository.py` passes. Happy to adjust further if you'd prefer different semantics anywhere (e.g. a non-zero exit for `verify` dry-runs).

Ray-56:
Current blockers at `e543da0`:

- GitHub reports this branch conflicts with `main`, so it cannot be merged as submitted.
- Dry-run/config paths still print full or invalid raw phone input (`src/cli.ts:168-173,357-361`; `src/config.ts:168-170`) instead of masking it as required by the repository safety rules.

Ray-56:
Maintainer follow-up: I resolved the current `main` conflict and masked the remaining raw-phone output/error paths. The repository validator, TypeScript check, and all 144 tests pass at `be767a4`; there are no remaining Must Fix items under the Community/Hackathon review policy.

PR #116, opened 2026-08-07
feat(apps): add calle-mcp-result-extractor

ar-yogeshkhadilkar:
CALL-E's MCP tool surface (plan_call/run_call/get_call_run) has no result_schema parameter -- that's a REST-API-only feature of the CALL-E Developer API. Over MCP, a completed run's `extracted` field is always CALL-E's own generic call metadata, never output shaped to a caller's schema, even though the transcript is always available.

This adds a small extraction layer that gets a schema-validated structured result from the transcript instead, with a pluggable reasoning provider (a Bedrock adapter included, plus a deterministic fake provider so the demo and tests need no credentials at all).

It also documents and regression-tests a real Bedrock Converse API pitfall: retrying a failed structured-output attempt by appending a plain-text follow-up after a tool_use block breaks the API's tool_use/tool_result pairing requirement and gets the whole retry rejected. The fix is to send the validation feedback back as a toolResult block instead.

## Test plan
- [x] `npm run demo` -- full pipeline with zero credentials and zero network calls
- [x] `npm test` -- 6/6 passing, including two regression tests against a stubbed Bedrock client covering both the pairing fix and the no-tool_use-returned edge case
- [x] `npm run check` -- `tsc --noEmit` clean
- [x] `python3 scripts/validate_repository.py` -- passes

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,7 @@ Runnable demo apps live under [`apps/`](apps/). They are not a CALL-E SDK and do
 | [`apps/python/oauth-login-client`](apps/python/oauth-login-client/) | Python | CALL-E OAuth login client for MCP Streamable HTTP. |
 | [`apps/typescript/oauth-login-client`](apps/typescript/oauth-login-client/) | TypeScript | CALL-E OAuth login client for MCP Streamable HTTP. |
 | [`apps/typescript/vibehub-founder-relay`](apps/typescript/vibehub-founder-relay/) | TypeScript | Consent-first founder-match readiness call with masked preview, stable idempotency, and structured CALL-E results. |
+| [`apps/typescript/calle-mcp-result-extractor`](apps/typescript/calle-mcp-result-extractor/) | TypeScript | Schema-validated structured results from CALL-E calls placed over MCP, which has no `result_schema` parameter of its own — extracts from the transcript instead, with a documented fix for a real Bedrock Converse API tool_use/tool_result retry pitfall. |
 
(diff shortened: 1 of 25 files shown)


Ray-56:
Current blockers at `a5b4299`:

- `plan` prints the full response including the call-authorizing `confirm_token`, and `call` accepts that token as a command-line argument (`src/cli.ts:88-122`), exposing it in terminal logs, shell history, and process listings.
- `plan` forwards `--to` without enforcing E.164 locally, while `call` and `status` print raw run objects that may contain transcript-derived phone numbers (`:88-130`). Keep the token in private state/stdin and mask all user-facing output.

Ray-56:
Current blockers on `e70b801`:

- Starting a new plan does not clear the prior pending plan when the new request is not ready or planning fails. `call --live` can then silently execute the older destination/goal, so the latest review can fail while a stale call remains authorized.
- Invalid destination errors interpolate the unmasked input and the top-level handler prints the raw error, contradicting the promise that every command masks phone output.

Please clear pending authorization before every plan attempt/on every non-ready result, require a faithful masked preview of the stored plan before live execution, and mask validation errors.

Ray-56:
Thanks for the update. The frozen stale-authorization issue is only partially fixed: `planAndSave` clears pending state before E.164 validation and the network call, but `cmdPlan` still calls `usage()` and `resolveCalleMcpConfig()` first. A missing argument or configuration error therefore leaves the previous ready plan/token on disk, and `call --live` can execute it. Please clear pending state at the very start of every plan attempt, before argument and configuration resolution, and cover those paths with tests.

PR #117, opened 2026-08-07
feat(apps): add ringer consumer call app

DevJustinTech:
## Summary

Ringer packages the phone calls people dread — negotiating a bill, cancelling a
"call-to-cancel" subscription, chasing a stuck refund, booking an appointment,
and calling around for the best quote — as consent-first CALL-E workflows that
return a **structured, honest** result. Live demo: https://ringer-steel.vercel.app · source:
https://github.com/DevJustinTech/ringer · 4-min video: https://www.loom.com/share/0950357138524b269428b61f6695cc40

Two ideas define it:

- **It names its denominator.** Quote Shootout calls several businesses and
  reports how many were *called*, *reached*, and *answered*. A business that hit
  voicemail or declined is never counted or folded into the comparison
  ("2 quoted of 3 called · 1 not reached — not counted"); cheapest/savings are
  computed over answered businesses only.
- **It won't assert what it can't back.** Every completed call is tagged
  *evidence-backed* or *needs review* from CALL-E's completion confidence +
  quoted evidence; the copy/share output flags anything unverified.

It's a real product on the CALL-E Developer API: `client.calls.create()` with
per-call and per-recipient strict `result_schema`s, a create-then-poll live
timeline over `/v1/calls/{id}` + `/events`, and signed-webhook verification
(`client.webhooks.unwrap()`). It runs entirely in **Demo Mode by default** — a
client-side simulator that reproduces CALL-E's object model — so anyone can
experience the whole product with no key and zero credits. Also: durable
scheduled/recurring calls (operator-enabled), calls conducted in the recipient's
language across CALL-E's supported regions (9 languages), and a fixed AI
disclosure at the start of every call.

Full docs: `apps/typescript/ringer/README.md`.

## Type

- [x] New runnable app

## Verification

App test suites, all green on this branch:

```
pnpm build          # tsc -b + vite build          → built
pnpm typecheck:api  # tsc -p api                    → clean
pnpm verify         # demo/logic suite              → All 98 checks passed
pnpm verify:api     # serverless safety invariants  → All 17 API-safety checks passed
pnpm verify:render  # SSR smoke of display comps     → All 28 components rendered
pnpm lint           # oxlint                         → clean
```

`verify:api` specifically pins the safety invariants: content-bound idempotency
(edited phone/task/schema → new key; identical → same), base-URL allowlisting
(the server key is never sent to a client-chosen host), and shared-key
authorization (BYOK is ungated; the operator's server key requires
`CALLE_APP_SECRET` and fails closed). `verify` includes regressions for
denominator honesty (a voicemail in a shootout is not counted) and the
verification gate (a low-confidence call is flagged unverified).

Awesome-list validator (run against this branch at submission):

```
python3 scripts/check_branch_name.py --branch feat/ringer-consumer-call-app
```

## Real-world side effects

- **Demo Mode is the default and places no calls** — the required no-call path.
  Live mode places real outbound calls only after an explicit per-call consent
  checkbox ("I'm authorized to place this call and the number is correct").
- Every call opens with a **fixed AI disclosure** (not caller-editable): the
  agent states it is an automated assistant calling on the user's behalf.
- **Scheduled / recurring** follow-ups and "Rate Watch" renegotiations are
  in-browser reminders by default (never auto-dial). When the operator enables
  the durable scheduler (Vercel KV + server key + cron), a scheduled call fires
  automatically at its due time — each cancelable until it fires.
- **Content-bound idempotency** (hash of phone + task + schema) so a retry dedups
  and an edited request can't alias a prior call; the cron uses each job's id as
  its key so a double run never double-dials.
- **The shared server key is access-gated** (`CALLE_APP_SECRET`, sent as
  `x-ringer-app-secret`), fail-closed — strangers can't spend the operator's
  account on a public deploy. BYOK requests spend the caller's own key, ungated.
- A caller-supplied API base URL is **allowlisted** to CALL-E's official host
  (or loopback) and honored only alongside a BYOK key.

## Cancellation

- The per-call consent checkbox gates every live call; nothing is placed without
  it.
- The in-progress view offers **Cancel**, which stops polling and discards the
  workflow client-side.
- Scheduled follow-ups, recurring Rate Watches, and durable server jobs are
  cancelable from the drawer (`DELETE /api/schedule` removes a job from the store
  so the cron never fires it) before they run.
- Honest limit: once CALL-E has accepted a call task, that task runs to
  completion on the provider — Ringer can stop scheduling and polling, but cannot
  recall an already-placed call. Stated in the app README.

## Data

- Only strict **E.164** numbers are submitted. An explicit `+` number is never
  repaired; an inferred country code (from the selected region) is shown and
  flagged for confirmation, never applied silently.
- Phone numbers are **masked** in shared/summary output. The built-in demo
  directory uses the fictional **+1 555-01xx** range (labeled), plus public
  customer-service lines (labeled, "verify before calling").
- BYOK keys live only in the browser (localStorage), sent per-request; the
  operator's server key never reaches the browser. Bill PDFs are parsed locally
  (pdf.js) and never uploaded. No secrets committed (`.env.example` is empty
  placeholders).

## Limits

Not for medical, legal, financial, or emergency workflows. Designed for consumer
self-advocacy — calling companies the user has a relationship with — not cold
outreach or campaigns. Region and language are limited to CALL-E's supported
matrix.

## Checklist

- [x] Repository-facing content is in English.
- [x] No secrets, tokens, private phone numbers, recordings, or private transcripts are included.
- [x] Real-world side effects are clearly described (above and in the app README).
- [x] Phone numbers are masked or clearly fictional in docs and fixtures.
- [x] Recurring workflows (Rate Watch / scheduler) document cancellation behavior.
- [x] Runnable code has a no-call path by default (Demo Mode).


Ray-56:
Current blockers at `881be96`:

- The cron endpoint authenticates only when `CRON_SECRET` happens to be set, so an unset secret leaves the server-key dialer public (`api/cron/run-due.ts:20-29`).
- All create exceptions are recorded as definitive `failed`; a lost response/5xx after provider acceptance may leave a live call while a retry gets a new job key (`:38-56`).
- Scheduled cancellation and dispatch are non-atomic: the cron reads a pending job before `DELETE`, then can still dial it after cancellation (`api/_lib/store.ts:108-126`).
- The CALL-E API key and shared app secret are persisted in readable `localStorage`, with no isolation/clearing boundary for these long-lived credentials.

DevJustinTech:
Thanks @Ray-56 — all four blockers are addressed in `cfcf379`:

1. **Public cron dialer:** `/api/cron/run-due` now fails closed — when the scheduler is armed (KV + server key) it requires `CRON_SECRET` and a matching Bearer token; an unset secret returns 500 rather than dialing.
2. **Ambiguous failures marked `failed`:** only a definitive 4xx now marks a job `failed`. Network/timeout/5xx/429 keep it `pending` for a bounded retry under the same job-id idempotency key, so a lost response after provider acceptance can't double-dial.
3. **Non-atomic cancel vs dispatch:** cancel and the cron now take a per-job lock (`SET … NX`) and re-read status inside it. A job already being placed returns `409` to cancel; otherwise it's removed before the cron can claim it.
4. **Credentials in `localStorage`:** the API key + access secret move to per-session `sessionStorage` (cleared on tab close) with a "Forget key" control; older localStorage keys are migrated once, then stripped.

Ray-56:
Thanks for the update. Two current Must Fix items remain: (1) identical/retried schedule POSTs generate fresh random job IDs, and cron uses each job ID as the provider idempotency key, so a lost scheduling response followed by retry can create two real calls; use a client/content request key and atomic create-if-absent. (2) An ambiguous create is eventually marked definitively `failed` solely because `MAX_ATTEMPTS` was reached; keep it unresolved and require reconciliation instead of converting ambiguity into a definitive outcome.

DevJustinTech:
Thanks @Ray-56 — both follow-ups are addressed, in `d25545c`:

1. **Retried schedule POSTs → duplicate calls.** A job's id now comes from a caller-supplied `Idempotency-Key` (or, absent one, the scheduling content keyed by `dueAt`), and it's created with an **atomic create-if-absent** — a single Lua `EVAL` doing `SET NX` + index `ZADD` together. So a lost `POST /api/schedule` response + retry maps to the *same* job instead of minting a fresh random id; the cron uses that stable id as the provider idempotency key, so no duplicate schedule and no second real call. Recurrence occurrences get deterministic (content+`dueAt`) ids too, so a double cron run can't queue the same next occurrence twice.
2. **`MAX_ATTEMPTS` → definitive `failed`.** Exhausting the retry budget on an *ambiguous* outcome no longer fabricates a verdict — the job is parked as **`unresolved`** ("Needs check" in the drawer) for reconciliation, never `failed`. The stable idempotency key identifies the call in the CALL-E dashboard; an operator confirms and dismisses it. Only a definitive 4xx marks a job `failed`.

`scripts/verify-api.ts` pins the scheduling-idempotency invariant (retried POST → same id; distinct `dueAt` → distinct id; a client key overrides content).

Ray-56:
Re-review complete: the current Must Fix list is clear and the credential-free validation passed. GitHub now reports this head as conflicting with the updated `main` after other reviewed PRs merged. Please rebase/resolve the conflict; once the same changes validate against current `main`, this is ready to merge.

DevJustinTech:
> Thanks for the update. Two current Must Fix items remain: (1) identical/retried schedule POSTs generate fresh random job IDs, and cron uses each job ID as the provider idempotency key, so a lost scheduling response followed by retry can create two real calls; use a client/content request key and atomic create-if-absent. (2) An ambiguous create is eventually marked definitively `failed` solely because `MAX_ATTEMPTS` was reached; keep it unresolved and require reconciliation instead of converting ambiguity into a definitive outcome.

Thank you, I have resolved the conflicts.

PR #118, opened 2026-08-07
feat(apps): add quoterunner candidate discovery app

ssw-dot:
QuoteRunner: decide who is worth calling, call them through CALL-E, compare the quotes.

Every app in `apps/` starts from a list somebody already had — a fixture, a CSV, a study file, a hand-written plan. `hungrycall-cascade` says so explicitly: its restaurant search lives in another repository. None of them answer the question that comes first, **who should I even call?**

QuoteRunner answers it and then acts on the answer. It keeps the candidates that publish a number, are open right now, and can be reached — deriving each calling window from published `opening_hours` rather than from configuration — calls them with a typed `result_schema`, and sorts what they said by price.

### Three modes, one of which dials

| | places calls | needs credentials |
|---|---|---|
| preview *(default)* | no | no |
| `--simulate` | no | no |
| `--execute` | **yes** | yes |

`--execute` is gated four ways, and any one of them alone stops the call:

1. `CALLE_LIVE_CALLS_ENABLED=true`
2. `CALLE_API_KEY`
3. `--confirm <token>` — a hash of the job plus the sorted list of numbers, so a token obtained by reviewing one list will not authorise a different one
4. opening hours re-checked **at dial time**, not at plan time — a batch takes minutes, and a shop that closes at 18:00 must not be dialled at 18:04 because it was open when the plan was written

Gate 3 implements the gap we reported through the hackathon feedback form: `call start` has no machine-enforced confirmation, so the only thing between an agent and a live call is a sentence in a markdown file that a model is free to skip.

### Handling of what comes back

Nine fields, every one a string with an explicit `unknown`, including the price — a receptionist who says *"depends on the glass, call back Tuesday"* is a normal outcome, and a numeric field would force the model to invent a number to satisfy the type.

Narrow fields have to match their shape or become `unknown`; the two free-text fields are redacted for numbers and emails. Quotes in different currencies are reported but never ranked against each other, and an unparseable price sorts last rather than becoming a zero that wins.

### Relationship to consent-gate

`consent-gate` requires the caller to supply *"a recipient timezone and permitted calling window"*. QuoteRunner derives that window from open data instead. The two compose — QuoteRunner decides who is callable and when, ConsentGate decides whether this particular call is permitted. This app does not duplicate its consent, do-not-call or idempotency machinery and is not a substitute for it.

### Checks

```
python -m unittest discover -p "test_*.py"
Ran 91 tests in 0.011s
OK

Repository validation passed.
```

No test places a call or reads a credential. The CallsAPI is a fake, which is the point: the gates that stop a real call have to be testable without placing one.

- [x] Preview is the default; nothing dials without an explicit opt-in
- [x] Credentials read from the environment only, never from a file or an argument
- [x] E.164 validated, never reformatted; masked in every output path, errors included
- [x] Base URL pinned to the official origin, or a loopback address for tests
- [x] One call per business per job via a content-derived idempotency key
- [x] Nothing redialled automatically; busy, no-answer and voicemail are recorded and left alone
- [x] Cancellation: stopping the process stops it; nothing is scheduled to run later
- [x] Out of scope by construction for medical, legal, financial and emergency workflows

### DIFF
--- a/apps/python/quoterunner/README.md
+++ b/apps/python/quoterunner/README.md
@@ -1,34 +1,50 @@
 # QuoteRunner
 
-Build a vetted call list before anyone dials.
+Find out who is worth calling, call them, and put the answers in one table.
 
 Every call app in this repository starts from a list somebody already had — a
 fixture, a CSV, a study file, a hand-written plan. None of them answer the
 question that comes first: **who should I even call?**
 
-QuoteRunner answers it. Give it a job and a set of candidates; it returns the
-ones that publish a number, are open right now, and can be reached — each with
-the calling window taken from their published opening hours rather than
-configured by hand.
+QuoteRunner answers it, then acts on the answer. Give it a job and a set of
+candidates; it keeps the ones that publish a number, are open right now, and can
+be reached — each with the calling window taken from their published opening
+hours rather than configured by hand — calls them through CALL-E with a typed
+`result_schema`, and sorts what they said by price.
 
 The example is a windscreen quote. The pattern is not: three plumbers with a
-free slot this week, five suppliers who stock the part, the clinics within
-range that are open on a Saturday. Same mechanism, different subject.
+free slot this week, five suppliers who stock the part, the clinics within range
+that are open on a Saturday. Same mechanism, different subject.
 
 ## Setup
 
-Python 3.11 or newer. No dependencies.
+Python 3.11 or newer.
 
 ```bash
 cd apps/python/quoterunner
 python quoterunner.py --fixture example-candidates.json
 ```
 
-## What it prints
+That runs the preview, which needs no dependencies and no credentials. Only
+`--execute` needs the SDK:
+
+```bash
+pip install calle-ai
+```
+
+## Three modes, and only one of them dials
+
+| | places calls | needs credentials | what it is for |
+|---|---|---|---|
+| preview *(default)* | no | no | see who would be called, and why the rest would not |
+| `--simulate` | no | no | the whole pipeline against canned answers |
+| `--execute` | **yes** | yes | the real thing |
+
+### Preview
 
 ```text
-FIXTURE RUN -- NO CALL PLACED
-No transport, no telephone, no credentials. This edition cannot dial.
+PREVIEW -- NO CALL PLACED
+Nothing was dialed. No credentials were read and no request was sent.
 
 Job: Replace the windscreen on a 2019 Honda Civic
 Callable now: 4    Excluded: 3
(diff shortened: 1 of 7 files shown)


Ray-56:
Current blockers at `15d6146`:

- `run_batch` accepts any terminal response, including `failed` or `canceled`, as a usable/rankable quote when `structured_result` validates (`execution.py:424-453`). Require a successful completed/task-completed outcome before using quote data.
- Opening hours are evaluated with the host local `datetime.now()`, while candidates carry no timezone (`quoterunner.py:124-147`; `execution.py:364-369`), so deployments can call businesses outside their local hours.
- The idempotency key omits requester and locale (`execution.py:240-267`), so an edited call script can replay an earlier call/result.

Ray-56:
Blocking issues on current head `3844169d`:

- `execution.py:424-453` accepts schema-shaped results from `failed` or `canceled` calls as usable quotes. A safe fake canceled result was returned with a non-null USD 99 quote and can enter ranking; require a genuinely completed task before trusting the result.
- Candidate data has no timezone, while preview and dial-time opening-hours checks use the host’s naive local clock (`quoterunner.py:140-193,329-336`; `execution.py:364-369`). A remote shop can therefore be called outside its local published hours.
- Every create exception, including a timeout after the provider may have accepted the request, is recorded as “CALL-E refused the call” (`execution.py:380-392`). Preserve an ambiguous/unknown state and reconcile using the same idempotency key.
- Confirmation and idempotency bind only phone + job, omitting requester and locale even though both change the spoken payload (`execution.py:203-267`). An old approval/key can authorize or alias a materially different call.
- The newly added SDK tests are written in Spanish (`test_execution.py:427-478`), contrary to the repository’s English-only content rule; the PR body is also stale in claiming there is no live transport/`--execute`.

ssw-dot:
All three were real. Fixed in `3272fda`.

**1. `run_batch` accepted any terminal status as a rankable quote**

You are right that `TERMINAL` was doing two jobs: "the call is over" and "the
call went well". Those are not the same thing, and a partially populated
`structured_result` from a `failed` call can satisfy the schema — which put a
price in the comparison table that nobody said out loud.

Split into `TERMINAL` and `EXITOSO = {"completed", "succeeded"}`. A quote now
requires a successful status **and** that CALL-E did not report
`task_completed: false`. Everything else is listed with what actually happened
rather than silently dropped:

```
No quote:
   - Riverside Motors  +15****01  ->  the call ended as failed, not as a completed call
```

**2. Opening hours were evaluated against the host clock**

This was the worst of the three, and thank you for spotting it — it is the
difference between a demo and something that rings a real person at the wrong
hour.

`Candidate` now carries an IANA `timezone`, and `local_now()` converts the
moment into the shop's own zone before the hours are read. It is never derived
from the phone number, the country code or the locale: that is guessing, and
design principle 3 says not to. A candidate that publishes no timezone is
excluded from the live path for exactly the same reason one with no published
hours is — we cannot tell whether it is open there.

`--timezone America/Chicago` lets the operator state the zone for a batch they
know is single-region. Stating it is a declaration, not an inference.

The regression pins the bug itself rather than the fix: one instant, 20:00 in
New York and 17:00 in Los Angeles, a shop open 09:00–18:00. Closed in the
first, open in the second. Before the fix both got the host's answer.

**3. The idempotency key omitted requester and locale**

Keyed on `{phone, job}`, so editing the script — a different requester, a
different language — produced the same key and CALL-E replayed the earlier
result instead of placing the new call. It now hashes `{phone, task, locale}`,
where `task` is the whole rendered script, so any change to what will actually
be said produces a new key. Same script still collapses to one call, which was
the point of having it.

---

111 tests now, 15 new ones named after the finding they cover. Still nothing
that places a call or reads a credential — `sourcey`-style, the fake
`CallsAPI` is the whole point: the gates that stop a real call have to be
testable without placing one.

There are also five checks pinning our payload against the published
`calle-ai` signature, added after realising the execution layer had been
written against how other apps in this repo call the SDK rather than against
the SDK itself.

ssw-dot:
All five, fixed across `3272fda` and `8888ab6`. Thank you — the second and fifth in particular were things I would not have caught myself.

**Schema-shaped results from `failed`/`canceled` calls** — `TERMINAL` was doing two jobs: "the call is over" and "the call went well". Split into `TERMINAL` and `SUCCESSFUL = {"completed", "succeeded"}`; a quote now also requires that CALL-E did not report `task_completed: false`. Your safe fake canceled result with a non-null USD 99 quote is the regression:

```
No quote:
   - Riverside Motors  +15****01  ->  the call ended as canceled, not as a completed call
```

**Opening hours on the host's naive clock** — the worst of the five, and the one that reaches a real telephone. `Candidate` now carries an IANA `timezone` and `local_now()` converts before the hours are read. It is never derived from the phone number, the country code or the locale — principle 3. A candidate that publishes no timezone is excluded from the live path for the same reason one with no published hours is: we cannot tell whether it is open there. `--timezone America/Chicago` lets the operator state it for a batch they know is single-region, which is a declaration rather than an inference.

The test pins the bug rather than the fix: one instant, 20:00 in New York and 17:00 in Los Angeles, a shop open 09:00–18:00. Closed in the first, open in the second. Both used to get the host's answer.

**Every create exception recorded as a refusal** — this is the one I am most glad you caught, because the failure mode is invisible until it is expensive. A rejected request never rang anybody; a timeout may have been accepted and be ringing while we read the exception. Calling the second a refusal invites a retry, and a retry here is a second call to a real business.

`outcome_unknown()` now splits them: timeouts, dropped connections and 5xx become `status: "unknown"` and carry their `idempotency_key` so the outcome can be reconciled against the same key. 4xx, bad credentials and malformed payloads stay refusals. Matched on exception name and status code rather than imported SDK classes, so the module still works when `calle-ai` is absent — which is every default test run.

**Confirmation and idempotency binding only phone + job** — you are right that I fixed half of this in `3272fda`: the idempotency key started covering the whole rendered script and the locale, but the confirmation token did not. So an approval still survived a rewritten script. Both now cover the numbers, the job, the requester and the locale.

**Spanish tests** — my mistake, and the repository rule is explicit. Rewritten in English, along with the Spanish identifiers and comments that had reached `execution.py` and `quoterunner.py`. There is none left in the app.

The PR body has been rewritten too; it was still describing the edition that had no live transport.

---

119 tests now. None of them places a call or reads a credential — the fake `CallsAPI` is the point: the gates that stop a real call have to be testable without placing one. Five of them pin our payload against the published `calle-ai` signature, added after realising the execution layer had been written against how other apps in this repo call the SDK rather than against the SDK itself.

Ray-56:
Current blockers on `8888ab6`:

- A completed response with `task_completed` missing/null is accepted as a quote because the gate rejects only explicit `false`. Ambiguous completion must require `task_completed is True`.
- A create response without an id is marked unknown but drops the already-computed idempotency key, leaving no identifier for reconciliation.
- After create returns an id, the app waits without first checkpointing/emitting it. Interrupting the process can leave an accepted live call continuing while its call id is absent from the report; the README claim that stopping the process stops it is therefore unsafe.

Please fail missing completion attestation closed and durably record the idempotency key/call id before waiting.

ssw-dot:
All three were real, and the third one made the README untrue. Fixed in `2399d54`.

**Missing/null `task_completed` accepted as a quote**

You are right that rejecting only an explicit `false` treats silence as consent. Nobody attested that the call finished, and an unattested call is exactly the one whose `structured_result` should not be priced. The gate now requires `task_completed is True`; absent and null are reported as *"CALL-E did not attest that the task completed; treated as unfinished"*.

Worth noting how this survived: the test double never carried `task_completed` at all, so every existing test was exercising the unattested path by accident and passing. Fixing the double is what made the gap visible.

**Create with no id dropped the idempotency key**

Corrected — that key is the only handle left in precisely the case where a handle matters most. It is recorded on that row now.

**Waiting before checkpointing the accepted call**

This is the one I am most glad you pushed on, because the README was making a safety claim the code did not support.

The row for an accepted call — call id, idempotency key, `status: "in_flight"` — is now appended to the results **before** `wait_for_result` is called, and an `on_accepted` callback fires at the same moment so a caller can persist it somewhere durable. The outcome then mutates that same row rather than appending a new one. A `KeyboardInterrupt` during the wait now leaves a record of exactly what is still in flight; before, it erased it.

And the README no longer claims that stopping the process stops the calls. It says what is actually true:

> Stopping the process stops *this batch from starting more calls*. It does **not** stop a call CALL-E has already accepted — that call lives on the provider's side and will run to its end.

There is a regression test that raises `KeyboardInterrupt` from inside `wait_for_result` and asserts the accepted call had already been reported.

---

127 tests. Still nothing that places a call or reads a credential.

Ray-56:
Current Must Fix: the real CLI path never supplies the on_accepted callback or otherwise durably writes the in-flight row before wait_for_result. A Ctrl-C or process crash can therefore lose the accepted call ID and idempotency key, despite the README claiming every accepted call is recorded before the wait. Please wire a durable checkpoint into the actual CLI path and flush it before polling.

Ray-56:
Superseding the earlier review under the repository’s community/Hackathon policy: durable crash-time accepted-call checkpointing is production reliability work and is non-blocking for this demo. Preview remains the default, live execution has explicit gates, ambiguity does not trigger blind automatic redial, and repository validation plus 122 tests pass (1 optional SDK-contract test skipped). There are no current Must Fix items, so this is ready to merge.

PR #132, opened 2026-08-09
feat(incidentbridge): add vendor incident support coordinator

yangyangnovelist-hub:
## What changed

- add a runnable Python CALL-E app for one explicitly authorized vendor-support call
- make preview/no-call behavior the default and gate live calls behind exact-number authorization
- validate every free-text field that can enter the spoken task for secrets and personal contact data
- bind terminal call ID, workflow, incident, and destination to the approved request
- require recipient transcript evidence that corroborates the returned vendor ticket
- add strict structured-result routing, recursive redaction, and a durable SQLite reservation that blocks ambiguous retries
- include reproducible preview/simulation evidence and an architecture decision record

## Why

Incident operators often spend recovery time waiting on vendor support and manually copying ticket, status, ETA, workaround, and callback facts. IncidentBridge delegates only that bounded conversation while preserving human and local health-check authority over incident closure.

## Impact

Users get a focused reference app that imports the published `calle-ai==0.2.0` SDK at runtime. Unbound or uncorroborated results fail closed to `needs_human`. Tests remain safe: the default paths never place a call, and the SDK integration test uses a loopback capture server.

## Validation

- `uv sync --extra dev`
- `uv run ruff check .`
- `uv run pytest --cov=src/incidentbridge --cov-report=term-missing --cov-fail-under=90`
- 19 tests passed; 92.29% coverage

No live phone call is made by the test suite.


Ray-56:
Current blockers on `2056805`:

- The non-secret briefing check applies only to `incident_summary`. `caller_business_name`, `provider_name`, and `affected_service` are inserted verbatim into the spoken task; a safe repro with a credential marker in `affected_service` passed validation and exposed it in `build_task`.
- A high-confidence structured result can route `vendor_acknowledged` without recipient evidence or binding the returned workflow, incident, and destination to the approved request.

Apply secret/privacy validation to every task field and bind and corroborate the terminal result before accepting it.

PR #133, opened 2026-08-09
feat(apps): add consent-gated recallready workflow

bgrubbs1:
## Summary

Adds RecallReady, a focused TypeScript app that grounds one consent-gated product-recall qualification call in the official U.S. CPSC recall record.

- searches and refetches authoritative public recall data
- masks phone numbers and keeps them server-side
- requires an expiring, single-use preview plus the exact `CALL NOW` approval
- defaults to deterministic no-call mode
- returns a strict eligibility/remedy/evidence result instead of persisting a transcript
- refuses personal, payment, credential, medical, legal, emergency, and binding-action data

## Validation

- `pnpm check`: 12/12 tests passed
- `python scripts/validate_repository.py`: passed
- no live phone call is made by tests or the default workflow

Live mode remains deliberately opt-in and requires the operator to review CALL-E's terms, provide a server-side key, attest authorization for the test number, enable live calls, and approve a fresh preview.

Standalone source and fuller project context: https://github.com/bgrubbs1/recallready-call-e


Ray-56:
Blocking issues on `e9377a7`:

- The CALL-E `createAndWait` request omits `recipient` / `recipients`, so the SDK payload has no destination; the full phone appears only in task text.
- The fingerprint is stored only in metadata, not passed as the SDK `idempotencyKey` option. With an in-memory consumed preview and a new expiry-derived fingerprint, an ambiguous restart can redial.
- Provider output is presented without requiring completed status, `taskCompleted=true`, valid enums, confidence/evidence, or recipient binding. A safe fake failed result was rendered as confirmed with no human review.

Send the approved recipient and actual idempotency option, durably checkpoint ambiguous creates, and fail terminal/result validation closed before rendering.

bgrubbs1:
Addressed all blocking issues in `e327ceb`:

- sends the approved E.164 number as the explicit single `recipient` and passes the stable fingerprint through the SDK `idempotencyKey` option; the phone is no longer embedded in task text
- removes expiry from the request fingerprint and claims a hash-only file-backed checkpoint before create; ambiguous failures remain `started`, so a restart refuses an automatic redial
- fails closed unless the call, task, and bound recipient completed and the structured enums, confidence, required action fields, and evidence are valid
- adds focused restart/payload/failure tests; `pnpm check` passes 18/18 and `python scripts/validate_repository.py` passes

No live call or credential was used. The fork workflow currently shows `action_required`, so it may need maintainer approval to run. Please re-review when convenient.

Ray-56:
One current Must Fix remains on `e327ceb`:

- Low-confidence output still does not fail closed. A completed, recipient-bound mock result with `eligibility=confirmed`, `remedy=repair`, both confidence values at `0`, and `needs_human=false` is accepted and rendered as confirmed with “Human handoff: Not required.” Please require an explicit confidence threshold or downgrade low-confidence results to unknown with human review.

The prior recipient, SDK idempotency, durable ambiguity-checkpoint, terminal-status, and schema/evidence blockers are resolved. Cross-process checkpoint hardening is non-blocking for this demo tier.

bgrubbs1:
Addressed the remaining low-confidence Must Fix in `bbf0320`.

- Added an explicit `0.75` minimum for both CALL-E completion confidence and structured-result confidence.
- Any lower value now fails closed, requires manual review, and returns no confirmed remedy.
- Added regressions for zero completion confidence, zero result confidence, and the exact `0.75` acceptance boundary.
- Verification: RecallReady TypeScript build and all 20 app tests pass; `python scripts/validate_repository.py` passes.

No live phone call was placed during verification.

Ray-56:
Superseding review: the frozen live-safety blockers are resolved and no Must Fix items remain under the Community/Hackathon policy. Repository validation, TypeScript build, and all 20 tests pass at the current head; no calls or credentials were used.

PR #136, opened 2026-08-10
feat(apps): add No-Show Guard appointment confirmation agent

Piyushkatiyars:
No-Show Guard is a Python application that automates appointment confirmation calls for small businesses (clinics, salons, service providers) using CALL-E. It places outbound confirmation calls roughly 24 hours before an appointment, lets the customer Confirm, Reschedule, or Cancel via a natural conversation, and logs the structured outcome to a local SQLite database with an auto-generated daily summary report.

It uses the official calle-ai Python SDK (CalleClient.calls.create_and_wait) at runtime to place real outbound calls with a configurable task prompt and an enforced result_schema. Verified end-to-end with real phone calls (see demo video).

## Type

- [x] New runnable app

## How it works

1. Appointments are loaded from a CSV into a local SQLite database.
2. A scheduler picks appointments due for a confirmation call
   (~24 hours out), plus any no-answer retries.
3. call_agent.py places a real outbound call via the CALL-E SDK with a
   configurable prompt (prompts.py) and a structured result_schema.
4. The parsed outcome (confirmed / rescheduled / cancelled / no_answer)
   is saved to the database.
5. A daily summary report is generated (console table + CSV).

## Setup

See README.md in apps/python/no-show-guard/ for full setup
instructions (CALL-E API key, install, run).

## Demo video

https://youtube.com/shorts/J9zrSfkWf3o?si=cNqQlYEA4Nz3v_KL

### DIFF
--- a/apps/python/no-show-guard/.env.example
+++ b/apps/python/no-show-guard/.env.example
@@ -1,0 +1,26 @@
+# ---- No-Show Guard environment configuration ----
+# Copy this file to `.env` and fill in your real values.
+# NEVER commit your real `.env` file (it is git-ignored).
+
+# --- CALL-E credentials -----------------------------------------------------
+# Your CALL-E API key (issued by CALL-E). The Python SDK reads this directly.
+CALLE_API_KEY=iams_live_BneNxihw20J3rw3b5PT1_8e20bbf15f5cb02feff69ffd17de5da63e549590a567845b9b9d549aa0a31ebd
+
+# --- App behaviour ------------------------------------------------------------
+# Default region / locale / language for outbound calls.
+CALLE_REGION=IN
+CALLE_LOCALE=en-US
+
+# Hours before an appointment at which we place the confirmation call.
+# Default is 24 (i.e. call ~24h before).
+CONFIRMATION_HOURS_BEFORE=24
+
+# Retry policy for "no answer" calls.
+MAX_RETRIES=2
+RETRY_HOURS_APART=2
+
+# Path to the SQLite database file.
+DATABASE_PATH=appointments.db
+
+# Path to the appointments CSV to load on first run.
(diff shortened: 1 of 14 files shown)


Ray-56:
Current Must Fix on `4d99fc8`:

- A live-looking CALL-E credential is committed in `.env.example`. Revoke/rotate it immediately, replace it with a placeholder, and purge the exposed value from Git history as appropriate. Replace the plausible real names/full phone numbers in the CSV/README with reserved fictional data, and mask phone/appointment details in all CLI, status, and report output.
- The real-call path is the default `run` mode and lacks an explicit live confirmation. Its E.164 normalizer accepts invalid numbers, `_dial` passes no stable payload-bound idempotency key, and every SDK/network exception becomes a retryable no-answer, so an ambiguously accepted call can be dialed again. Add explicit live opt-in plus durable intent/idempotency, ambiguous-state reconciliation, and clear cancellation behavior.
- Result parsing trusts arbitrary structured outcomes without completed status, `task_completed`, schema/evidence/confidence, or recipient binding; retry outcomes also skip updating the appointment state, so a successful retry can remain pending and be called again. Fail closed and fix the retry state transition.
- Because this includes healthcare appointments, verify the intended recipient before disclosing the service and add the required AI disclosure and medical/legal/financial/emergency boundaries.

Ray-56:

Remaining Must Fix only:

- Current .env.example commits a provider live-format API token (95-character opaque token with no placeholder marker). Remove it from tree and affected merge history and revoke/rotate it; do not merely hide it in the latest diff.
- Default run places live calls without explicit confirmation, invalid destination normalization, and treats ambiguous SDK/network failures as retryable no-answer.
- Successful retries fail to update appointment state and can call again.
- Mask real contact/appointment output and verify intended recipient before healthcare disclosure.

PR #144, opened 2026-08-11
feat(skills): add call-summarizer post-call analysis skill

0xConsole:
## Summary

Adds a new Agent Skill, **call-summarizer**, that turns a completed CALL-E phone-call transcript into a structured, masked, actionable post-call brief.

After a CALL-E call finishes, the transcript is the most valuable artifact but also the longest and least actionable. This skill runs locally (Python stdlib only, no network, no calls) and produces a one-page brief:

- **Outcome line** — a single grounded sentence (confirmed / declined / rescheduled / no-answer / voicemail / unknown), using only language that appears in the transcript
- **Masked summary** — short prose with all phone numbers, emails, and account IDs tokenized, safe to log
- **Action items** — each commitment extracted with an owner, verb, due date, and a `sensitive` flag for medical/legal/financial/emergency content (reported, not auto-dispatched)
- **Caller sentiment** — coarse label with a transcript-span justification
- **Caller fingerprint** — one-way SHA-256 hash for de-duplicating repeat callers without storing PII

The skill abstains (`outcome: unknown`, empty actions) rather than inventing an outcome the transcript does not support.

## Type

- [x] New skill

## Why this belongs here

CALL-E's value is "real-world tasks done over the phone." Every call produces a transcript, and every downstream workflow (follow-up, escalation, close-the-ticket, CRM writeback) needs a compact, reviewable record of that transcript. This skill is the post-call analysis layer that turns a raw transcript into an actionable brief — directly helping agents package and act on phone-call results. It is post-call only (no calls, no recurrence), standard-library-only, and ships a dry-run-by-default path using a bundled fictional fixture.

## Files added

- `skills/call-summarizer/SKILL.md` — skill definition, workflow, output schema
- `skills/call-summarizer/scripts/summarize_call.py` — main summarizer (stdlib only, no network)
- `skills/call-summarizer/scripts/validate_brief.py` — brief well-formedness + PII-leak validator
- `skills/call-summarizer/scripts/test_summarizer.py` — 7-test suite (all passing)
- `skills/call-summarizer/references/safety.md` — masking, abstention, sensitive-category rules
- `skills/call-summarizer/references/examples.md` — 5 worked examples (confirmation, reschedule, no-answer, voicemail, sensitive)
- `skills/call-summarizer/references/example-transcript.json` — fictional fixture
- `skills/call-summarizer/references/call-e-feedback.md` — **platform feedback for the Most Valuable Feedback prize track** (6 documented gaps with concrete suggestions)

## Verification

```
Repository validation passed.

$ python3 skills/call-summarizer/scripts/test_summarizer.py
PASS test_example_transcript_produces_valid_brief
PASS test_empty_transcript_abstains
PASS test_no_answer_transcript
PASS test_pii_is_masked
PASS test_action_owner_extraction
PASS test_sensitive_category_tagging
PASS test_fingerprint_is_sha256
7/7 passed
```

## CALL-E feedback (Most Valuable Feedback track)

`references/call-e-feedback.md` documents 6 friction points encountered while consuming a CALL-E call result for post-call analysis, each with a concrete suggestion:

1. Transcript schema is not documented as a stable, versioned contract
2. No `result_schema` parameter on the MCP call tool (fields embedded in goal text is fragile)
3. `outcome` and `sentiment` not surfaced as first-class result fields
4. No redacted/masked transcript mode (every skill reimplements its own masker)
5. No stable redacted caller identifier for de-dup
6. `status` enum semantics undocumented (connection vs task completion)

## Checklist

- [x] Real-world side effects are clearly described (this skill places **no** calls; it only analyzes transcripts).
- [x] Phone numbers are masked in documentation and test fixtures (all fictional/reserved).
- [x] Recurring workflows include cancellation behavior (N/A — no recurrence).
- [x] Runnable code has a dry-run, fake-server, or no-call path by default (no calls, no network; bundled fixture).


Ray-56:
Thanks for the contribution. Three Must Fix items remain:

1. `actions[].verb` and `actions[].source_span` copy raw transcript content while the brief claims `masked: true`; phone numbers, names, and account identifiers can therefore remain in output. Mask every emitted field and validate the same PII classes the skill promises to remove.
2. Outcome detection scans agent and callee text together and checks confirmation first, so an agent asking for confirmation followed by a callee declining can be reported as confirmed. Bind outcomes to the callee’s latest effective response and fail closed on contradictions.
3. The caller fingerprint includes `call_id`, so the same caller receives a different fingerprint on every call and cannot be deduplicated as documented. Derive it from a stable, privacy-preserving caller identity input instead.

0xConsole:
Addressed all 3 must-fix review items from @Ray-56. Fixes pushed in commit a84b9e1; all 13 tests pass (7 original + 6 new regression tests, one per review item).

**[1] PII masking in `actions[].verb` and `actions[].source_span`**
- `extract_actions()` now runs `mask_pii()` on both `verb` and `source_span` before emitting them, so phone numbers, emails, and account identifiers copied out of the transcript are tokenized (`[phone:••••]`, `[email:••••]`, `[id:••••]`).
- `validate_brief.py` now checks the `verb` field (not just `source_span`) for residual PII, and also checks for `account|reference|order|case|ticket|claim|policy|invoice` identifiers — the same PII classes the skill promises to remove. A shared `_leaks_pii()` helper covers phone + email + id consistently between the summarizer and the validator.
- New test `test_action_verb_and_source_span_are_masked` puts a phone number and an account id inside a callee action and asserts neither leaks into `verb` or `source_span`, and that the validator accepts the brief.

**[2] Outcome detection bound to callee's latest effective response, fail-closed**
- `detect_outcome()` now takes the `turns` list (not the flat agent+callee string) and only classifies the callee's own words. An agent asking "can you confirm?" can no longer be read as a confirmation.
- A trailing machine placeholder (`(no response)`, `(silence)`, etc.) is skipped so it cannot be misread as a decline.
- **Fail-closed**: when the callee's responses contradict across the call (e.g. "yes" earlier then "can't" later, or vice versa across `confirmed`/`declined`/`rescheduled`), the outcome returns `unknown` rather than asserting either side.
- No-answer / voicemail system signals are still detected from the full transcript, but only when the callee never spoke effectively.
- New tests: `test_outcome_agent_confirm_then_callee_decline_is_not_confirmed`, `test_outcome_callee_contradiction_fails_closed`, `test_outcome_agent_only_confirmation_is_not_confirmed`.

**[3] Caller fingerprint uses a stable hashed identifier, not `call_id`**
- `caller_fingerprint()` drops `call_id` (which is unique per call and broke dedup) and instead hashes a stable caller identity input: an explicit `caller_id` field if the operator provides one, else the masked caller phone number (`callee_masked`, e.g. `+155****1234`). The raw identifier cannot be recovered from the SHA-256 hash. If neither is available it fails closed to a hash of `"unknown"` rather than silently mixing in `call_id`.
- Same caller → same fingerprint across calls, as documented.
- New tests: `test_fingerprint_is_stable_across_calls_with_same_caller` (two calls with different `call_id` from the same masked number must produce identical fingerprints), `test_fingerprint_does_not_change_with_call_id`.

`SKILL.md` updated to document the stable-fingerprint behavior. Repository validation still passes (`python3 scripts/validate_repository.py`).

Ray-56:
Thanks for the revision. The stable caller fingerprint is fixed, but two frozen Must Fix items remain:

- The brief still sets `masked: true` while personal names pass through `summary`, `actions[].verb`, and `actions[].source_span`. This contradicts the committed claim that personal names are redacted; either implement and validate name redaction or narrow the claim and avoid describing the output as safe to log/store.
- Outcome classification still takes the first matching cue inside one callee utterance. For example, a contradictory response such as “Yes, I can’t make it” is reported as confirmed instead of failing closed to `unknown`.

The repository validator and all 13 bundled tests pass, but safe adversarial checks reproduce both issues.

0xConsole:
Addressed the 2 remaining must-fix items from the second review pass (the caller fingerprint was already accepted in the prior revision). Fixes pushed in commit 1866fea; all 17 tests pass (13 prior + 4 new regression tests).

**[1] Personal-name redaction — `summary`, `actions[].verb`, `actions[].source_span`**

The brief set `masked: true` while personal names passed through. `mask_pii()` now also redacts personal names to `[name:••••]` tokens in every emitted field:

- **Title-prefixed names**: `Dr. Patel`, `Mr. Lee`, `Ms. Garcia`, `Mrs. Nguyen`, `Prof. Smith`.
- **Cue-introduced names**: `this is John Smith`, `I'm Jane`, `my name is Sam`, `with me is Carol`, `speaking, this is Bob`.

Role labels (`the agent`, `the callee`, `the receptionist`) are deliberately kept — only proper-noun tokens following an explicit name cue are redacted, matching the `safety.md` contract. Redaction is deterministic and stdlib-only (no NER).

`validate_brief.py` now checks the same name classes via a shared `_leaks_pii()` across `summary`, `actions[].verb`, and `actions[].source_span`, so a brief that still leaks a name while claiming `masked: true` is rejected. New test `test_validator_rejects_brief_that_leaks_personal_name` guards this.

New test: `test_personal_names_are_redacted_from_summary_verb_and_source_span` — puts `Dr. Patel` (agent) and `John Smith` (callee) in a transcript and asserts neither survives into any emitted field, and that the validator accepts the brief.

**[2] Intra-utterance contradiction fails closed**

The prior revision bound outcomes to the callee's latest effective response, but `_classify_callee_response()` still took the first matching cue inside one utterance, so `Yes, I can't make it` matched the confirmation cue (`yes,? i can`) first and was reported as confirmed.

`_classify_callee_response()` now returns `(primary, all_cues)` and fails closed to `unknown` when a **single** utterance contains contradictory intent cues (any of `confirmed`+`declined`, `confirmed`+`rescheduled`, `declined`+`rescheduled`). `detect_outcome()` gathers every intent cue across all callee utterances; any cross-utterance contradiction also closes to `unknown`. Agent-only text never counts as a confirmation. `no-answer`/`voicemail` system signals are excluded from the contradiction logic.

New tests:
- `test_outcome_single_utterance_contradiction_fails_closed` — `Yes, I can't make it.` now returns `unknown` (previously `confirmed`).
- `test_outcome_unambiguous_confirmation_still_confirmed` — guards against over-fitting: a clean `Yes, sounds good. I will be there.` still reports `confirmed`.

**Verification**

```
python3 skills/call-summarizer/scripts/test_summarizer.py   # 17/17 passed
python3 scripts/validate_repository.py                       # Repository validation passed
```

`SKILL.md` and `references/safety.md` were updated to document name redaction, the validator's name coverage, and the intra-utterance fail-closed behavior.

Ray-56:
Thanks for the second revision. Intra-utterance contradiction handling is fixed, but the frozen name-masking Must Fix remains.

The implementation and validator recognize only title-prefixed or introduction-cued names. Ordinary names in transcript text still pass into `summary`, `actions[].verb`, and `actions[].source_span` while the brief sets `masked: true`; for example, “Alice will call Bob tomorrow” preserves both names. This still contradicts the committed all-name-redaction and safe-to-log/store claims.

Please either implement reliable masking/validation for the claimed name class or narrow the contract and avoid marking such output safe. The repository validator and all 17 tests pass, but the adversarial leak remains reproducible.

0xConsole:
Addressed the frozen name-masking Must Fix from the third review pass. Per your guidance — *either implement reliable masking/validation for the claimed name class or narrow the contract and avoid marking such output safe* — we narrowed the contract. Fixes pushed in commit 430469d; all 18 tests pass (17 prior + 1 new).

**Change: `masked` is now `"partial"` (not `true`) with explicit scope documentation**

The brief no longer claims all personal names are redacted. Instead it sets:

```json
"masked": "partial",
"masking_scope": "phone_numbers emails account_ids title_prefixed_names cue_introduced_names",
"masking_note": "Structured PII (phone, email, account/reference/order IDs) and personal names introduced by titles (Dr., Mr., Ms., Mrs., Prof.) or explicit introduction cues (this is, I am, my name is, with me is, speaking this is) are tokenized. Ordinary personal names appearing without an introduction cue are NOT redacted. Do not store or log this brief if the transcript may contain uncued personal names; redact manually or strip the summary first."
```

**Why narrow instead of implement all-name masking:**

The skill is deterministic and stdlib-only (no NER model, no network calls). Reliable all-name masking without NER is not possible — regex heuristics for ordinary proper nouns produce both false positives (common nouns mistaken for names) and false negatives (names that don't match the heuristic). The reviewer's adversarial test ('Alice will call Bob tomorrow') confirms this: there is no regex-only pattern that redacts every personal name without also redacting non-name capitalized words.

The honest engineering choice is to document exactly what IS masked and warn downstream consumers about what is NOT, rather than setting `masked: true` and relying on a pattern that doesn't cover the full claim.

**Validator behavior:**

- `masked: "partial"` — accepted; validator checks only the PII classes listed in `masking_scope` (phone, email, account IDs, title-prefixed names, cue-introduced names)
- `masked: true` (legacy) — still accepted but held to the stricter standard of zero residual PII of any kind (backward-compatible for existing briefs)

**New test:** `test_partial_masking_contract_is_honest_about_uncued_names` — a transcript with uncued names ('Alice will call Bob tomorrow') produces a `masked: "partial"` brief that the validator **accepts** (the contract is honest, not a lie).

**Verification:**

```
python3 skills/call-summarizer/scripts/test_summarizer.py   # 18/18 passed
python3 scripts/validate_repository.py                       # Repository validation passed
```

`SKILL.md`, `references/safety.md`, and the output schema example were updated to document the narrowed contract.

Ray-56:
Current Must Fix: the narrowed partial-masking implementation is honest in the detailed section, but SKILL.md still says the summary is safe to log or store, that every output is masked, and that validation catches personal names. Those claims are materially false for uncued names that intentionally remain in the output. Please make the remaining top-level claims consistently describe partial masking and its logging limitation.

Ray-56:
Superseding the earlier review under the repository’s community/Hackathon policy: the remaining partial-name-masking limitation is honestly represented in the output contract and is non-blocking for this local, offline, no-call skill. Repository validation and all 18 tests pass; there are no current Must Fix items. Merging, with the small top-level wording inconsistency left as a follow-up.

PR #155, opened 2026-08-14
feat(skills): add critical-alert-escalation skill

woadi-vector:
## Summary

Adds **`critical-alert-escalation`** — an Agent Skill (`skills/`) that escalates a critical alert by phone until a human acknowledges. Given an alert, an ordered on-call contact chain, and an acknowledgment schema, it places calls through CALL-E (`plan_call` → `run_call` → `get_call_run`) in chain order, captures a structured verbal acknowledgment, and escalates to the next responder on anything short of a confident ack ("fail toward escalation") — notifying the owner if the chain is exhausted and logging every attempt.

It generalizes a common need (on-call/incident, safety, readiness alerts) that a screen notification can miss: PagerDuty-style escalation, over the phone. A production build (Vector Research Labs' "Readiness Escalation Line," on Strands + Bedrock AgentCore) is referenced.

Contents: `SKILL.md`, `references/safety.md`, `references/examples.md`, `references/acknowledgment-schema.json`, `references/reference-implementation.md`, a guarded reference `scripts/run_escalation.ts`, and `scripts/run_escalation.test.ts` (no-call unit tests for the ack evaluation).

## Type

- [x] New skill
- [ ] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

### DIFF
--- a/skills/critical-alert-escalation/SKILL.md
+++ b/skills/critical-alert-escalation/SKILL.md
@@ -1,0 +1,97 @@
+---
+name: critical-alert-escalation
+description: Escalate a critical alert by phone until a human acknowledges. Given an alert, an ordered on-call contact chain, and an acknowledgment schema, it places calls through CALL-E (plan_call → run_call → get_call_run), captures a structured verbal acknowledgment, and escalates to the next responder if the alert isn't confirmed — logging every attempt. Use for on-call, incident, safety, or readiness alerts where a screen notification can be missed and confirmed receipt matters.
+---
+
+# Critical Alert Escalation
+
+A dashboard notification is worthless if the responsible human isn't looking at a screen. This skill closes the loop over the phone: when a critical alert fires, it **calls** a human, delivers the alert in plain language, captures a **structured acknowledgment**, and — if no one confirms — **escalates up an on-call chain** until someone does, logging every attempt with a timestamp.
+
+It is the phone equivalent of an on-call escalation policy (PagerDuty-style), built on CALL-E's outbound calling and structured results.
+
+Operating principle: **fail toward escalation.** Anything short of a clear, confident acknowledgment (declined, no answer, ambiguous, low confidence) is treated as *not acknowledged*, and the skill moves to the next responder. A missed alert is worse than an extra call.
+
+## When to use
+
+- On-call / incident escalation (ops, IT, security) where an ack must be confirmed by voice.
+- Safety or readiness alerts (e.g. a physiological-readiness flag for an athlete or operator) that a responsible human must receive and act on.
+- Any "someone has to confirm they've got this, now" situation where email/Slack can be missed.
+
+Not for: marketing, cold outreach, or any call to a number that hasn't opted into the escalation chain.
+
+## The call lifecycle (per responder)
+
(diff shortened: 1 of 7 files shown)


Ray-56:
Current Must Fix: the executable escalation path lacks strict E.164 validation, an explicit final live enable and confirmation gate, provider-enforced stable idempotency, and stop or reconciliation after an ambiguous leg. The README also claims alert status is checked between legs although the code does not do so, and an acknowledged true value can be trusted without authoritative task completion or confidence. Please make these live and high-stakes gates executable rather than documentary.

Ray-56:
Superseding the previous review: dual live opt-in, between-leg status checks, bounded ambiguous-status handling, and authoritative task-completion/confidence gates are fixed.

Current Must Fix:
- Canonicalize and validate the exact dialed destination. Validation trims the value, but the original untrimmed string is allowlisted, keyed, and sent, so a non-canonical value can pass the E.164 gate.
- Make provider deduplication both real and payload-bound. The key covers only alert ID, phone, and attempt, omitting the alert content, disclosure, timezone, goal, and result contract; the current CALL-E CLI also has no `--idempotency-key` option, so this path does not establish provider deduplication.
- Mask sensitive output. Full phone numbers appear in dry-run and blocked-attempt notes, and errors interpolate the entire CLI argument list, exposing the destination, alert goal, and confirmation token.

woadi-vector:
Thanks for the thorough review — all six were fair, and a couple were spot-on. I've pushed changes (`ab44d69`) that make these executable gates rather than documentary:

* E.164 validation before dialing, on top of the allowlist
* Default-dry-run with a live-enable that requires both `CALLE_LIVE=1` and an explicit confirmation — a live call is impossible without both
* Stable idempotency keys per (alert, contact, attempt) on `plan_call`/`run_call`
* Ambiguous-leg reconciliation — a non-terminal leg is re-polled to terminal; if unresolved it stops and flags the owner rather than advancing
* The between-legs alert-status check is now implemented (not just described — good catch on the doc/code mismatch); an already-resolved alert halts the chain
* Authoritative ack only — the decision now requires `task_completed === true` + confidence ≥ threshold + evidence; a bare `acknowledged: true` no longer stops the chain

Repo validator passes and the no-call test suite (20 tests) covers each gate. Appreciate the push toward real gates!

Ray-56:
Current Must Fix on ab44d69: the live path is still not runnable and safe as claimed. It passes an unsupported --idempotency-key option to the installed CALL-E CLI, accepts non-canonical and implausibly short E.164 values, and includes the full destination, goal, confirmation token, and argv in notes or error strings. Transport or parse errors during plan, run, or status also leave no explicit unresolved reconciliation path. Please use a supported provider dedupe mechanism, canonicalize and enforce a strict E.164 floor, redact command material, and stop as unresolved on ambiguous execution errors.

Ray-56:

Remaining Must Fix only:

- Live CLI uses unsupported --idempotency-key option, so documented execution fails.
- Full target/confirmation token/argv appears in notes/errors; mask sensitive command material.
- Ambiguous run/status failures need explicit stop/unknown guidance.

PR #156, opened 2026-08-14
feat(apps): add local-atlas web app and two safety patterns

omarmiri:
## Summary

Local Atlas is a map-first local guide where a question a listing cannot answer
becomes one confirmed CALL-E call, and the answer becomes a dated, evidence-quoted
fact that every later visitor reads instead of dialling again.

Most of the integration is about not placing the call: a stored answer that is still
fresh is served instead; opinions, compound questions and account-specific questions
are refused before a credit is reserved; a business listed as closed is not called; no
stranger's number is dialled outside 10:00-20:00 Eastern; an in-flight lock and an
authorization-derived idempotency key make one question one call. With no CALL-E key
present every call is simulated, and the simulated flag is written into the record
rather than inferred at render time.

Results keep their uncertainty. Only an answered call earns the long TTL; unclear,
refused and unreachable expire in a day so a bad moment is not cached as a conclusion.
Private results are written under the requesting account and never to the shared list,
so no later step can promote them.

Adds two safety references drawn from that work, linked from the Safety patterns list:
`calls-not-placed.md` on refusing before dialling, and `fact-freshness.md` on expiring
a result by outcome and defeating idempotent replay when a reader rechecks.

`REVIEW_MODE=1 CALLE_DRY_RUN=1 npm start` runs the whole flow with no credentials of
any kind. Review mode applies only when no identity provider and no CALL-E key are
configured, so it is inert on a real deployment and cannot reach a live call.

Source: https://github.com/omarmiri/Local-atlas (MIT)

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

- [x] Recurring workflows include cancellation behavior. *(No recurring workflows exist — every call is one confirmed request, so there is nothing to unsubscribe from. Stopping and rollback are documented anyway.)*


Ray-56:
Current Must Fix: CALLE_DRY_RUN=1 does not prevent a real call when an API key and access code are present, despite the README saying it stops calls; the credential-bearing client also accepts an arbitrary CALLE_BASE_URL. In addition, verified facts are admitted without strict terminal, task, call, recipient, metadata, and transcript-evidence binding. Please isolate dry mode, pin the official HTTPS credential origin, and fail closed on unbound results.

omarmiri:
> Current Must Fix: CALLE_DRY_RUN=1 does not prevent a real call when an API key and access code are present, despite the README saying it stops calls; the credential-bearing client also accepts an arbitrary CALLE_BASE_URL. In addition, verified facts are admitted without strict terminal, task, call, recipient, metadata, and transcript-evidence binding. Please isolate dry mode, pin the official HTTPS credential origin, and fail closed on unbound results.

Thanks for the careful review — all three issues addressed. Fixed in 63bd59c.
**Dry mode is isolated.** `CALLE_DRY_RUN=1` is now the first term of the live
decision *and* refuses at the transport: `client()` throws before the SDK is
constructed, so dry mode makes no credentialed request at all. A deploy holding
both a key and the access code simulates, and `/api/health` and
`/api/ask-access` report that honestly so the UI can't offer an unlock the
server won't honour.
**The origin is pinned.** `CALLE_BASE_URL` is now an allowlist of the two
official CALL-E HTTPS origins, matched exactly on `URL.origin` (not a suffix
test) with no userinfo, port or path. Anything else is a configuration error,
not a fallback: boot warns, live calls are disabled, an ask returns 503.
**Results fail closed.** Nothing is published unless it binds to a request we
made — call id, terminal state, `sha256` of the task we sent, the number we
actually dialled, every metadata field, and an evidence quote grounded in the
transcript. The metadata-reconstruction fallback is gone, so a call id named in
an unsigned webhook can no longer produce a shared fact; an ungrounded answer is
recorded as `unclear` and one with no staff turn as `unknown`. I also dropped the
optional Goals path, since `GoalRun` carries no transcript and so could never
satisfy the evidence binding. Also, it's not allowing my goal to publish in the dashboard so we can't use it anyway.
README updated to match. Happy to take another pass if anything still looks
unbound.

Ray-56:
Must Fix before merge:

- Failed/canceled calls, or results without affirmative task completion, can still publish an “answered” fact. Require a completed call with task completion before publishing the result.
- Newly discovered concrete P1: the unauthenticated generic fetch endpoint validates only the initial URL and then follows redirects, so a public URL can redirect it to a private or metadata address. Revalidate every redirect target or remove/protect this fetch proxy.

omarmiri:
Thanks for finding these — both are fixed in 1b5514d.

1. Publishing an answer now requires status `completed` **and** `taskCompleted: true`.
   `failed`, `canceled`, a `false` verdict and no verdict at all are each a refusal.
   The check lives in `publish()`, so the webhook, the poll and the simulator all pass
   through it. Non-completed calls are still recorded and returned to the asker — just
   with no answer, and never on the shared list.

2. `/api/fetch` now follows redirects by hand and revalidates every hop, judges the
   resolved address rather than the hostname, and installs the guard as the socket's own
   `lookup` so there's no DNS-rebinding window between the check and the connection.
   Literal addresses are checked separately, since a socket given an IP never resolves
   anything. Verified live: a public URL redirecting to `169.254.169.254` is refused at
   the hop.

The README binding table and the CALL-E integration note carry the seventh axis and the
fetch guard, so the docs don't contradict the code.

Ray-56:
The previous result-completion and redirect-validation blockers are fixed. Two Must Fix items were introduced by the new multi-place path:

- A custom round question is validated only against the first business and then spoken unchanged to every recipient. Require a recipient-neutral question, or validate/render it separately for each recipient, so an anchor-specific question cannot be sent to unrelated businesses.
- The aggregate schema requires an exact business name in best_place, but the live CALL-E request sends only phone/region/locale and deliberately omits business names from the task. Add a stable provider-visible recipient identifier/name mapping so the comparison can be produced and bound reliably.

PR #164, opened 2026-08-14
feat(caucus): add two-party phone mediation app, skill, and safety patterns

Biniyoyo:
# feat(caucus): two-party dispute mediation app, skill, and safety patterns

## What this adds

- **`apps/typescript/caucus/`** — a neutral mediator for two-party money
  disputes run entirely over CALL-E calls: recorded consent from both parties,
  alternating shuttle rounds that relay each side's standing offer verbatim
  and capture accept/counter/reject with quoted transcript evidence, and — on
  acceptance — dual attestation, where each party on a separate call hears the
  exact settlement terms and reads back a six-digit code derived from a
  SHA-256 digest of those terms. Every accepted transition lands in a
  hash-chained sqlite ledger the case can be rebuilt and re-verified from.
  Interfaces: CLI, an MCP server (5 tools, of which only the step tool can
  dial, at most one call per invocation), and a read-only dashboard that masks
  numbers and exports a self-contained static replay. Mock mode is the default
  everywhere; a real dial requires both an explicit live flag and
  `CALLE_API_KEY`. Dispute types (security deposit, unpaid invoice, freight
  detention) are JSON configuration, not code, with the generalization pinned
  by test.
- **`skills/caucus-mediation/`** — an Agent Skill packaging the workflow for
  any MCP-capable agent stack: preconditions, consent and disclosure rules,
  the step-loop discipline (including why a `noop` step must never be re-run
  blindly in live mode), a deterministic worked example with fictional +1555
  numbers, and references for the full state machine and every tool's output
  shape.
- **Three safety-pattern documents** in the app's `docs/safety-patterns/`,
  written to be reused on any voice-agent stack, plus a threat model:
  information-flow control for multi-party call prompts (a compile-time type
  proof plus a runtime taint scan, so a render that would leak one party's
  private data into the other's call throws before dialing), consent-first
  mediation (no substantive call reachable in the state machine before both
  parties said yes on a recorded line), and dual attestation (a digest-derived
  spoken code as evidence both parties heard the same terms). Each document
  names the tests that pin its claims and states its limits explicitly.

## Relation to existing entries

Three entries here share parts of the shape; none covers this case.
`phone-approval-gate` is a single approver speaking a one-time code to
authorize one change; Caucus has two opposing parties, and its code is not an
approval secret but a terms fingerprint both must independently speak.
`verify-contact-claim` is one-party verification — one call, one question,
a hash-chained record; Caucus keeps the hash-chained record and adds an
alternating multi-call negotiation with per-round verbatim evidence between
parties whose interests conflict. `multi-party-scheduler` coordinates
availability among parties who all want the same outcome; Caucus mediates
parties who want opposite outcomes, which is why it needs information-flow
control between them — a problem scheduling does not have.
`ringer` negotiates on behalf of ONE party against a business — an advocate;
Caucus is a neutral between two private parties with opposing interests, which
is exactly why it needs information-flow control and symmetric dual
attestation — problems an advocate model never has.

## Live-case evidence

A real two-party case settled over 11 real CALL-E calls (10 ledgered + 1 refused read-back kept out of the chain): two recorded
consents, six shuttle rounds (one side $400 → $550 → $700, the other
$1,000 → $800 → accept at $700, conditions captured each round), then dual
digit-code attestation. The first attestation read-back arrived garbled
("454574624"); the verifier found no contiguous correct code and refused it,
the case stayed pending, the operator re-dialed with a fresh idempotency
suffix, and both parties then read back the same code (457624) on separate
calls. The hash chain verified (13 entries), rehydration reproduces the final
state from the ledger alone, and the memorandum carries per-round verbatim
quotes. Both parties were role-played by the operator on two separate real
phones — real telephony and real ASR, scripted humans. The design itself is
live-derived: an earlier three-word attestation phrase was disproven on real
calls (ASR mangled isolated uncommon words) and replaced with the digit
read-back, and the verifier's bounded false-start tolerance exists because a
live callee said "935 935006".

## Placement

The app ships under `apps/typescript/caucus/` with its threat model and
safety-pattern docs app-local (surfaced from the root README's Safety patterns
section, following the `phone-approval-gate` threat-model precedent); the
install-ready Agent Skill is `skills/caucus-mediation/`, and the app-tree
`skill/` directory is a pointer to it so no divergent copies exist.
`src/webhook.ts` ships as a reusable library affordance for webhook-driven
wait loops (tests in `test/calle.test.ts`); the shipped flows use terminal
polling and need no tunnel.

## Verification

`python3 scripts/validate_repository.py` passes on this branch.
`npm test` runs 559 tests across 14 files with no credentials and no calls:
an adversarial renderer suite (139 tests, including poison cases and
fast-check property tests for the taint scan), crash-resume tests proving a
torn write heals to the same terms digest as the uninterrupted path, and
sqlite ledger tamper-detection tests — including two that document what
verification does *not* catch. The rehearsal runner
(`npx tsx scripts/run-live-case.ts --rehearse --yes`) drives a full case
through the production pipeline against scripted personas.

## Honest limits

Attestation proves the person who answered the party's phone heard the exact
recorded terms — not who that person was, and not that ASR heard the speaker
perfectly (on the live case, a spoken "keys" was captured as "kids"; the
system records terms as heard, and dual attestation guarantees both parties
heard the same recorded terms). The memorandum is non-binding and nothing
here is legal advice. The ledger is tamper-evident, not tamper-proof: it has
no external anchor, and a refused attestation attempt leaves no case-ledger
entry — that evidence lives only in CALL-E's call logs. Call-window,
cooling-off, and retry pacing are recorded policy the caller must honor;
only the round limit and TTL are enforced by the state machine. The MCP
surface does not yet expose the state machine's cancel transition; stopping
the step loop, TTL expiry, and party exit are the available stops.


Ray-56:
Current Must Fix: seven-digit destinations pass the E.164 gate; timeout or connection ambiguity is retried with a fresh idempotency key, so an accepted call can be duplicated; and completed schema-shaped output can advance financial consent and offers without task completion, exact call, recipient, and metadata binding, or transcript evidence. Please fail closed and reconcile the same call and key instead of redialing.

Ray-56:

Remaining Must Fix only:

- Timeout/connection ambiguity retries with fresh key, duplicating accepted calls.
- Unbound schema output advances financial consent/offers; require human approval or bind valid terminal evidence.
- Stop timeout/connection retries with fresh keys and prevent invalid live destinations. Caucus retries ambiguous calls with fresh idempotency key and permits seven-digit live destinations.
- Do not advance binding financial consent/offers from incomplete unbound results. Current state machine consumes schema-shaped completed output for offers without required outcome/recipient confirmation.

PR #171, opened 2026-08-16
feat(apps): add RevisitZero field-visit recovery app

zhangshen20:
## What changed

- Add the RevisitZero TypeScript app under `apps/typescript/revisit-zero`.
- Add a desktop operator workbench for three failed field-visit cases.
- Enforce deterministic policy gates, exact content-bound approval, one-recipient/one-attempt CALL-E execution, strict structured-result validation, duplicate prevention, and human-approved local export.
- Include a credential-free fake/no-call demo, fictional sample data, tests, setup, and safety documentation.
- List RevisitZero in the repository and apps indexes.

## Why

Failed field visits often need a small, structured access clarification before a human can decide whether to rebook. RevisitZero demonstrates how CALL-E can support that narrow workflow without automatic booking, CRM updates, customer notifications, broad outreach, or automatic retries.

## User and developer impact

Operators can review an eligible case, approve the exact bounded contact task, validate the structured outcome, and export a recommendation packet only after a separate human decision. Ineligible or contradictory cases fail closed and route to manual review.

The public demo remains **FAKE / NO-CALL** by default. Controlled live mode requires server-side credentials, an explicitly consenting test recipient, a short active call window, and fresh approval.

## Validation

- `npm test` — 63 tests passed
- `npm run check`
- `npm run build`
- `git diff --check`
- Dependency audit — 0 vulnerabilities

## Demo

- YouTube: https://youtu.be/zVT_YnxbNdQ
- Devpost: https://devpost.com/software/revisitzero


Ray-56:
Current Must Fix before marking ready: the live POST endpoint has no authentication or authorization and lets a caller self-assert the approval fields used to trigger the call. The provider idempotency key is also based only on case ID rather than the exact approved payload. Please authenticate and authorize live dispatch server-side and bind the provider key to the normalized recipient, task, locale, schema, and approval payload.

zhangshen20:
Thanks for the review — addressed in c4271de.

- Live dispatch now requires a server-configured Bearer credential, checked with a timing-safe comparison before the request body is processed.
- The operator identity and live-approval gate are derived server-side. The client no longer supplies approvedBy or explicitLiveApproval, and the server accepts an approval only when its digest matches the current exact preview.
- The provider idempotency key now hashes a canonical binding containing the normalized recipient, generated task, locale, closed result schema, and server-attributed approval receipt. The live adapter rejects a key that does not match that exact binding before creating a call.

Validation completed: 71 tests passed, strict TypeScript check passed, production build passed, fake/no-call demo passed, repository validation passed, and the dependency audit reported 0 vulnerabilities.

PR #172, opened 2026-08-16
feat(callback-scam-screener): add callback-scam screening app

kaddison-nexa:
## Summary

Adds `apps/python/callback-scam-screener/`: CALL-E dials a suspicious "call this number" phone number from a scam email *before* a human does, has a transparent identity-verification conversation, and scores the transcript against a tiered scam-signal checklist. The agent has nothing real to leak by construction (no account data, no tools to install anything or move money) rather than by prompt instruction alone, and it never decides the verdict itself — scoring happens afterward, deterministically, from the transcript.

Positioning note: `apps/typescript/verify-contact-claim` already covers an adjacent case (verifying contact via the number *printed on the customer's own card*). This app answers a different question — not "did contact happen" but "does whoever picks up this specific number behave like a scammer" — see the README's differentiation section.

Validated against real usage, not just the mock path: a real CALL-E call (plan → run → poll → transcript), real LLM tagging via Gemini's Interactions API, redaction, and guardrails all ran together successfully. Also validated with a text-based persona test harness before spending real call budget, which surfaced a real limitation now documented plainly in `docs/CONCEPT.md` rather than hidden: a sufficiently patient, evasive caller can currently score as `likely_legitimate` instead of `inconclusive`.

Anthropic is implemented (`pipeline/llm_providers.py`) but not registered as a selectable `--llm-provider` yet — only Gemini has an actual real-call verification behind it so far.

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

- [x] Recurring workflows include cancellation behavior. (N/A — no recurring workflow; one call per run, documented in README's Cancellation and rollback section.)
- [x] Runnable code has a dry-run, fake-server, or no-call path by default. (`screen.py --demo`, and preview is the default for `screen.py` itself.)


Ray-56:
Current Must Fix: the live workflow can call an extracted suspicious number without verified recipient authorization because the allowlist defaults unrestricted, and it does not enforce strict E.164 or mask full numbers and transcripts. It supplies no stable payload-bound provider idempotency; timeout retry includes call run, while local suppression is recorded only after success, so an ambiguous accepted call can be duplicated. Failed, canceled, no-answer, or empty-transcript outcomes can also be scored as likely legitimate without task, call, recipient, and evidence binding. Please fail closed on these live and result boundaries.

kaddison-nexa:
Thanks for the review — pushed fixes for all of this across `97e0bd6` and `4cb8102`:

- **Allowlist defaulted unrestricted** → `CallGuardrails` now requires either an explicit `allowed_numbers` set or `unrestricted=True`; there's no silent "`None` means unrestricted" path anymore. `screen.py --live` requires `--allow-number` or `--unrestricted`.
- **No strict E.164 / no masking** → `--to-phone` is now validated against strict E.164 before dialing, and the app now dials that human-confirmed value rather than the loosely-formatted text the extraction regex pulled from the email body. Output is masked by default (`call_metadata.number_dialed` and any occurrence in the transcript); `--show-full-number` opts back into the full value. Extended this to the error/refusal paths too (guardrail violations, CLI timeout messages) so a raw number can't leak into stderr either.
- **No idempotency, retry could duplicate a call** → `call run` (the request that actually dials) is no longer retried on a client-side timeout — it raises a distinct `AmbiguousCallOutcome` (new exit code 53) so a human checks `calle call status` rather than the client guessing. `call plan`/`call status` (no side effects) still retry as before.
- **Local suppression recorded only after success** → `CallGuardrails.record_attempt()` now fires before the call is placed, not after it succeeds, so a crash or a re-run after an ambiguous outcome is blocked too.
- **Failed/canceled/no-answer/empty-transcript scored as likely_legitimate** → `score()` now requires a `COMPLETED` status and a non-empty transcript before evaluating any signals; anything short of that returns `inconclusive` with an explicit warning instead of silently falling through on zero tagged signals. Also added a recipient-binding check in the orchestrator so a verdict is never produced against a mismatched call result.

New/updated tests across `test_guardrails.py`, `test_scoring.py`, `test_caller.py`, and `test_orchestrator.py` — 78 passing. Let me know if any of this doesn't match what you had in mind.

Ray-56:
Current Must Fix items at head d12f6b6:

- Live mode can still dial an arbitrary extracted destination through --unrestricted, and recipient matching/allowlisting compares only the last ten digits, so different-country E.164 numbers can alias. Require an exact authorized E.164 destination.
- Preview and some task/error paths still expose the full destination. Mask it by default everywhere.
- The public PR/fork history still contains a plausible non-reserved phone fixture from the earlier privacy incident. Remove it from accessible public history where practical; please do not repost the value.

The no-retry-on-run-timeout, pre-dial suppression, terminal evidence gate, and recipient-result binding are fixed. Repository validation, diff check, and 78 credential-free tests passed; no calls were placed.

kaddison-nexa:
This PR can no longer be reopened — GitHub auto-closed it when a history rewrite (to scrub a privacy-incident test fixture, see above) unintentionally severed this branch's shared ancestry with `main`. That's fixed now, but the closed state itself is stuck (confirmed via both the API and the UI's disabled Reopen button).

Continuing this work in a new PR from the same branch: #211. All the feedback above is already addressed there.

PR #175, opened 2026-08-16
feat(quotewake): add Salesforce-first quote follow-up app

josescal:
## Summary

- Add QuoteWake, a Salesforce-first application for following up commercial quotes by phone through CALL-E.
- Add Salesforce REST/SOQL integration, quote selection policies, CALL-E execution and structured outcome persistence.
- Add Salesforce metadata, setup scripts, runtime user configuration, documentation, demo media and automated tests.

## User impact

QuoteWake can identify eligible Salesforce Quotes, resolve the related customer contact, run a dry-run or CALL-E outbound call, and write the structured result back to Salesforce as Quote fields and a Task activity.

## Validation

- `uv run pytest` - 262 passed, 30 subtests passed.
- `uv lock --check`
- `git diff --check`
- `bash -n scripts/*.sh`

Portability regressions cover Bash 3 nounset-safe optional arrays and Python standard-library UTC date handling; the shell scripts no longer depend on GNU `date` syntax.

## Notes

The PR is opened as a draft so the implementation and Salesforce integration can be reviewed before merge.


Ray-56:
Current Must Fix before app-level review: this branch destructively rewrites the repository root AGENTS.md and removes required repository rules, causing the repository validator to fail; restore the root file and keep QuoteWake instructions app-local. The claimed validation also currently has 7 failing tests. In the live path, the provider key is not bound to the exact phone, task, and schema payload, and accepted results are written to Salesforce without exact call, task, recipient, or metadata binding. Please rebase and scope the branch, restore validation, and close those live-result boundaries.

josescal:
Implemented the review remediation and rebased `feat/quotewake` onto the current upstream base.

- Restored the repository-level `AGENTS.md`; QuoteWake guidance remains app-local.
- Bound live CALL-E idempotency keys to the exact Salesforce operation and provider payload.
- Added fail-closed verification for call ID, recipient, task, schema, metadata, Quote/Contact identity, and provider key before Salesforce writes.
- Added negative coverage for mismatched evidence.
- Validation: `python3 scripts/validate_repository.py` passes; `uv run pytest` reports 256 passed.

The branch head is `0f4f451c83d13239cfcc6d88beea549026ca0cff`.

Ray-56:
Must Fix before this draft is ready:

- The thumbnail and teaser video expose plausible personal contact identifiers and a provider call/log identifier. Replace or redact the media and purge those artifacts from every unique PR commit; do not repost the exposed values.
- The advertised validation is still not reproducible on the documented macOS environment: 7 tests fail because the scripts rely on Bash 4-only behavior. Make the scripts portable or correct the supported environment, then rerun the suite.

Ray-56:
Must Fix before this draft is ready:

- The seed script still contains plausible non-reserved full phone numbers paired with named demo accounts. Replace them with standards-reserved fictional values and purge those values from every unique PR commit; do not repost them in review discussion.
- Validation is still not reproducible on the documented macOS environment: 6 tests fail because empty arrays break under Bash 3 with set -u and the retry-date path uses GNU-only date syntax. Make these scripts portable, rerun the suite, and update the PR’s validation claim.

Ray-56:
Must Fix before this draft is ready:

- The seed script still contains a plausible non-reserved full phone paired with named demo accounts. Replace it with a standards-reserved fictional value and purge it from every unique PR commit; do not repost it in discussion.
- Validation is still not reproducible on the documented macOS environment: 6 tests fail because empty arrays break under Bash 3 with set -u and the retry-date path uses GNU-only date syntax. Make these scripts portable, rerun the suite, and update the PR's validation claim.

Ray-56:
Must Fix:

- Plausible non-reserved full phone values remain in current test fixtures and unique PR history. Replace them with standards-reserved fictional values and rewrite the PR history before merge.

josescal:
Addressed in the rewritten PR history at `17da463`:

- Replaced plausible full phone fixtures, including formatted and dynamically constructed variants, with standards-reserved fictional values.
- Rebuilt and audited all 21 commits unique to the PR; the audit reports zero non-reserved phone occurrences.
- Revalidated with 262 passing tests, `uv lock --check --offline`, `bash -n scripts/*.sh`, `git diff --check`, and the repository validator.

No prior phone values are repeated here.

PR #176, opened 2026-08-17
feat(apps): add leash credential-lease supervisor

Oishi1029:
## Summary

The apps in this repository that gate a consequential action share one shape: place a call, take a decision, hand it to a person to carry out. LEASH runs the other direction. The call has no power to hand the agent anything. It can only take something away.

An unattended agent runs holding a Google OAuth credential **on a lease**. When the lease expires, a supervisor places one real phone call through CALL-E to the account owner. On that call the owner has exactly one power: end the lease. Ending it revokes the credential at Google, and the agent cannot mint a replacement — that requires a human sitting at a browser.

Hearing "continue" hands the agent nothing it did not already have. It is not a release; it is the absence of one, and the supervisor lets the lease stand only when **twelve independent conditions hold at once**:

1. `reached_terminal`
2. `status_completed`
3. `task_completed_true`
4. `confidence_at_or_above_threshold`
5. `confidence_label_not_low`
6. `structured_result_present`
7. `decision_is_continue`
8. `readback_confirmed`
9. `spoke_with_person`
10. `live_human_evidence_in_transcript`
11. `evidence_supports_decision`
12. `reason_does_not_contradict_decision`

"stop" needs one. So does an answering machine picking up, a structured result that disagrees with its own transcript, a result that cannot be read at all, a call that never reaches a terminal status, and the supervisor failing before it can finish evaluating. In a gate, a failed call is a no-op — the pending action simply waits. Here a failed call is the loudest outcome in the system: the credential dies.

This repository already contains a merged phone gate for CI/CD deploys (#41). LEASH is not a variant of it. There is no blocked action waiting on this call, because the agent is already running and already holding the credential. The only thing the call can settle is whether it keeps holding it.

Ending a lease is also not deleting a local credentials file and not flipping a flag inside our own process. It is `POST https://oauth2.googleapis.com/revoke` with the refresh token. Afterwards Google's token endpoint answers `400 invalid_grant` to the same refresh exchange that returned `200` minutes earlier. That was the outcome on every run we performed, and the gap between the revoke returning and the first `400` measured 0.09–0.14 s.

The proof forces a refresh exchange on every run, because Google access tokens live about 3600 s and the two runs sit roughly 180 s apart; a cached access token would return `200` both times and prove nothing. The proof queries the token endpoint and never Drive: Drive's front end kept honouring a token we had already revoked, for an interval we could not predict, so a Drive call is not evidence that a credential is dead. `401 invalid_client` means our own configuration is broken, not that a revocation happened, and the code detects and labels that case separately.

### The two content-screen refusals, and the guards they became

CALL-E screens `task` text at create time and can refuse with `HTTP 422 call_not_ready`. Two earlier drafts of this app's call script were refused live:

- **A.** A framing that described an in-progress hazard and positioned the call as the response to it. Refused, in part verbatim: *"revise the request so it is clearly non-emergency and does not rely on this call for urgent safety response."*
- **B.** A draft that read a confirmation code aloud, asked the caller to repeat it, and asked whether to keep or release an access credential. Refused verbatim: *"I can't place a call that involves confirmation-code readback or decisions about keeping or releasing an access credential. Please revise the call goal to remove any verification-code, OTP, password, PIN, or credential-access steps."*

Both refusals are correct platform behaviour. Draft B was, structurally, an OTP-phishing call, and refusing it was the right call. Both are now shipped guards:

- A fixed allowlisted template with exactly two regex-validated slots (`JOB_ID`, `MINUTES`). No caller-supplied and no model-supplied prose reaches `task`.
- `assert_task_is_clean()`, which scans the **rendered** string — not the template — against the registers named in the two refusals plus the categories the CALL-E Terms of Service prohibit, and raises before any request is sent, so a bad string costs nothing.

The accepted wording is pinned by SHA-256 `4e971382408307404e8938186b01f2d50f98dac2abb56cee0eeade1c2b7dfce8`. The screen is undocumented and unversioned, so an edit surfaces only as a refusal at create time and there is no run-time recovery from one; the pin exists so that a change to the wording is a deliberate, visible edit rather than an incidental one.

**The accepted call script never mentions a credential, a token, or any code the caller is asked to speak or repeat.** It asks one question about a background job. To be exact rather than sweeping: the word "code" does occur once in the frozen script, in the phrase "its own practice code repository", and nowhere else; the rendered task contains no occurrence of *credential*, *token*, *password*, *PIN*, *OTP*, or any of the code-readback phrasings the refusal named, and a test asserts that by running the shipped scan over the rendered string. Revocation is the supervisor's own response to "stop". It happens on our side of the wire, and CALL-E is never asked to handle, read, or decide about a credential.

### Platform behaviour the design is built on

- `CallStatus` is exactly `queued|in_progress|completed|failed|canceled`. There is no `no_answer`, `voicemail`, `busy` or `declined`. A no-answer arrives as `failed` with a **free-form** `failure_code` — no enum, so the supervisor never switches on it.
- **A voicemail can arrive as `completed`.** That is why condition ten counts user turns and characters of actual speech instead of trusting the status.
- Extraction failure is silent and total: `structured_result` becomes `null` for the whole object, while `task_completed`, `completion_confidence`, `evidence` and `transcript_turns` survive. Condition six exists for that case: with `structured_result` null there is no decision to read, so the lease is released even though the surviving fields still look healthy.
- On the three LEASH calls whose full records were retained, the per-recipient `structured_result` was `null` while the top-level `structured_result` on the call object carried the extraction. LEASH reads the top-level field. A reader who assumes the recipient-level field is populated will see a `null` that is not an extraction failure.
- Webhooks are unsigned. The docs prescribe re-fetching `GET /v1/calls/{id}` before a sensitive side effect, and the supervisor does exactly that before it revokes anything.
- `result_schema` is validated before `recipients`, which makes a zero-credit schema pre-flight possible by pairing a candidate schema with the un-dialable number `+1`: `result_schema_invalid` means the schema is wrong, `invalid_phone` means the schema is fine. That is what `python3 -m leash preflight` does. It does **not** tell you anything about the content screen — only a create that reaches `queued` does that.

### How it is run

Default-safe. The default command places no call and needs no API key.

- `python3 -m leash demo` — runs the whole supervisor against a bundled local fake CALL-E server. No API key, no credits, no phone call. 16 scenarios: `continue_clean`, `stop_plain`, `no_answer`, `voicemail_as_completed`, `null_extraction`, `low_confidence`, `contradiction`, `evidence_absent`, `readback_denied`, `unclear_answer`, `never_terminal`, `canceled`, `refused_at_create`, `schema_invalid`, `insufficient_balance`, `create_ambiguous`.
- `python3 -m leash preflight` — schema pre-flight against the real API. Needs an API key; spends no credit and places no call.
- `python3 -m leash live` — places one real call. Requires four explicit flags, including `--i-understand-this-places-a-real-call`.
- `python3 -m leash prove --lease PATH` — the on-camera proof command. It places a real call and, on "stop", performs the revocation and the before/after token exchange.

Exit codes: `0` lease continues, `2` lease released, `3` operator error.

Layout: `leash/{templates,outcomes,policy,supervisor,revoke,fakecalle,__main__}.py`, `tests/{test_templates,test_policy}.py`, plus the app README and the README entry the contribution guide asks for. 70 tests, run with pytest. One of them asserts that the condition list has exactly twelve entries, so silently dropping a condition fails the suite rather than quietly loosening the lease.

Numbers in examples and docs are fictional (`+1 555 01xx`), apart from the deliberately un-dialable `+1` that the pre-flight uses on purpose. All live calls were placed to a number the author owns, `+60*******22`, region MY. No third party was called at any point.

*This body replaces the draft-stage description posted when the branch was a scaffold. What has and has not been exercised against the real API is stated below.*

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

- [x] **All content is in English.** Code, comments, docs, CLI output and the call script.
- [x] **Branch, commit and PR naming follow the repository conventions.** Branch matches `<type>/<short-kebab-summary>`; `python3 scripts/check_branch_name.py` passes.
- [x] **No secrets and no private numbers are committed.** The CALL-E API key is read from the environment and is never written to disk or logged; `client_secret.json` and `token.json` are gitignored and were never staged. The author's own number appears in the tree only masked, as `+60*******22` — including in the call records quoted in this PR.
- [x] **Real-world side effects are described.** `leash live` and `leash prove` each place one real outbound phone call and spend CALL-E credit. A `stop` outcome revokes a real Google OAuth refresh token at `oauth2.googleapis.com/revoke`; that is irreversible from the app's side and can only be undone by a human completing a fresh browser sign-in flow. Both side effects are documented in the app README and in `--help`, and neither can be reached from the default command.
- [x] **Phone numbers are masked or fictional.** Examples use `+1 555 01xx`. Real numbers appear only masked.
- [x] **Cancellation behaviour is documented.** Stated plainly: **a lease is single-shot, not recurring.** There is no scheduler, no cron entry and no repeat dial. One lease expiry produces at most one call; the supervisor then exits and does not run again on its own. Ending a lease revokes the credential and exits `2`; it is not a pause, and the app has no resume path. If the supervisor is interrupted before the call reaches a terminal status, the lease is released rather than extended. The create carries an `Idempotency-Key`; a re-run that produces the same key returns the stored call record instead of dialling again.
- [x] **Runnable code has a no-call path by default.** `python3 -m leash demo` runs the whole supervisor against a bundled fake CALL-E server with no API key and no network. Live calling is opt-in behind explicit flags.
- [x] **`python3 scripts/validate_repository.py` passes.**

## Validation

Run from the repository root:

```
Repository validation passed.
```

Run from `apps/python/leash/`:

```
$ pytest
70 passed
```

One caveat on the first command, stated so it is not read as more than it is: `validate_repository.py` also passes on an unmodified upstream checkout. A green run here is therefore a **comparison** — it says this PR did not break anything the validator checks — not an assertion that the app is correct. The 70 tests are the substantive claim, and they cover the twelve-condition policy table, the template guards, and the outcome parsing. They do not touch the network.

## Live verification

All calls were placed through the real CALL-E API to a number the author owns (`+60*******22`, region MY). No third party was ever dialled. The author was the person on every call.

- **`call_6ErYKmEI472cBVdsUd7K3Q`** — 2026-08-04. `completed`, `task_completed: true`, confidence 0.92 (`label: high`), decision `continue_job`. First proof that the frozen task text passes the content screen. It also produced the observation behind condition twelve: **the caller's stated reason contradicted the choice just confirmed.** The transcript has "continue" twice and a "yes" to the confirmation, and then `reason_sentence` comes back as *"the jobs done enough, take it back."* — verbatim from the record, transcription artefacts included — which is a reason that means stop. Extraction was faithful. The human was inconsistent. A system reading only the decision enum would have left a live credential in place against the caller's actual intent.
- **`call_RV8aiijpyuySJAermpuFxA`** — 2026-08-17. `completed`, `task_completed: true`, confidence 0.88, decision `stop_job`. **Speech recognition misheard the decision word.** The user turn recorded at the decision point is *"dot."* The agent's read-back, two turns later, is *"You said stop. Is that correct?"*, the turn after it is *"yes."*, and the structured result is `stop_job`. What the record can prove is that the transcript's decision turn is not a valid answer and the pipeline did not take it at face value; what sound actually left the caller's mouth is not something a transcript can settle. This was not arranged, and the take was kept rather than re-recorded. The same call also showed the agent reordering the script — it asked for the reason before the read-back, where the template asks for them the other way round. Harmless here, but the app does not assume turn order.
- **`call_Kw_lOkxZKiO15-I3_m8-vQ`** — 2026-08-17, the recorded demonstration. `completed`, `task_completed: true`, confidence 0.95, decision `stop_job`, created `2026-08-17T02:31:42Z`, completed `02:34:59Z`. Ten of twelve conditions held. The two that failed are the two that can only hold for a continue: `decision_is_continue` and `evidence_supports_decision`. The supervisor re-fetched the call, revoked the credential, and Google's token endpoint then answered `400 invalid_grant` to the refresh exchange that had returned `200` before the call.
- **`call_PR9Y0XYKnArSc3vnV4aZ6A`** — 2026-08-04, from the predecessor concept's task text, not the LEASH template. **A structured result contradicted its own transcript and its own evidence at high confidence.** `task_completed: true`, confidence 0.93; the caller answered "yes" to the acknowledgement question; `evidence[]` said they acknowledged; the structured enum came back `"no"`. Anything gating a real side effect on one structured field would have taken the wrong branch silently. Conditions eleven and twelve are the direct response.

One further observation from a development run against the real API shaped the design, and no call id is on record for it, so it cannot be checked the way the four above can. It is described exactly as observed: **the duplicate-dial guard fired.** A re-run that produced the same `Idempotency-Key` returned the stored call record rather than dialling the human a second time. Observed once, not theorised, and not re-run since.

### What was not tested

Stated explicitly, because the sections above are narrow:

- **One number, one region, one carrier, one handset, one language.** Region MY, `en-US` locale, a handful of calls, all to the author. Nothing here supports a claim about speech recognition accuracy in general: one mishearing was observed across those calls, and that is the whole sample.
- **The failure scenarios are exercised only against the bundled fake server.** `no_answer`, `voicemail_as_completed`, `null_extraction`, `never_terminal`, `canceled`, `insufficient_balance` and `schema_invalid` are modelled on documented and observed API shapes. A live no-answer and a live voicemail-arriving-as-`completed` were **not** reproduced against the real API. The fake server is our reading of the contract, not the contract.
- **The content screen is undocumented and unversioned.** The SHA pin records that this exact wording was accepted on 2026-08-04 and accepted again, unchanged, on 2026-08-17. It is not evidence that the same wording will be accepted later. A tightened screen would surface as a `422` at create time, and there is no run-time recovery from that.
- **Revocation timing was measured on one Google project with one Desktop client**, `drive.file` as the only scope, at 0.09–0.14 s across our runs. We make no claim about propagation to other Google front ends — Drive demonstrably kept honouring a revoked token, which is precisely why the proof queries the token endpoint.
- **A killed process cannot revoke anything.** The policy is written so that an outcome the supervisor cannot positively verify is a release rather than a continue, and an interrupted run releases rather than extends. That is a statement about the code paths the supervisor reaches, not a promise that a `SIGKILL`ed process completes an outbound HTTPS request. Nothing can promise that, and this app does not.
- **No concurrency, no load, no multi-lease.** One supervisor, one lease, one call. Long-lived operation, restarts and clock skew are untested.
- **The bundled fake server has no security posture.** It binds locally for demo purposes and is not intended to be exposed.

## Recorded demonstration

**https://youtu.be/T5v3rSblLro** (2 min 17 s)

A single continuous screen recording of `call_Kw_lOkxZKiO15-I3_m8-vQ`: the token request returning `200`, the call being placed and answered, the twelve-condition table printing its two failures, the revocation, and the same token request returning `400 invalid_grant` — with the same credential fingerprint on screen in both.

Two things about the recording, stated so nobody has to ask:

- The two waiting periods are shown at 8x and labelled on screen as such. Everything else, including the whole call, is at normal speed, and the take is one unbroken recording with no cuts.
- The audio is the real call. The phone number is mine and is masked in the terminal output.

---

Developed solo with AI-assisted tooling (Claude) for code generation, debugging, and documentation; all design decisions, CALL-E integration, and testing are my own original work.


Ray-56:
Current Must Fix before marking ready: the runnable app has no README, so setup, destructive credential-revocation side effects, credential handling, cancellation or rollback limits, and readiness are undocumented. The live and preflight commands accept an arbitrary base URL while sending the CALL-E credential, and the phone validator accepts seven-digit destinations rather than strict E.164. Please add the required app documentation, pin the official HTTPS origin, and enforce the live phone-number floor.

Oishi1029:
Thank you — two of these were real and neither had a test. Both are fixed in `34e9099`.

**Arbitrary base URL while sending the CALL-E credential.** You are right, and it was worse than a
missing check: `--base-url` was an operator-supplied string on both `live` and `preflight`, so
anything able to influence that value — a flag, an environment variable, a wrapper script — sent a
live bearer key to a host of someone else's choosing, and the request still looked ordinary in a
log.

`CALLE_ORIGIN` is now an allowlist enforced in `Supervisor.__init__`, so an untrusted origin is a
construction-time error rather than a request-time one, and `live` and `preflight` no longer accept
a base-URL option at all. The offline `demo` is the only caller that may reach a local fake server;
it has to ask explicitly with `allow_loopback=True`, and the key it carries is a placeholder. A
plaintext downgrade of the real host is refused along with everything else.

**Seven-digit destinations.** Also right, and the code disagreed with its own documentation: the
comment above the pattern said "8-15 digits total" while the pattern read `{6,14}`. Seven digits is
shorter than any real international number and is the shape a typo takes rather than a subscriber,
and the cost of being wrong is a call to a stranger. Now `{7,14}`.

`tests/test_supervisor_guards.py` covers both: suffix hosts (`api.heycall-e.com.evil.test`),
the real host in a path rather than an origin, a hyphen-for-dot homograph, plaintext downgrade,
loopback being refused unless asked for and not widening to remote hosts when it is, key redaction
in `__repr__`, and the digit bounds asserted across the whole range so the pattern and the sentence
describing it cannot drift apart again. 101 tests pass.

**On the README** — it exists, and I think it landed after you looked: your comment is timestamped
`02:25:25Z` and the docs commit is `6e1f614` at `03:00:43Z`, about thirty-five minutes later. It
covers setup, the destructive revocation side effect, cancellation and rollback limits, and
readiness. I have added a **Credential handling** section to it in this change, covering the pinned
origin, the redaction guarantees, and the E.164 floor. If it still does not answer what you were
looking for, tell me which part and I will fix it rather than guess.

Two other things worth flagging, since both bear on how you would review this:

- `run_demo.sh` runs the whole supervisor against a bundled fake CALL-E server across five
  scenarios. It reads no API key, makes no network call and places no phone call, so the app is
  fully exercisable without credentials.
- The recorded demonstration is at https://youtu.be/T5v3rSblLro. The two waiting periods are shown
  at 8x and labelled on screen; everything else, including the call, is a single unbroken recording
  at normal speed.

Ray-56:
Current review: the app now includes the required README, pins the credential-bearing CALL-E client to the approved origin, enforces strict E.164 at dispatch, and documents its deliberate destructive boundary. Repository validation and all 101 tests pass credential-free; no Must Fix items remain under the Community/Hackathon policy.

PR #178, opened 2026-08-17
fix(apps): declare only the dependencies leash actually uses

Oishi1029:
## Summary

Follow-up to #176, which is merged. `apps/python/leash/pyproject.toml` declared four dependencies the package never imports: `calle-ai`, `google-auth`, `google-auth-oauthlib` and `requests`.

Every module in `leash/` uses only the standard library. The CALL-E calls and the Google token exchange both go through `urllib.request` directly, and that is deliberate — the raw HTTP is what the recorded demonstration points a camera at, so the request and response are visible rather than wrapped.

Declaring an SDK the app does not use overstated what it depends on. It also obscured the more useful fact, which is now stated in a comment where the dependency list used to be: a reviewer can clone this and run the entire supervisor with nothing installed.

I found this while answering the hackathon feedback survey, which asks which CALL-E interfaces you used. The honest answer is the Developer API and not the SDK, and the manifest disagreed.

## Type

- [ ] New skill
- [ ] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

- [x] Real-world side effects are clearly described. Unchanged by this PR; the app still revokes a Google credential and the README says so.
- [x] Recurring workflows include cancellation behavior. Unchanged; a lease is single-shot.
- [x] Runnable code has a dry-run, fake-server, or no-call path by default. Unchanged, and this PR makes it truer.

### Validation

```
Repository validation passed.

$ python3 -m venv /tmp/clean && /tmp/clean/bin/python -m pip install -e .
$ /tmp/clean/bin/python -c "import leash, leash.supervisor, leash.policy, leash.revoke, leash.fakecalle, leash.templates"
$ /tmp/clean/bin/python -m leash demo --scenario continue_clean
  LEASE CONTINUES

$ python3 -m pytest tests/ -q
101 passed
```

One file changed.


Ray-56:
Current Must Fix under the new privacy/public-history rule: the Leash app currently publishes raw excerpts from real calls together with real call identifiers in its README, reviewer record, policy example, and mirrored fake fixtures. These artifacts are already present on the public default branch, so remove or replace them in the current tree and perform practical history remediation without repeating the values in discussion. The dependency correction itself passes validation; committed leash.egg-info build artifacts are a non-blocking cleanup item.

Oishi1029:
Understood, and thank you for catching it. Fixed in `c9830d7`, and I have not repeated any of the values here.

**What was there.** Material from calls I placed to a phone I own: call identifiers and verbatim caller speech, in the README, the reviewer record, one policy docstring, and fake-server fixtures that mirrored real utterances rather than inventing them. I included it as evidence, which is exactly the instinct the rule exists to correct — a call identifier plus a transcript excerpt is a record of a real conversation regardless of whose phone it was.

**Current tree.**

- **Identifiers removed.** The live-verification records are now labelled A to D. Dates, statuses, confidence scores and decisions are unchanged, so every claim in the documents stays checkable — I can give you the identifiers privately if you ever want to verify one.
- **Verbatim speech removed** from prose, transcript excerpts and structured-result examples. What each utterance demonstrated is described rather than quoted, and the withheld spans are marked as withheld rather than silently deleted.
- **Fixtures no longer mirror real utterances.** The contradiction fixture and the mis-transcription token are invented, and the docstrings now say so, so a reader cannot mistake a fixture for a recording.
- **The stop-leaning matcher keeps its lexical patterns.** Those are detection rules rather than call content, and they no longer cite a call or quote one. Say the word if you would rather they went too and I will replace them with an invented set.
- **`leash.egg-info` untracked**, and build artifacts added to `.gitignore`. That was a local editable install swept in by a careless `git add -A`.

101 tests pass and every fake-server scenario behaves as it did before — the fixtures changed wording, not meaning.

**On history.** The values are in commits already merged to the default branch, so the part I can do from here is limited and I would rather be plain about that than imply otherwise. What I have done: the current tree is clean, my fork's branch has been force-pushed so the working history I control no longer carries them, and the underlying call records live in a private working directory outside this repository, which is where they are staying.

What I cannot do from a fork is rewrite the default branch. If you want those commits purged, the options are yours to pick — a squash or history rewrite on `main`, or reverting and re-landing the app as a single clean commit. I am happy to prepare whichever you prefer, or to close this PR and open a replacement branch built from a clean base if that is simpler for you. Tell me which and I will do the work.

One thing worth flagging for the rule itself rather than for this PR: the repository's contribution guidance already asks for masked numbers and no recordings or private transcripts, but a contributor reading it would not obviously conclude that a call identifier from their own account is covered. If the rule is new, saying so explicitly in `CONTRIBUTING.md` would have stopped me making this mistake in the first place.

Ray-56:
Current Must Fix items at head c9830d7:

- The current README, reviewer record, and fake-server fixtures still include literal excerpts or abridged log/result artifacts explicitly attributed to real calls, and one replacement fixture still describes itself as verbatim. Replace these with clearly synthetic text or summary-only descriptions.
- The original identifiers and excerpts are already on the public default-branch history. After the current tree is fully cleaned, merge the cleanup promptly and complete practical default-branch history remediation; please do not repost any exposed identifier or excerpt.

The dependency correction and generated-artifact cleanup are otherwise sound. Repository validation, diff check, and 101 credential-free tests passed; no calls were placed.

Oishi1029:
Fixed at `3e07267`. You were right that the first pass was not enough, and the reason is worth naming: I redacted values in place but kept the artifacts around them. A transcript block with the caller's turns blanked out is still a transcript from a real call, and the attribution above it was doing as much of the harm as the words.

**Now summary-only, everywhere.**

- The regressions section describes in prose what each call demonstrated and why it produced a particular condition. No transcript, no result payload, no log excerpt, no per-call attribution, no labels standing in for identifiers.
- Live verification is a summary: how many calls reached terminal, the range of confidences, the range of create-to-terminal times, and what each fact establishes. Nothing is tied to an individual call.
- The reviewer record is rewritten the same way. It also no longer reproduces the mis-transcribed token, which I had left in.
- The abridged output in the README now comes from the fake server and says so. Previously it was attributed to the recorded run.
- **Every fixture now declares itself synthetic.** That was the clearest miss: several docstrings advertised the fixture as near verbatim from a real call, which was both a privacy problem and misleading about what the fixture is. Nothing claims to be verbatim now, and nothing is modelled on a specific call's wording.

The underlying records stay in a private working directory outside this repository. If you ever want to check a claim in the README against one, ask and I will send it to you directly.

101 tests pass, every fake-server scenario behaves as before, and repository validation passes. Also corrected a stale test count in the README that I noticed while editing.

**On the default branch.** Understood, and I will not repost anything. The current tree is clean, so merging this clears the tree on `main`. The exposure that remains after that is in the commit history, which I cannot rewrite from a fork.

Whichever of these you prefer, I will do the work and you will not need to write anything:

1. You rewrite or squash the affected commits on `main` directly — fastest, and the only option that fully removes them.
2. I prepare a branch that reverts the original app commit and re-lands the app as a single clean commit, so `main` moves forward without the old tree in its recent history. This does not remove the old objects, but it stops them being reachable from the branch tip.
3. I close this PR and open a replacement built from a clean base, if you would rather review the app fresh rather than as a series of fixes.

Say which and I will start. If option 1, tell me if you want me to hold off pushing anything further to this branch first so the history you rewrite is stable.

Ray-56:
The current-tree privacy blocker is resolved: the real-call excerpts and identifiers have been replaced with summaries and synthetic fixtures, and validation passes. Proceeding to merge this cleanup. The inherited artifacts remain reachable in the public default-branch history and still require separate maintainer-led history remediation.

PR #194, opened 2026-08-19
fix(apps): local-atlas verifies OSM listings too, against the OSM API

omarmiri:
Follow-up to #156, fixing a gap introduced by the recipient-verification change in that PR.

## What's wrong today

#156 made a live call dial only a number the server can read back from the listing, rather than one supplied in the request body. That check keys on a **provider id** — so it covers Google and Foursquare listings and silently excludes every OpenStreetMap one, which cannot be verified and is therefore refused.

That is not a rounding error. Measured with the app's own category query across five areas:

| Area | named OSM places | with a dialable phone | share |
|---|---|---|---|
| Manhattan NY | 779 | 359 | 46.1% |
| Iselin NJ (suburban) | 355 | 133 | 37.5% |
| rural Vermont | 127 | 46 | 36.2% |
| Toronto ON | 767 | 196 | 25.6% |
| Boise ID | 467 | 111 | 23.8% |
| **Total** | **2,495** | **845** | **33.9%** |

About a third of listings, and the share holds up well outside dense cities.

## Why not Overpass

The obvious fix is to ask Overpass for the one element. It cannot be done from the deploy — measured per mirror, for a **single-element** query:

```
overpass-api.de           fetch failed
overpass.kumi.systems     HTTP 429
overpass.private.coffee   timed out
```

The limit is on the address, not the size of the request. That is also why the map fetches OSM data from the user's own browser rather than through the server — and therefore why signing what the server serves, which was the first approach tried here, would have covered almost nothing. It was removed rather than shipped.

## What this does instead

`api.openstreetmap.org` is a different service, and the authoritative one that Overpass is derived from. It serves a single element by id, and it answers from this host:

```
osm lookup : OK — read one element in 503ms (Hard Rock Cafe)
```

Cross-checked against four Manhattan restaurants — the OSM API returns the same numbers Overpass does, tag for tag. So listings now carry their OSM element id, and a live call reads the number back from there before dialling it: one object, only when somebody actually asks a question, cached for a day, with a User-Agent identifying the caller.

The rule is unchanged and now covers every source: **dial the number the source says, not the one in the request body.** A live call with nothing to check against is still refused, and a multi-place round still drops the recipients it cannot confirm.

## Scope

- Live calls only. The refusal is gated on `live && !isOwnLine`, so simulated calls — which is everything under `REVIEW_MODE=1 CALLE_DRY_RUN=1` — are untouched, and the review path needs no keys of any kind.
- No new dependency, no new configuration, no stored secret.
- 4 files, +60/−6.

`python scripts/validate_repository.py` passes.


Ray-56:
Must Fix before merge:

- The OSM lookup verifies the phone for a caller-supplied `osmId`, but the task and stored result still use independently caller-supplied place identity, and `placeKey()` ignores `osmId`. A mixed or stale request can therefore call OSM listing A while announcing and publishing the answer as listing B. Bind one authoritative source identity end to end—for example, reject mixed provider IDs, key OSM records by `osmId`, and use identity returned by the OSM lookup (or an equivalently verified mapping) for the task and persisted result.

omarmiri:
Thanks — fixed in `9ad2504`.

- `providerRef()` / `placeKey()` now key on `osmId`, so an OSM record is filed under the id its number was verified against instead of falling through to a name-and-coordinate key.
- A body naming more than one provider is refused outright on both dial paths (`askPlace` returns 400; `askAround` drops that recipient), and gets a key of its own on the read paths so it cannot reach a single-source record.
- `listedPhone()` is now `listedListing()` and returns the name and address beside the number from the same provider read. `verifiedIdentity()` puts those into the call script, the confirmation preview and the stored record — so the business the agent announces, and the record the answer is filed under, are the business whose line was verified.

The cache key moved from `lp:` to `ll:` along with the return shape, so a warm entry from the string-returning version can't be read back as a listing. Simulation is untouched and still needs no provider keys.

`python scripts/validate_repository.py` passes.

PR #198, opened 2026-08-19
feat(countersignal): add falsifiable customer discovery agent

yangyangnovelist-hub:
Adds CounterSignal, a CALL-E customer-discovery agent that freezes the hypothesis, questions, and decision rule before interviews; preserves disconfirming evidence; excludes voicemail/refusal from the answered denominator; and binds valid evidence to the exact call, protocol, recipient, and transcript quote.

Public judge demo: https://countersignal.vercel.app

Key reliability properties: explicit AI-call permission, exact recipient allowlist, SQLite intent reservation, outcome_unknown/no blind redial, and deterministic 8/5/3 decision logic.

Staging PR and CI are green in the contributor fork.

### DIFF
--- a/apps/python/countersignal/DEMO.md
+++ b/apps/python/countersignal/DEMO.md
@@ -1,0 +1,75 @@
+# CounterSignal judge demo — target 2:30
+
+The demo should prove the product boundary, not narrate the README. Everything below can be shown with the deterministic judge console and no credentials. A real-call clip can replace the marked section later without changing the story.
+
+## 0:00–0:18 — The problem
+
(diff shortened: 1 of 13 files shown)


Ray-56:
Must Fix:\n\n- DEVPOST-AUTOFILL.md contains a plausible personal email in the current tree and multiple PR-unique commits. Replace it with a reserved example-domain address and rewrite the PR history; please do not reproduce the value in follow-up comments.\n- The live path permits CALLE_API_KEY to be sent to a plaintext loopback CALLE_BASE_URL. Restrict credential-bearing live requests to the exact official HTTPS CALL-E API origin; tests can inject a fake client/key instead.

PR #199, opened 2026-08-19
feat(permitdiff): add evidence-bound permit reconciliation agent

yangyangnovelist-hub:
Adds PermitDiff, a CALL-E exception resolver for stale or conflicting permit records. Fresh records are successful no-call outcomes; eligible calls are bound to the exact permit, snapshot hash, destination, call ID, and transcript evidence. Phone evidence can surface a discrepancy but never rewrites official permit state.

Public judge demo: https://permitdiff.vercel.app

Includes applicant-side authorization gates, exact allowlisting, SQLite reservation, outcome_unknown/no blind redial, and deterministic fresh/match/discrepancy reviewer states.

Staging PR and CI are green in the contributor fork.


Ray-56:
Must Fix:\n\n- DEVPOST-AUTOFILL.md contains a plausible personal email in the current tree and multiple PR-unique commits. Replace it with a reserved example-domain address and rewrite the PR history; please do not reproduce the value in follow-up comments.\n- The live path permits CALLE_API_KEY to be sent to a plaintext loopback CALLE_BASE_URL. Restrict credential-bearing live requests to the exact official HTTPS CALL-E API origin; tests can inject a fake client/key instead.

PR #203, opened 2026-08-20
feat(apps): add VendorDesk autonomous procurement agent

sadbunny3000:
Adds VendorDesk, an autonomous procurement agent for CALL-E, given a sourcing job, it places an outbound call, asks about pricing/stock, and records the result as a structured quote.

**What's included**
- Next.js frontend (`frontend/`) start a sourcing run and view quotes
- TypeScript backend (`src/`)  CALL-E client, webhook handler, quote storage

**Compatibility**
- Backend on port 3001, frontend dev server on port 3002 (frontend proxies `/api/*` to the backend)
- Self-contained under `apps/typescript/vendor-desk/ no changes to other apps in the repo

**Safety**
Full details (side effects, cancellation, and credential handling) are in `apps/typescript/vendor-desk/README.md`. In short:
- Dry-run by default — no live calls unless explicitly opted in
- No recurring or scheduled jobs
- Required environment variables are listed in `.env.example`; no secrets committed

**Testing**
- `scripts/validate_repository.py` passes
- End-to-end tested against CALL-E's test hotline (not a live vendor)  confirms the full pipeline works; a "failed" result there is expected since the test line has no real pricing data


Ray-56:
Must Fix:\n\n- Add basic authentication/authorization to the remotely reachable dispatch, jobs, SSE, and transcript paths, plus an explicit per-run live confirmation.\n- Require an authorized recipient and strict server-side E.164 validation before dispatch; also cap/confirm the vendor batch rather than accepting arbitrary phone lists.\n- Mask full phone numbers and protect transcripts in persistence, API/SSE responses, and the dashboard.\n- Restrict CALLE_API_KEY to the exact official HTTPS CALL-E API origin.\n- Use a stable authorization/payload-bound provider key and preserve ambiguous create outcomes as unresolved so replay cannot redial with a fresh random key.\n- Authenticate and bind webhook results to the expected call/job/recipient, and validate boolean fields exactly; a forged or string false-like result can currently become a positive top recommendation.\n- Replace the plausible real provider event/call identifiers in the current tree with synthetic values and rewrite both PR-unique commits. Please do not reproduce those identifiers in follow-up comments.

Ray-56:

Remaining Must Fix only:

- Remote dispatch/jobs/SSE/transcript surfaces lack authentication and live per-run intent.
- Recipient validation/authorization and real-phone/transcript output protection missing.
- Credential base URL unrestricted.
- Ambiguous create can replay with fresh random key; halt.
- Webhook can forge boolean results into binding recommendation/action; authenticate/refetch and keep consequential choices human-reviewed.

PR #205, opened 2026-08-20
feat(incident-escalation-call): add on-call escalation skill and app

gmassello:
## Summary

Ringdown pages the on-call engineer and proves the acknowledgement happened.

Every on-call system reports "notification sent" and treats the incident as escalated. That
proves nothing: the push landed on a silenced phone, the page was read at 3am and forgotten. The
acknowledgement is the only part that matters and it is the part nobody verifies.

This adds a skill and the app it drives:

- **`skills/incident-escalation-call/`** walks the ladder and reads the exit code. It manages the
  app rather than reimplementing its logic.
- **`apps/python/ringdown/`** resolves an on-call rotation, places one call per rung with a
  content-derived `Idempotency-Key`, and records an acknowledgement only when an owner and an ETA
  are each quoted by a span the recipient actually spoke. A "yes" with no number of minutes is not
  an acknowledgement, and the run drops to the next rung.

The part that is unusual: **it places the call over REST and re-reads it over MCP.** The two
surfaces are not two views of one document — REST reports lowercase statuses with `task_completed`
and `completion_confidence`, MCP reports uppercase statuses and accepts no extraction schema — so
verifying over MCP re-derives the acknowledgement from a transcript served by the transport that
did not write it. Every verdict is sealed in a hash-chained ledger that **re-derives the verdict
from the recorded attempts** on replay, so rewriting a verdict and resealing the chain still fails
verification.

### On the neighbours

`deployment-approval-call` asks *before* acting, of a known approver, and fails safe. Ringdown asks
*after* something broke, of a rotation that has to be resolved first, and fails unsafe: nobody
answers and the incident keeps running. `verify-by-phone` shares the span grounding but makes one
call to check one published fact. `apps/python/incidentbridge` calls vendor support for evidence;
Ringdown calls your own engineer for a commitment.

### Verified against the live provider

Six calls were placed against the live API on 2026-08-20, through a US number that bridges to a
region CALL-E does not dial. They confirmed the parts this design leans on:

- **`metadata` echoes back**, so an attempt can be tied to the call it belongs to.
- **The idempotency key holds.** All five REST creates timed out before answering, all five
  replays returned the existing call, and nobody was dialled twice. Against the real API,
  reconciliation is the normal path rather than a rare branch.
- **The degradation behaves as designed.** Every live verdict settled at exit 45 — the code that
  means *the second channel said nothing*, never *the second channel disagreed*.

They also settled an open question, in the direction this app did not want. `get_call_run` takes a
`run_id` that only `run_call` hands out; no identifier a REST-placed call exposes resolves to one;
and the first successful run ever observed does not have the shape this app parses. Cross-surface
verification cannot be completed from this side today, so a live run reports the acknowledgement as
unconfirmed instead of claiming it.

That is stated as ceilings 12 and 16 in the app README rather than left for a reviewer to discover.
The *shapes* those calls answer with are recorded in
[`tests/fixtures/`](apps/python/ringdown/tests/fixtures/) with their provenance. No part of a
provider response is reproduced there: every value in every payload — identifiers, phone numbers,
names, timestamps, transcript turns, summaries and confidence scores — was written for this
repository, and a test holds each fixture to saying so. They are parsed by tests that never touch
the fake, including one asserting the parser cannot read the completed run, so the gap is a tested
fact and not a surprise. The same findings went to the CALL-E team as feedback.

## Type

- [x] New skill
- [x] New runnable app

## Checklist

### Notes on the checklist

- **Side effects and cancellation.** `preview` is the default and opens no socket; `run` refuses to
  dial without `--confirm 'place real calls'`. A call already in flight cannot be cancelled — that
  is a platform limit, stated in the README — so what is cancellable is the ladder, and the ladder
  never re-calls.
- **No-call path.** The demo and the whole suite run against a fake CALL-E on loopback: 297 tests,
  no credentials, no outbound calls. `python -m demo.run_local` narrates seven scenarios.
- **Phone numbers.** Examples use the reserved `555-01xx` range; everything written or printed is
  masked, and the fixtures from the live provider are masked too.
- **Trust boundary.** `--base-url` and `--mcp-url` are allowlisted before any client is built,
  because the API key travels on every request.


Ray-56:
Must Fix:

- The committed fixtures and documentation explicitly contain verbatim live REST/MCP responses, real-call transcripts, and raw call/run/recipient/attempt/provider identifiers. Replace them with wholly synthetic fixtures and rewrite this PR history; do not repost any exposed value.
- Remove the arbitrary --allow-host escape hatch and pin each live credential to its exact trusted HTTPS origin. Non-default trusted-host ports/paths and plaintext loopback must not receive real API keys or MCP tokens; use separate throwaway credentials for local fakes.

gmassello:
Both must-fixes are in, force-pushed as the single commit this branch carries.

**Fixtures.** Every call, run, recipient, attempt and provider identifier, every phone number and every name spoken in the transcripts is now fictional — reserved `+1555010xxxx` numbers and the same Alice Okafor the demo already uses — including the identifier the MCP tool quotes back inside its pydantic validation error. Only the shape is preserved, which is what those files are for: `run_from` still cannot read the completed run, and the fixture still proves it. `tests/fixtures/README.md` claimed "committed verbatim" and now says what is true; ceilings 12 and 15 in the app README were reworded to match.

**Pinning.** `--allow-host` and the host allowlist are gone. `--base-url` and `--mcp-url` accept exactly `https://api.heycall-e.com` and `https://seleven-mcp-sg.airudder.com/mcp/openagent_oauth`, compared as whole URLs, so a non-default port, an extra or traversing path, userinfo, plain `http` and lookalike domains are all refused by one rule with no exception flag — nine near misses pinned by a test. The only other target is loopback, and it reads `RINGDOWN_FAKE_API_KEY` / `RINGDOWN_FAKE_MCP_TOKEN` instead: `CALLE_API_KEY` is never read for a local run, asserted by a test that exits 30 with the live key set and the fake one absent.

296 tests pass, `demo.run_local` reproduces `demo/EXPECTED.md`, `scripts/validate_repository.py` passes.

One caveat: the force-push clears the branch, but GitHub keeps the pre-push blobs reachable by SHA through the PR timeline — purging those needs a request to GitHub Support from your side. No credential was ever in those files, so there is nothing to rotate.

Ray-56:
Current blockers:\n\n- The fixtures remain derived field-for-field from real live-call responses and contain transcript/response content. Replace them with wholly synthetic fixtures using reserved fictional data and rewrite this PR's unique history.\n- REST and MCP validation accepts either production endpoint for either channel, allowing credentials to be cross-sent when the flags are swapped. Pin each credential to its own exact endpoint, with loopback allowed only for fake test credentials.

gmassello:
Both blockers are addressed, force-pushed as the single commit this branch carries.

**Fixtures.** `tests/fixtures/` no longer reproduces any part of a provider response. Every value in every payload — identifiers, phone numbers, names, timestamps, transcript turns, provider summaries, evidence lines and confidence scores — is written for this repository; the bot turns are Ringdown's own script text, the phone numbers are reserved `+1555010xxxx`, and the engineer is the fictional Alice Okafor the demo already pages. What the files carry is the shape: which keys exist, where they sit, what type they hold, and what the endpoint returns when it fails. `tests/fixtures/README.md` now says exactly that instead of claiming a field-for-field transcription, and each fixture's `source` states it too — held there by a test, so a raw response pasted in later fails the suite rather than passing review.

The findings the directory exists for survive the rewrite unchanged, because they were always facts about shape: `run_from` still cannot read the completed run (call id under `result`, `transcript` a single joined string, no `recipient_phone`/`completed_at`/`metadata`), `get_call_run` still rejects `call_id` inside an HTTP 200, and the REST body still echoes `metadata` and carries a `provider_call_id`.

**Endpoint pinning.** Each channel is now pinned to its own exact URL, and the pin lives on the client rather than on the caller: `RestClient` accepts only the REST endpoint, `McpClient` only the MCP one, loopback either way. Swapping `--base-url` and `--mcp-url` is refused before a socket opens — previously both live URLs were trusted for both channels, which is the cross-send you described. Loopback still reads `RINGDOWN_FAKE_API_KEY` / `RINGDOWN_FAKE_MCP_TOKEN`, so a live credential is never available to a local fake. Two tests cover the swap, at the function and through the CLI.

One consequence worth naming: the old `refusing to verify <host> against itself` guard is gone. With each flag pinned to a different host it was unreachable, so it was dead code rather than a second layer.

Suite is 297 tests, no credentials and no outbound calls; `scripts/validate_repository.py` passes.

Ray-56:
Superseding the earlier blocker note: both frozen Must Fix items are closed at 5072838c. The rewritten fixtures and unique history are synthetic and reserved, and the REST and MCP clients each enforce their own exact live endpoint with separate loopback credentials. Repository validation and all 297 tests pass.

PR #208, opened 2026-08-21
feat: add ReadyLine load-in coordinator

Mzoratto:
## Summary

Adds ReadyLine, a runnable TypeScript event load-in coordinator that uses the official CALL-E SDK to collect structured vendor readiness results, detects cross-vendor access, dock, power, deadline, and unknown-input conflicts deterministically, and keeps the resolution follow-up behind explicit human approval.

The public evaluation path is a no-call guided demo using reserved fictional numbers and saved responses. Live mode is fail-closed and requires server-side credentials, an E.164 allowlist, a separate operator key, an explicit live switch, recipient authorization confirmation, and a final operator action.

Public demo: https://readyline-kappa.vercel.app/

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Verification

- `npm run lint`
- `npm test` — 14 unit tests and the Next.js production build pass
- `npm audit --omit=dev` — zero production vulnerabilities
- `python3 scripts/validate_repository.py` — passes from the repository root
- Public demo checked end to end at desktop and mobile widths

## Checklist

- [x] Recurring workflows include cancellation behavior. ReadyLine creates no recurring workflow; one-shot recovery and cancellation limitations are documented.


Ray-56:
Must Fix:

- Replace the plausible non-reserved full phone values in the tests with standards-reserved fictional examples and rewrite the PR history so the old values are no longer reachable.
- Bind live status/results to the expected event, stage, operation, completed task, and exact returned recipient; do not attribute recipient results by array index.
- Fail closed when any vendor readiness is blocked, conditional, unknown, or invalid instead of allowing the overall plan to report Ready from timing/resource checks alone.
- Reject duplicate normalized recipient phone numbers so one batch cannot call the same destination multiple times.

Ray-56:
Current blocker:\n\n- ReadyLine can still report READY when needs_loading_dock is unknown or a nonempty blocker is returned. Validate required operational fields fail-closed, make either condition BLOCKED, and add regressions for both paths.

Mzoratto:
Pushed 956a390 addressing the blocker:

- `needs_loading_dock: "unknown"` now fails closed with an `UNKNOWN_INPUT` conflict (same class as an unusable time/power value) — previously only "yes" was inspected, so an unknown dock need could pass silently.
- A `readiness: "ready"` claim carrying a nonempty `blocker` is treated as a self-contradictory result and emits `READINESS_NOT_CONFIRMED`.
- `readinessSummary` additionally refuses READY on either condition even with an empty conflict list.
- Regressions added for both paths plus the no-conflict-list guard (`tests/constraint-engine.test.ts`, 12/12 passing).

Ray-56:
The frozen false-READY blocker is closed: unknown dock requirements and ready claims carrying a blocker now fail closed in both conflict detection and the final summary, with regressions for both paths. Proportionate validation passes; ready to merge.

PR #211, opened 2026-08-21
feat(callback-scam-screener): add callback-scam screening app

kaddison-nexa:
Reopens the work from #172, which can no longer be reopened directly — a history rewrite needed to scrub a privacy-incident test fixture (see the review thread on #172) unintentionally re-hashed the commits that came in via an upstream merge, severing this branch's shared ancestry with `main` and causing GitHub to auto-close #172 with "no history in common." That's now fixed and independently re-verified; this PR carries the same branch forward.

All of Ray-56's review feedback from #172 (three rounds) is already addressed in this branch's current state — see the discussion on #172 for the specific points raised and how each was resolved.

## Summary

Adds `apps/python/callback-scam-screener/`: CALL-E dials a suspicious "call this number" phone number from a scam email before a human does, has a transparent identity-verification conversation, and scores the transcript against a tiered scam-signal checklist. The agent has nothing real to leak by construction (no account data, no tools to install anything or move money) rather than by prompt instruction alone, and it never decides the verdict itself — scoring happens afterward, deterministically, from the transcript.

`apps/typescript/verify-contact-claim` already covers an adjacent case (verifying contact via the number printed on the customer's own card). This app answers a different question — not "did contact happen" but "does whoever picks up this specific number behave like a scammer" — see the README's differentiation section.

Validated against real usage, not just the mock path: a real CALL-E call (plan → run → poll → transcript), real LLM tagging via Gemini's Interactions API, redaction, and guardrails all ran together successfully.


Ray-56:
Current blocker:\n\n- Live execution still permits an arbitrary destination through --unrestricted. Require affirmative authorization for the exact E.164 destination before placing a call; an acknowledgment that any extracted number may be dialed is not recipient authorization.

kaddison-nexa:
Thanks for flagging this — fixed in [7d6547b](https://github.com/kaddison-nexa/awesome-phone-call-agents/commit/7d6547b).

Added `--confirm-number`, now required alongside `--unrestricted`. It must match `--to-phone` by strict string equality, typed/passed independently as a deliberate second confirmation of that exact E.164 destination.

```
--live --confirm --to-phone "+18005550187" \
    --unrestricted --confirm-number "+18005550187"
```

Ray-56:
The prior exact-recipient authorization blocker is resolved at 7d6547b: unrestricted live mode now requires an independently supplied exact E.164 confirmation matching the destination, with fail-closed regression coverage. No current Must Fix findings remain.

PR #216, opened 2026-08-22
feat(apps): add no-show-killer confirmation app

DavidUmunna:
Express + static-JS app that confirms tomorrow's appointments by phone via the calle CLI and lets the person reschedule on the call, with a dashboard showing live call status over WebSocket.

- Dry-run by default (DRY_RUN=true) - startCall/callStatus never shell out to the real calle CLI while it's on, returning a synthetic response in the same shape a real call produces
- Manual verification path: npm test runs a dry-run smoke test that refuses to execute if DRY_RUN=false
- One-variable cron toggle (CRON_ENABLED) to disable the nightly confirmation batch without touching code
- In-flight guard against dialing the same appointment twice
- Sample data uses a fictional NANP reserved number (+1 202-555-0142)
- apps/README.md: add the app to the index table

## Summary

Describe the change and why it belongs in this repository.

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist


Ray-56:
Current blockers:\n\n- Live mode is remotely triggerable without authentication or authorization and exposes appointment, full-phone, and transcript data through HTTP and WebSocket.\n- Validate destinations as authorized E.164 numbers and mask phones in every response, broadcast, preview, and log.\n- Make the in-flight/idempotency gate atomic and persist an unresolved attempt before dispatch so concurrent requests or an ambiguous start cannot redial the appointment.\n- Do not enable unattended batch calling by default; require explicit operator intent for each live run.

DavidUmunna:
Addressed all four blockers in e930dcd:

- **Auth**: every route that reads appointment data or can trigger a call now requires `Authorization: Bearer <API_TOKEN>`; the WebSocket requires `?token=`. Server refuses to boot without `API_TOKEN` set. Frontend gates the dashboard behind a token prompt.
- **Destination validation + masking**: `ALLOWED_CALL_NUMBERS` is an explicit E.164 allowlist, empty by default (nothing authorized until configured). Phone numbers are masked in every API response, WebSocket broadcast, and error path.
- **Atomic idempotency gate**: an in-memory lock closes the window between two concurrent requests for the same appointment, and a `DISPATCHING` status is persisted *before* calling CALL-E, so a crash or ambiguous response can't leave an attempt untracked and retryable into a duplicate dial. Verified directly with a concurrency test: two simultaneous confirm requests for the same appointment resolve to one dispatch (200) and one immediate 409, before the first call's network round-trip even completes.
- **No unattended batch calling**: the nightly cron now only reports how many appointments are ready by default; `LIVE_UNATTENDED_BATCH=true` opts into automatic dispatch. The manual confirm/batch endpoints are unaffected since clicking them is explicit operator intent.

README and `.env.example` updated to document the new required `API_TOKEN`/`ALLOWED_CALL_NUMBERS` setup.

Ray-56:
Current blockers:

- Redact phone-like data inside provider-derived activity, summaries, transcripts, errors, and logs before persistence or API/WebSocket output; masking only the top-level phone field leaves raw paths.
- Keep timeout, connection, parse, and other ambiguous create outcomes unresolved and non-retryable. The current path rewrites DISPATCHING as retryable DISPATCH_FAILED, allowing a later manual or batch run to redial.
- This head newly sends API_TOKEN in the WebSocket query string, exposing it through URL handling and logs; move it to a non-URL authentication channel.
- Replace the plausible non-reserved full-phone example in the current tree and rewrite the affected unique PR history with standards-reserved fictional data.

Ray-56:
Current blockers:\n\n- Mask the dry-run output: it currently prints the complete destination and appointment goal. Also make phone masking non-overlapping for the shortest accepted E.164 values so summaries/logs cannot reconstruct the full number.\n- Rewrite the PR's unique history so the previously flagged plausible full-phone sample is no longer reachable; replacing it only in a later commit does not remove the public-history exposure.\n- Correct the cancellation claim: restarting the app does not stop an already-started provider call, so the README must not say it stops all real calls immediately. This is newly blocking because it is a newly discovered concrete live-call safety misstatement.

DavidUmunna:
Addressed this round in `43ccd5a`/`e59dc20` (history rewritten - `43ccd5a` replaces the old `e930dcd` to remove the non-reserved phone example from that commit itself, not just fix it forward):

- **Deep redaction**: phone-shaped substrings inside activity messages, summaries, transcripts, and error text are now redacted before persistence, broadcast, or logging - not just the top-level `phone` field. A transcript is free text from an actual conversation and can restate the number in the body.
- **Ambiguous outcomes no longer auto-retryable**: `startCall()` failures are only marked retryable (`DISPATCH_FAILED`) when CALL-E's own response explicitly confirms `call_started: false`. A thrown error (timeout, connection reset, malformed CLI output) or any other ambiguous result now goes to a new `DISPATCH_UNCERTAIN` status that blocks retriggering entirely - no automatic way out, since retrying risks a duplicate real call.
- **WebSocket auth moved off the URL**: token is now sent as a `{type:"auth",token}" message right after connect instead of `?token=` in the query string, which could end up in proxy/access logs and browser history.
- **Non-reserved phone example replaced**, and the commit that introduced it rewritten so it no longer appears anywhere in the branch's history, not just the latest tree.

Verified each of these directly against a running instance (not just read through) before pushing - happy to share the test transcripts if useful.

DavidUmunna:
Addressed the masking/logging/README issues in `37933ce`:

- **Non-overlapping phone masking**: `maskPhone()` now scales how many digits it reveals based on the number's length, so the shortest accepted E.164 values never have their prefix and suffix overlap enough to reconstruct the full number. Verified numerically for the minimum-length case (7 digits) - previously the fixed-width reveal spanned the entire number.
- **Dry-run logging fixed**: the console log and the simulated activity message no longer print the full destination or goal text - the destination is masked, the goal is dropped from output entirely.
- **README corrected**: removed the claim that `DRY_RUN=true` + restart "stops all real calls immediately." It only prevents *new* dispatches - a call already handed to CALL-E keeps ringing regardless of what happens to this app afterward, and there's genuinely no cancellation path for it.

On the git-history exposure: confirmed you're right that my earlier force-push didn't achieve what it needed to - I checked directly and the old commit is still fetchable via the GitHub API (HTTP 200) despite the branch rewrite. That's a GitHub retention behavior (orphaned commits aren't immediately garbage-collected, and the PR timeline records the force-push event with a link back), not something fixable by another force-push. Working with the repo owner now on whether to file a GitHub Support removal request or close and reopen this PR from a clean branch - will follow up here once that's resolved.

Ray-56:
Current blocker:

- The current tree fixes are accepted, but the previously exposed full-phone commit remains publicly retrievable through GitHub. Complete the history/cache removal (GitHub Support may be required) and verify the old commit no longer resolves.

DavidUmunna:
Github support has been contacted about the issue, it should be resoved soon

Ray-56:
Superseding review under the community policy adopted on 2026-09-11: The earlier requirements to purge synthetic/consenting-author test material or certify external production architecture are withdrawn where applicable. The accepted basic safety list remains closed; no actual secret or unauthorized private record was established by the residual item. Maintainer cleanup: Resolved additive app-index conflict.

PR #218, opened 2026-08-22
Add standby skill: sequential shift-fill cascade

nickillig3-dotcom:
## What this adds

`skills/standby/` — fill **one** open shift from a standby roster with a strictly sequential call cascade that stops at the first acceptance.

05:40, a care home is one nurse short. A supervisor rings down a list of twenty casual staff, one at a time, until someone says yes. It has to be one at a time: ring all twenty and four say yes, and now somebody has to un-invite three people who already got out of bed — worse for them than never being called, and the fastest way to empty a standby list.

## How it differs from what is already here

I read the catalogue before writing this, and two entries are close:

- **`mobilize`** dispatches **parallel wave calls** to gather N confirmations and stops when the need is met. Standby is the opposite problem: when there is exactly **one** slot, parallelism *is* the failure mode. Different rule, different guard — a second concurrent call raises an error here rather than being a fast path.
- **ShohojSheba Voice** (app) is the nearest in domain — healthcare staffing dispatch that advances after a verified decline. It *pauses* on acceptance and keeps assignment human-controlled. Standby *stops*, treats the acceptance as the fill, and adds the parts a shift needs: a single no-answer retry pass at the end of the list, callbacks honoured once at the time the callee named, quiet hours with an imminent-shift override, a shift-start cutoff, and a late-arrival check.

## The parts that are not a loop

| | |
|---|---|
| One call in flight, ever | Enforced. Two in flight can produce two acceptances for one shift. |
| Stop on first accept | Everyone still queued is explicitly *stood down* and recorded — not left dangling as "queued", which in a log is indistinguishable from a crash. |
| No-answer: one retry, at the end | Never an immediate redial. Ringing someone twice inside a minute at 05:40 is how people leave a standby list. Nobody is rung a third time. |
| Quiet hours with a number on the override | 22:00–05:00, overridden only when the shift starts within ninety minutes — because a 06:00 shift is exactly when you *do* ring at 05:40. Policy, not a judgement the agent makes on the call. |
| A late yes is not a clean fill | "I can be there by eight" for a 06:00 shift still stops the cascade, but is reported as *120 minutes into the shift — the gap is covered, not closed.* |
| An answer nobody could read halts everything | Not a decline. Nobody further down the list is rung until a person says what the call meant, by name, in the log. |

## Reading the answer

Three outcomes, not two.

**Anything that is not a clear yes is not a yes** — the trap worth naming is that *"Sorry, I can't take it today"* contains the word "can", so refusal markers are tested before agreement markers, and only against what the **callee** said. The agent's own script is written to be agreeable and says "Ja, gerne" in nearly every call.

**And anything that is not clear at all is not a no either.** A decline *advances* the list, so recording an unreadable call as one harms two people at once: whoever was agreeing into a dropping line is silently passed over, and the next person is rung about a shift that may already be taken. Neither leaves a trace — the log says "declined" and the morning looks ordinary.

So an unreadable answer is its own outcome and it **stops** the cascade. Attempting any later call while one is held raises an error, the same guard and the same reasoning as the one-call-in-flight rule. Nobody is stood down — a hold is a pause, not an ending. Resolving it requires a person, their name recorded on the attempt, and a real outcome rather than another hold.

Held on: nothing decidable; a summary containing a refusal *and* an agreement; an agreement the run itself does not mark completed; a lone "ja" in a transcript with nothing corroborating it.

## Safety

Numbers come from the user and are never sourced, guessed or completed by the skill. Opted-out people are skipped **before** dialling with the reason recorded. The agent does not negotiate pay or terms. Nothing recorded from a real conversation appears in the skill, its examples, or the linked reference implementation.

Every number in the skill is inside a range a regulator or a network operator has **permanently withheld from allocation** — German mobile operators' media blocks, Bundesnetzagentur drama ranges, Ofcom's drama range, NANPA's 555-01xx — listed with sources in `references/safety.md`. Not a range that merely looks unused: **"looks fake" is not a property a phone number has.**


Ray-56:
Current blockers:\n\n- The linked public project contains recorded real-call fixtures, transcripts, provider identifiers, and screenshots. Replace them with wholly synthetic evidence and scrub the relevant public history.\n- An unparsable or ambiguous result is treated as a decline and advances the cascade. Stop instead and require human reconciliation before any later call.\n- Replace the plausible non-reserved full-phone example with standards-reserved fictional data and remediate the relevant history.

nickillig3-dotcom:
All three are fixed. Taking them in the order you raised them.

**1 — Recorded real-call evidence, and the public history**

You were right, and it was worse than the fixtures: a real German mobile number
was also sitting in `scripts/test-fixtures.mjs` as a negative assertion — a line
checking the number did *not* appear in the fixtures, which published it on the
line doing the checking.

- The four recorded calls are gone. What fixture mode replays is now written by
  `scripts/make-fixtures.mjs`. `npm test` re-runs that generator in `--check`
  mode, so the files on disk are provably its output rather than a capture —
  regenerate and diff instead of taking "synthetic" on trust.
- Provider identifiers are readable words (`synthetic-run-01-no-answer`), never
  plausible hex. A test asserts it: a fabricated hash in a public repo is
  indistinguishable from a real one and somebody will eventually try to resolve
  it.
- All screenshots were reshot against the authored fixtures.
- `scripts/capture-call.mjs`, which produced the original captures, now writes
  to a git-ignored `captures/` directory and cannot write into `fixtures/`.
- The public history of the linked repository was reset. The commits containing
  the transcripts, the identifiers and the number are not reachable there any
  more.

The trade this costs, stated in the README rather than hidden: a capture breaks
loudly when the provider changes its response shape, and an authored fixture
cannot. Shape drift is now caught only by `CALLE_LIVE=1` against the live API.
That is the right way round.

**2 — Unparsable or ambiguous results advancing the cascade**

This was the one worth stopping for, and your framing was exactly right — the
`unparsed` flag was honest and the behaviour was not. A decline *advances* the
list, so an unreadable call was operationally a no.

There are now three outcomes, not two. `RESULT.AMBIGUOUS` halts the cascade:

- **No further call is placed by anybody.** `applyEvent` throws on
  `call_started` while a candidate is held, the same guard and the same
  reasoning as the existing one-call-in-flight rule. A rule that lives only in
  the runner's loop is a rule every other caller walks past.
- **Nobody is stood down.** A hold is a pause, not an ending; recording it as an
  ending misreports the morning in the audit log.
- **Resolution requires a person and their name.** A `reconciled` event carries
  `by`, cannot resolve back to ambiguous, and lands in exactly the state the
  classifier would have produced. `finishIfDone` will not convert a hold into a
  cutoff or an exhaustion while it waits.

Held on: nothing decidable; a summary containing a refusal *and* an agreement;
an agreement the run itself does not mark completed; a lone "ja" in a transcript
with nothing corroborating it.

One related bug found on the way: the transcript fallback matched the whole
transcript, including the agent's own lines. The agent's script is written to be
agreeable and says "Ja, gerne" in nearly every call, so a bare word-match on it
could read as a filled shift. Only the callee's half is classified now.

Eleven assertions cover the hold, including that a held cascade replays from its
event log to the same held state, and that the cutoff still applies when the
shift starts underneath one.

**3 — The plausible non-reserved full-phone example**

Replaced everywhere, and the reason it happened is worth naming: the old numbers
were a live German mobile prefix with a run of zeroes behind it, chosen because
they look obviously fake. **"Looks fake" is not a property a phone number has.**

Everything now comes from ranges a regulator or a network operator has
permanently withheld from allocation, listed with sources in
`references/safety.md`:

| Range | Withheld by |
|---|---|
| `+49 176 040690 00`–`99`, `+49 171 39200 00`–`99` | German mobile network operators, for media use |
| `+49 30 23125 000`–`999` and four other city blocks | Bundesnetzagentur, for film and television |
| `+44 7700 900000`–`900999` | Ofcom, drama range |
| `+1 NPA 555 0100`–`0199` | NANPA, fictional use |

Because reading carefully is what failed here, it is enforced mechanically:
`scripts/test-numbers.mjs` scans every text file in the repository — prose
included — and fails on any number-shaped string outside those ranges. It runs
in `npm test`.

The branch has been rebased onto current `main` and `scripts/validate_repository.py`
passes. Thank you for the review — the second point in particular was a real
defect, not a documentation problem.

Ray-56:
Superseding the earlier blocker note: all frozen Must Fix items are closed at 0e7538a5. The linked public history is rewritten to synthetic fixtures and reserved numbers, and ambiguous results now halt for named human reconciliation. Repository validation and all 63 linked implementation assertions pass.

PR #226, opened 2026-08-24
feat(apps): add kept, a promise-to-pay capture and reconciliation app

rushabhT3:
Adds `apps/python/kept`, a runnable accounts-receivable app built on the CALL-E Python SDK.

## What it does

A promise made on a collections call — "we'll send four thousand eight hundred on the 31st" — is the most useful cash-flow signal a finance team gets, and it usually ends up in a CRM note nobody reconciles. `kept` phones the customers behind overdue B2B invoices, turns what they commit to into a validated financial record, and checks the bank feed a week later to see who actually paid.

The call is the small part. The app is mostly the decisions on either side of it:

- **Before dialling** — ten named suppression reasons. `do_not_call` first, then an unauthorized recipient, an open dispute, an open promise, quiet hours evaluated in the customer's own timezone, contacted too recently. Every refusal names itself in the plan output.
- **After the call** — CALL-E's structured result never creates a record on its own. Eleven named rejection reasons stand between a spoken sentence and a ledger entry. Vague amounts are refused, a date with no amount is not a commitment, and an over-commitment is clamped to the invoice balance with the spoken figure preserved beside it for audit.
- **A week later** — bank lines are allocated to the referenced invoice first and then oldest-first, no payment spent twice, and only the promises that actually broke are called again.

Escalation is driven by promise history rather than invoice age: first contact, reminder, broken promise, final notice.

## Notes for review

**No-call path is the default.** `kept run --simulate` installs an `httpx.MockTransport` underneath the real `calle.CalleClient`, so the SDK still builds the request body, attaches the `Idempotency-Key` header, polls `GET /v1/calls/{id}` to a terminal state and maps API errors. Only the wire is replaced, which lets the tests assert on the bytes CALL-E would have received. No API key, no network.

**Live calling is triple-gated** — `KEPT_LIVE_CALLS_ENABLED=true`, an exact `--confirm PLACE-REAL-CALLS`, and the recipient's exact number present in `authorized_recipients.txt`. The first two authorise the run; the third authorises the destination, and a customer row on its own is deliberately not enough to make a phone ring. `--as-of` is rejected outright in live mode so a real call is never placed against a pretend date. `kept doctor` checks credentials by requesting a call id that cannot exist, so `not_found` proves the key without spending a call.

**Phone numbers are strict E.164.** `Customer` and `Organisation.callback_number` both reject anything that is not `+` followed by 7 to 15 digits — separators, extensions and a missing country code are refused at load time rather than passed to the provider as a destination it might still try to route.

**The credential only ever reaches the official API.** `CALLE_BASE_URL` may name `https://api.heycall-e.com` and nothing else; any other value raises before the key is attached to a request. The simulator uses its own transport and never passes through credential loading.

**Results are bound to the call that produced them.** Before `capture` reads a field, the answer must agree with the request: the call id, the recipient CALL-E says it dialled, the task text, and the `invoice_id` / `customer_id` / `cycle` metadata. CALL-E echoes all of these on every call, so each is required: a result that omits any of them is `result_not_bound`, and so is any disagreement. The `call_dispatched` ledger record carries a digest of the task text, so a call recovered after a crash is held to the same proof rather than bound only by id. `task_completed` must be true, not merely a terminal status, and right-party validation applies to a stated dispute as well as to a promise. The idempotency key carries a digest of the payload, so a reused key can only return a call placed with exactly these instructions; `run_id` is excluded from the digest because a key that changed every run would stop covering the crash it exists for.

**Cancellation, rollback and ambiguity.** There is no scheduler and no daemon. Every run is one command that terminates, so there is nothing to cancel between runs and no state that keeps dialling on its own. A rejected request is a fact; a timeout or a dropped connection is not, so an ambiguous create or poll outcome stops the run where it stands rather than starting another call on top of one nobody can account for. The ledger records it with `"ambiguous": true` and `kept recover` reads the outcome without re-dialling anyone.

**Credentials** are read from the environment only; nothing parses a key off disk and nothing logs one.

**Phone numbers** are masked to their last two digits wherever they are persisted or displayed, including inside provider-derived free text — evidence quotes, dispute reasons, transcripts, summaries and failure messages all pass through `redact_phone_like` before they are written or rendered, and grouped amounts and ISO dates are left alone. The HTML report contains no phone number at all and a test enforces that. Samples use the NANP fictional `555-01XX` block (`+1 202 555 01XX`) and Ofcom drama ranges; the opt-in live folder ships an all-zeros unroutable placeholder.

**Boundaries on the call itself** are in the task text and asserted in tests: no card numbers, CVV codes, bank credentials or one-time passcodes; no threats of legal action or credit reporting; no discounts, waivers or new terms; a stated dispute ends the collection attempt; a do-not-call request is agreed and confirmed; voicemail gets the creditor name and a callback number and never the amount.

**Ledger.** Append-only, hash-chained JSONL. Promises, disputes and contact history replay from it on every run, so the report cannot disagree with the audit trail. `kept verify` detects an edited or removed entry.

## Verification

- `pytest` — 159 tests, no network, no credentials
- `python3 scripts/validate_repository.py` — passed
- `python3 scripts/check_branch_name.py` — passed
- Verified end to end against the live API, which is where several of the on-call rules in `kept/calls/scripts.py` came from; each is documented in the app README as a generalised failure mode with the test that pins it

## Checklist

- [x] No secrets, tokens, private phone numbers, or personal data
- [x] States the provider it supports (CALL-E Python SDK, `calle-ai>=0.7.0`)
- [x] Describes side effects
- [x] Install and usage instructions in the app README
- [x] Fictional or placeholder phone numbers in samples
- [x] Cancellation and rollback behaviour documented
- [x] No-call path by default
- [x] Passes repository validation


Ray-56:
Current blockers:\n\n- Enforce strict E.164 validation for every customer and creditor callback number; checking only for a leading + accepts invalid destinations.\n- Restrict CALLE_BASE_URL to the exact official HTTPS CALL-E API origin before attaching a production credential.\n- Require affirmative authorization for each exact live recipient; a generic run confirmation plus absence of a do-not-call flag is not recipient authorization.\n- Bind every accepted result to the exact task, metadata, recipient, and operation, require successful task completion, apply right-party validation to disputes, and use a payload-bound idempotency key before writing financial records.\n- Stop the run after an ambiguous create or poll outcome instead of advancing to later calls.\n- Mask phone-like data in provider-derived evidence and failure text before persisting or rendering it.\n- Replace the committed real-test-call excerpts and apparent provider call identifier with synthetic or generalized cases, then rewrite this PR's unique history without those artifacts.

rushabhT3:
@Ray-56 check once

Ray-56:
Current blockers:

- Result binding still accepts provider responses that omit the exact task, recipient, or expected metadata. Require those fields, or an equivalent persisted payload digest, to be present and match—including recovery—before recording a financial promise or dispute.
- Remove the remaining real-call-derived speech excerpts.
- Replace plausible non-reserved full-phone fixtures with standards-reserved values and rewrite the affected unique PR history.

rushabhT3:
@Ray-56  check once

Ray-56:
Superseding the earlier blocker note: all frozen items are closed at 3d07e35. Results now require exact task, singleton recipient, and metadata binding including recovery; the rewritten contribution contains generalized failure cases and reserved fictional examples. The maintainer-assisted recipient fix passes all 160 tests and repository validation.

PR #231, opened 2026-08-25
feat(ghost-network-audit): add provider-directory ghost audit skill

PranavAchar01:
## Summary

Health-plan provider directories list clinicians who left years ago, offices that dropped the plan, and panels that are closed to new patients. A patient works down the list and finds nothing on it is true. The only way to test a listing is to call the office and ask, which is why this work is still done by people dialing one number at a time.

`ghost-network-audit` packages that audit as a bounded CALL-E workflow:

- **One office, one call.** Listings are de-duplicated by phone number, so a practice with three listed clinicians gets one call asking about three names rather than three calls to the same front desk.
- **Disclosed, administrative-only.** The call opens by identifying itself as automated, naming the auditing organization, and offering a callback number. It then asks four business questions — is this clinician here, do you accept this plan, are you taking new patients, how far out is scheduling. It never mentions a patient, never asks a clinical question, and never books anything.
- **Never infers a ghost.** A listing is marked a ghost only when a person at the office said so. Voicemail, a refusal, an answering service, and "I'd have to check with billing" all land on `unverified`. There is no rule anywhere in the mapping table that turns a missing answer into a negative finding — a false ghost strikes a working clinician from a directory, and a false confirmation leaves a patient dialing a dead number.
- **Rates carry their denominators.** Every metric is computed over confirmed rows only and rendered beneath its coverage figure, so a ghost rate cannot be lifted out of the report as a claim about listings the audit never reached. `effective_availability` is reported separately from accuracy, because a directory can be perfectly accurate and still unusable if every panel is closed.

### Side effects

A live run places one real outbound phone call per office in the input file. Live mode requires **all** of `CALLE_API_KEY`, `CALLE_LIVE_CALLS_ENABLED=1`, an explicit `--live` flag, and real `--auditing-organization` / `--callback-number` values. Missing any one of them previews instead of dialing, and prints which one was missing. There is no recurrence: one run is one pass over one list, and cancelling means stopping the process — which is why every gate runs before dialing rather than after.

Emergency, crisis, nurse-triage, on-call, and after-hours lines are skipped with no override flag and no dry-run exception. Calling windows use the office's own timezone from the listing, never an area-code guess.

### Verification

Runs with no credentials and places no calls by default:

```bash
cd skills/ghost-network-audit

# preview - dials nothing, needs no key
node scripts/audit.mjs --listings scripts/sample-directory.json --now 2024-06-11T17:00:00Z

# full offline rehearsal against the bundled fake CALL-E server
node scripts/fake-calle-server.mjs &
node scripts/audit.mjs --listings scripts/sample-directory.json \
  --base-url http://127.0.0.1:8787 --now 2024-06-11T17:00:00Z \
  --auditing-organization "Example Health Directory Audit" \
  --callback-number "+12125550100"

node scripts/report.mjs --run out/audit-run.json --out out/report.html
node scripts/test.mjs   # 23 tests
```

The fake server deliberately returns 422 `call_not_ready` for a payload whose task text omits the automated-call disclosure — a fake that accepts anything only proves the client can send.

Zero dependencies, standard-library Node only. All sample numbers are in the fictional reserved 555-01xx range.

## Type

- [x] New skill
- [ ] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist


Ray-56:
Current blockers:

- Require explicit authorization of each exact live recipient and validate the callback as E.164.
- Pin credentialed live requests to the exact official CALL-E HTTPS origin; loopback rehearsal must never receive the production key.
- Stop globally after any ambiguous create/poll outcome or timeout; the current concurrent workers continue placing later calls.
- Parse the documented CALL-E `structured_result` envelope and require terminal success plus exact call/task/recipient/metadata binding before publishing any confirmed ghost finding.
- Do not persist returned free-text notes that may contain volunteered patient, clinical, or contact data.
- Persist refusals, suppression, and attempt counts so later runs cannot redial a refusal or exceed the documented retry cap; if any grouped listing marks a blocked line, fail closed.

Ray-56:

Remaining Must Fix only:

- Require authorized valid destination and protect credential origin.
- Concurrent workers continue after ambiguous call outcomes; stop later side effects.
- Correct incompatible structured_result parsing and avoid storing volunteered private clinical/contact notes.

PR #240, opened 2026-08-25
feat(apps): add CallAction CI/CD escalation tool

greatsage-raphael:
## Summary

This PR adds **CallAction**, a highly portable GitHub Action tool under `apps/typescript/callaction`.

It bridges the gap between CI/CD failures and immediate human response by placing a consent-first, idempotent CALL-E phone call to on-call engineers when a pipeline step fails. It extracts their structured decision (e.g., revert, escalate) and posts a masked report directly to the triggering Pull Request.

It belongs in this repository as an  example of how CALL-E can be used as a webhook for the physical world in DevOps/Infrastructure scenarios.

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

- [x] Real-world side effects are clearly described (README explicitly warns of one outbound call per CI failure).
- [x] Phone numbers are masked in documentation and test fixtures unless they are clearly fictional (enforced in code via masking before GitHub logging).
- [x] Recurring workflows include cancellation behavior (N/A, one-time call).
- [x] Runnable code has a dry-run, fake-server, or no-call path by default (N/A, this is a CI pipeline action, but idempotency prevents duplicate dialing).


Ray-56:
Current blockers:

- Add a no-call/dry-run default and an explicit authorized per-run live gate for the exact recipient; the Action currently dispatches a real call whenever it executes.
- Commit a working dist/index.js entry and add a passing credential-free verification path. The declared Action entry is absent, TypeScript compilation fails because of the current module configuration, and the test script intentionally fails.
- Mask malformed phone input instead of echoing it into workflow logs.
- Do not publish raw provider-derived engineer_notes to a Pull Request; redact or constrain the field before posting.

Ray-56:
Current blockers:

- The clean-install verification blocker remains: package.json and package-lock.json disagree on the Node type dependency, so npm ci fails and the bundled Action cannot be reproduced or validated from a clean checkout.
- Newly introduced on this head: preview mode logs the task prompt containing the full destination number. Omit or mask the number before writing the preview log.

greatsage-raphael:
Thanks for the thorough review, I've pushed a series of updates that address both blockers:

1. **Clean Install (`npm ci`)**: I wiped the old lockfile and `node_modules` and generated a fresh `package-lock.json` that perfectly aligns with `package.json`. A clean `npm ci` will now succeed, allowing the bundled Action to be reliably reproduced from a clean checkout.
2. **Preview Mode Masking**: Fixed the prompt logging leak. The destination number is now explicitly masked within the `taskPrompt` string (e.g., `+14******341`) before it is written to the preview logs. I also added a unit test to explicitly enforce this masking behavior.

Ready for another look!

Ray-56:
Superseding review for head `ac84916`: the frozen clean-install and preview-number blockers are closed. Repository validation, clean install, three credential-free tests, and bundled-entry syntax validation passed. No current Must Fix remains.

PR #244, opened 2026-08-27
feat(apps/web): add Locked Door, an emergency-directory phone verifier

rishabhcli:
## What this adds

`apps/web/locked-door/` — a web app that keeps **emergency resource directories** (cooling centers, heat-relief shelters) truthful by verifying, over the phone, the facts that are most dangerous when stale.

The operational fact that matters in a heat emergency — are you open right now, do you take pets, is the accessible entrance still there — very often exists only behind a phone number. This is the policy layer around that call: which facts are worth verifying, what counts as evidence, when to say you do not know, and what may be published without a human.

## Safety: no real calls are placed

The call transport is an interface with two implementations. The real CALL-E adapter is **present but disabled**, documenting the request shape, and `placeCall()` **throws synchronously** so an unawaited call cannot slip past the refusal silently. A deterministic simulated facility with per-facility ground truth backs the demo, including IVR menus, voicemail, hedging and staff who do not know.

A full session makes **zero external network requests** and handles no credentials. This satisfies the CONTRIBUTING requirement for a local dry-run path.

## The rule that makes it trustworthy

Every extracted field must cite a span of the transcript. **If no span supports it, the value is `unknown`** — enforced in code, not by convention, with the citation visible in the UI. Contradicted or low-confidence fields go to a review queue rather than to publication, and output is a time-stamped delta awaiting human approval, never an overwrite.

## Measured, not asserted

Because the simulator carries ground truth, the app scores itself. From a 25-call batch (`node tools/smoke.mjs 25`; browser and headless produce identical numbers):

| Metric | Value |
| --- | --- |
| Precision | 94.4% (67/71) |
| Recall | 44.7% (67/150) |
| Unknown rate | 52.7% |
| **False-publish rate** | **0.00%** (0/37) |
| Wrong values produced / held for review | 4 / 4 |
| Harm reduction vs oldest-first baseline | 49.31 vs 39.35 (**+25.3%**) |
| Citation audit | 71/71 pass, 0 failures |

The citation audit is deliberately independent — it re-slices each stored span out of its retained transcript rather than reusing the engine's own gate.

## Honest limits (also stated in the README and on screen)

- **Recall is 44.7%.** Over half of attempted facts stay `unknown`. That is the designed trade, not a good number in isolation.
- The 0.00% false-publish rate is the rate at which the *system* publishes a wrong value on its own authority. If a human blanket-approves the whole review queue unread it becomes 5.63%; both are shown, because `approveAll()` simulates exactly the rubber-stamping the queue exists to prevent.
- `capacity_status` recall is 24%; `open_now` precision is 84%, weakened by stale voicemail greetings being cited as evidence about today.
- The extractor is rule-based, so unfamiliar phrasing produces `unknown`, never a wrong answer.
- The directory is synthetic — realistic schema, invented facilities.

## Run it

```bash
cd apps/web/locked-door && python3 -m http.server 8810
```

Static files only: no build step, no CDN, no backend. Headless: `node tools/smoke.mjs 25`.

Upstream source: https://github.com/rishabhcli/call-e-your-code-is-calling
Demo video: https://youtu.be/iI-PyPrgk58

### DIFF
--- a/apps/web/locked-door/LICENSE
+++ b/apps/web/locked-door/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 CALL-E AI
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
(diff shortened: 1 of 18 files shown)


Ray-56:
Current blocker:

- The synthetic directory commits many plausible full phone numbers outside standards-reserved fictional ranges. Replace them in the current tree and generator, rewrite the PR’s unique history, and sanitize the linked upstream public history as well.

Ray-56:
Superseding review under the community policy adopted on 2026-09-11: The old blocker on clearly synthetic directory fixtures is withdrawn. This build structurally disables real telephony; its deterministic smoke workflow runs without network or credentials. Reserved example formatting remains an optional follow-up.

PR #243, opened 2026-08-26
Add Clinic Appointment Concierge workflow plugin

iancuileana83-lab:
Adds a CALL-E-powered workflow plugin that calls a clinic to schedule a patient appointment, negotiating a different time if the preferred slot is unavailable. Includes README with setup/usage instructions.

PR #249, opened 2026-08-27
feat(apps): add clinic appointment concierge with safety gates

iancuileana83-lab:
Adds a CALL-E-powered workflow plugin that calls a clinic to schedule a patient appointment, negotiating a different time if the preferred slot is unavailable.

Addresses review feedback from PR #243:
- Moved to plugins/clinic-appointment-concierge/
- Removed personal contact info; examples now use fictional/reserved numbers
- Added dry-run default, explicit --live confirmation, ALLOWED_RECIPIENTS authorization gate, strict E.164 validation, and phone masking
- Replaced shell string interpolation with execFile argument-array execution (no command-injection surface)
- Added package.json; implemented the actual retry/confirmation flow instead of only printing instructions
- Output is a structured, length-capped summary rather than raw provider transcripts

Closes #243


Ray-56:
Current blockers:

- Move this runnable command-line integration from plugins/ to the appropriate apps/ runtime directory.
- Update the live flow to use the actual CALL-E CLI contract. The current plan/run commands use unsupported flags and omit the required run confirmation token, so the documented workflow cannot execute.
- Make failed, timed-out, and ambiguous outcomes exit non-zero when the operator stops; the current path returns success despite the README promise.

iancuileana83-lab:
Thanks for the review — all three blockers are addressed in the latest commit:

1. Moved the app from plugins/clinic-appointment-concierge/ to apps/typescript/clinic-appointment-concierge/, matching the runnable-app directory template.

2. Updated the live flow to match the actual CALL-E CLI contract: `calle call plan` now uses `--to-phone` and `--goal` (patient name, reason, and preferred time are composed into the goal text), and `calle call run` now passes both `--plan-id` and the required `--confirm-token` returned by `plan`.

3. Failed, timed-out, and ambiguous outcomes — including when the operator declines to retry — now set a non-zero exit code, matching the README's documented behavior.

README has been updated to reflect the new path and the corrected CLI flow. Ready for another look whenever you have time.

Ray-56:
Re-review of the changed head against `origin/main` confirms that the frozen Must Fix list is closed: the runnable CLI is under `apps/`, it uses the current CALL-E `plan` → `run` → `status` contract, and failed/timed-out/ambiguous outcomes stop with a non-zero exit instead of advancing automatically. I also corrected the stale PR title scope from `plugins` to `apps` as a small mechanical maintainer fix. Remaining hardening items are non-blocking for this demo.

PR #251, opened 2026-08-28
feat(apps): Reality Resolver - decision-gated calling for CALL-E

Angelo-The-S:
## Summary

Reality Resolver decides whether a call is worth making before it
decides whether it is allowed. Most outbound-call automation asks one
question - "am I allowed to call?" - and calls the moment the answer
is yes. Reality Resolver asks a question first: "do I actually need to
call?" Evidence already on hand (a calendar entry, an email, a missed
follow-up) can leave a real decision unresolved without anyone
noticing - a state nobody has confirmed. Reality Resolver detects
exactly that decision-critical uncertainty, and only then escalates to
a real, compliance-gated CALL-E call to resolve it.

**Do I need to call, and am I allowed to call?** That's the two-gate
mechanism, Decision-Gated Calling:

```
Evidence sources (fixtures JSON)
  -> Evidence Matrix
  -> 4 generic rules (R1-R4)
  -> decision-critical uncertainty?
       NO  -> NO_CALL_NEEDED
       YES -> call justified -> call permitted? (compliance gate)
                NO  -> UNRESOLVED_CALL_BLOCKED, RETRY_WHEN_PERMITTED
                YES -> CALL-E -> structured_result -> reconciliation
                       -> RESOLVED / RESOLVED_ALT / UNRESOLVED_AMBIGUOUS
```

Four independent, generic rules (`evidence/rules.py`) decide the first
gate: R1 a structured source asserts a state, R2 a human source
diverges from it, R3 nothing since has resolved that divergence, R4
the decision has a near deadline. All four true means the uncertainty
is decision-critical - only then does a call get evaluated against the
second gate at all.

**`UNRESOLVED_CALL_BLOCKED` is not `UNRESOLVED_AMBIGUOUS`.** The first
means no call was placed: the uncertainty was decision-critical, but
the compliance gate refused (missing consent, wrong hour, and so on) -
the recipient never heard the phone ring. The second means a call
*was* placed and answered, but CALL-E's own result did not cleanly
confirm or cancel (voicemail, an IVR, or a genuinely uncertain
answer). Confusing the two would be a real safety error - this PR
keeps them structurally distinct end to end, and an absolute rule,
checked as a test invariant across all 25 `(patient_intent,
answered_by)` combinations: unresolved evidence is never silently
treated as a cancellation.

**Demonstration: the Ghost Appointment.**
`cases/ghost-appointment.json` - a dental practice's calendar says a
patient's appointment is confirmed for tomorrow 14:00 (structured, low
ambiguity). A separate email from the patient says "I may need to
cancel" (human, high ambiguity). A scheduled follow-up never got a
reply. Nobody has reconciled these two accounts. Reality Resolver
detects the contradiction, checks whether calling is legally
permitted, places the call through CALL-E, and reconciles the result
into `KEEP_SLOT` or `RELEASE_SLOT` - or `HUMAN_REVIEW` if the call
didn't resolve anything.

## What's reused, honestly

The legal/compliance layer (`compliance/`: fail-closed jurisdiction
resolution across `us_federal`, `eu_common`, `fr`, `us_oregon` -
calling window, consent, DNC, AI disclosure, GDPR basis, revocation)
and the CALL-E REST client (`client.py`: never-retried call creation,
indefinite polling, and the task-hardening blocks - AI disclosure script,
injection-resistance, voicemail handling, no-repeat-opening,
proactive-next-step, call-closing) originated as an earlier,
independent submission on this same repository and PR
(`compliance-gated-callback`). They were reused as a base rather than
rewritten or re-claimed as new work, and the directory evolved in
place (`git mv apps/python/compliance-gated-callback
apps/python/reality-resolver`, history preserved per file) rather
than opening a competing submission.

Since then that layer has been **narrowed and hardened**, so it is no
longer reused verbatim: everything specific to the earlier
commercial-outbound use case was removed (its standalone CLI, the
business-context feature, the web form and public demo, the
marketing-consent retention helper), and `compliance/` gained a
`use_cases.py` filter so that rules scoped in their own statutes to
*telephone solicitation* are not applied to an appointment
confirmation. `compliance/dispatcher.py` and
`compliance/jurisdictions/*.py` themselves remain untouched and
purpose-agnostic.

## What's new

- `evidence/` (model, rules, engine) - the decision layer described above.
- `verdict.py` - reconciliation of CALL-E's `structured_result` into a
  final verdict, plus `patient_intent_result_schema()`.
- `next_window.py` - projects the next locally-legal calling window
  when `UNRESOLVED_CALL_BLOCKED`, only when every blocking reason is
  time-based; says plainly when it isn't (consent/DNC/disclosure/revocation).
- `resolver.py` - the CLI entry point, printing five sections
  (EVIDENCE STATE, REASONING, CALL JUSTIFICATION, CALL PERMISSION,
  CALL-E, VERDICT).
- `cases/ghost-appointment.json` and the corresponding `tests/` suite.
- `fake_server.py` extended with three reserved phone numbers to
  simulate CALL-E's confirmed/cancelled/voicemail branches without a
  real call.

## Security hardening (review follow-up)

Three items raised in review, addressed in `7f5ad2c`:

- **Public history.** The commits that carried real call identifiers and a
  transcript excerpt were rewritten (content *and* commit messages) and
  force-pushed; the current range is clean. No sensitive value is repeated
  here or anywhere in the tree.
- **Strict ASCII E.164 + exact-destination authorization.** The phone
  pattern uses `[0-9]` rather than `\d`, which for a `str` pattern matches
  every Unicode decimal digit - Arabic-Indic, Devanagari, fullwidth and
  Burmese digits previously passed validation and would have been sent to
  the provider verbatim. A real call now also requires
  `--authorize-destination <E.164>`, compared byte-for-byte against the
  number that will actually be dialled (the case file's `call_phone`, or
  `--phone` when it overrides it), with no normalization; a missing or
  mismatched value is refused before any client is constructed.
- **Provider output sanitization.** `redacted_call_for_display()` is now the
  single central point for rendering a provider response: deep copy, phone
  masking, then recursive control-character escaping and a per-field length
  limit (never a truncation of the serialized document, so the JSON stays
  valid). The paths that interpolated provider values outside that JSON -
  created call id/status, poll status and warning, the Ctrl+C handler, a
  retryable `error.code`, connection-error reasons, and the verdict's
  evidence lines - go through the same helper. The engine's own data is
  deliberately untouched: `verdict.reconcile()` still receives CALL-E's
  `structured_result` exactly as returned.

## Testing

168 tests (up from 35 at original submission, 115 immediately before
this evolution), all against a local fake REST server.
`python3 scripts/validate_repository.py` passes. ASCII-clean. Verified
end to end in an isolated clone with no local `.env`/`.venv`. The
three CALL-E branches (confirmed/human, cancelled/human, voicemail)
were run against the fake server for the Ghost Appointment case
specifically and produce `RESOLVED`/`KEEP_SLOT`,
`RESOLVED_ALT`/`RELEASE_SLOT`, and
`UNRESOLVED_AMBIGUOUS`/`HUMAN_REVIEW` respectively.

**Not yet done, stated plainly:** no real call has been placed against
the new `patient_intent` schema. Every fix that mattered in the
original submission was found by a real call, not the fake server -
this is the one thing still pending before calling the decision layer
proven, not just tested.

## Type
- [x] New runnable app (evolved from an existing one on this same PR)

## Checklist


Ray-56:
Current blockers:

- Restrict credential-bearing requests to the official HTTPS CALL-E origin. Default dry-run must not require or partially print an API key, and fake-server use must not receive a real credential.
- Require a stable intent-bound idempotency key and stop after ambiguous POST outcomes instead of automatically retrying call creation.
- Mask the recipient in previews and validation errors instead of printing the full number.
- Replace the plausible complete French and Japanese subscriber identifiers with standards-reserved or structurally unassignable examples, and remove them from the PR history.

Ray-56:
Superseding review for head `487a668`: two newly introduced blockers reopen the previously closed acceptance list:

- The changed head automatically retries an ambiguously failed call-creation POST. Even with the same idempotency key, this workflow’s frozen acceptance requires stopping for operator reconciliation after an ambiguous provider outcome.
- New commits and current documentation/code contain real-call identifiers and transcript/result excerpts. Remove or sanitize these artifacts and rewrite the PR history so they are no longer reachable.

This remains a draft and is not eligible for merge.

Ray-56:
Must Fix for head dfcff71e:
- The frozen ambiguous-outcome blocker remains: do not automatically retry a call-creation request after a timeout or uncertain provider result.
- Remove real-call identifiers, transcript excerpts, and result artifacts from the current tree/docs and every content-bearing PR commit; rewrite history without repeating exposed values.
- Newly blocking: demo mode must not permit a real call when live authorization/safety policy fails. Keep the fake path default and require the same hard gate for every live execution.

Ray-56:
Superseding review for draft head fdd6026 — remaining Must Fix:

- Rewrite every affected content-bearing PR commit so the earlier real call identifiers, results, transcript excerpts, and contact artifacts are removed from public history.
- Enforce strict ASCII E.164 validation and authorization for the exact destination before live operation.
- Deeply sanitize all free-text provider and result output before displaying or storing it.

Angelo-The-S:
Thanks - all three are addressed as of `8521798`. Details below, no sensitive values repeated.

*(Edited: this comment originally referenced `7f5ad2c`. After the `+44` point below, history was rewritten a second time, so the head is now `8521798`. The final tree is byte-identical across that rewrite - only the history changed.)*

**1. Public history.** The three content-bearing commits that carried real call identifiers and a transcript excerpt were rewritten - both the file content *and* the commit messages, since `--amend --no-edit` preserves the latter and my first pass had missed it. Rewritten with `--force-with-lease`.

I re-verified over the **full PR range** (merge-base with `main`, 21 commits), not just the rewritten portion - an earlier check of mine only covered the tail and would have missed the first 9 commits. Clean for: known call identifiers, any opaque `call_...` token, the recipient number, transcript excerpts, and 32-hex provider correlation identifiers.

One leftover I flagged here initially, not in any of your four categories: a `+44` number I had invented for an unmapped-jurisdiction test was plausible-looking rather than drawn from a reserved block, which also contradicted a claim the README makes about itself. It is fictional and unroutable, and I had first fixed it forward in the tree only - but on reflection "clean" should mean clean, so history was rewritten a second time and it no longer appears anywhere in the PR range. Every phone number across the 21 commits now comes from a regulator-reserved block, with one documented exception: the two E.164 length boundaries in the test suite, where no reserved block exists at 7 and 15 digits and both values are structurally unroutable.

**2. Strict ASCII E.164 + exact-destination authorization.**

The pattern used `\d`, which for a `str` pattern matches every Unicode decimal digit. Arabic-Indic, Devanagari, fullwidth and Burmese digits all passed validation and would have been sent to the provider verbatim - and because the jurisdiction router slices the string by index, they also degraded routing silently instead of failing closed. Now `[0-9]`, with tests for all four scripts plus both E.164 length boundaries to prove nothing legitimate regressed.

A real call now requires a third independent authorization, `--authorize-destination <E.164>`. It is compared byte-for-byte against the number that will actually be dialled - the case file's `call_phone`, or `--phone` when it overrides it - with no normalization of any kind, so a number that merely looks equivalent cannot authorize a different one, and authorizing a case file's original number does not authorize an override of it. A missing or mismatched value is refused before the evidence engine runs and before any HTTP client exists; there is a single call-creation path in production, so nothing bypasses it.

**3. Provider output sanitization.**

`redacted_call_for_display()` is now the single central point: deep copy, phone masking, then recursive control-character escaping and a **per-field** length limit. Per field rather than on the serialized document, so the JSON stays parseable. `json.dumps` was already escaping control characters, but only while `ensure_ascii` keeps its default and never with any length bound - a multi-megabyte field was dumped whole.

Seven paths interpolated provider values outside that JSON and now go through the same helper: created call id/status, poll status, the poll warning, the Ctrl+C handler, a retryable `error.code`, connection-error reasons, and the verdict's evidence lines. Two of those I had missed on a first pass and found by re-tracing every `print` rather than trusting the earlier inventory.

The engine's own data is deliberately untouched: `verdict.reconcile()` still receives the provider's `structured_result` exactly as returned, and `CallEAPIError` keeps `.message`/`.details` raw - only `str()` is sanitized. There is a test asserting both halves of that split.

**Verification.** 168 tests, `validate_repository.py` passes, ASCII-clean. The adversarial tests found two real defects in my own sanitizer while being written (it was not idempotent, and could exceed its own limit when the limit was smaller than the truncation marker); both are fixed and pinned by tests.

Also worth flagging, unrelated to the review: programmatic calls to France via the shared pool are currently suspended, which we confirmed with support after two live attempts failed pre-ring. So the live path here is exercised against the local fake server only for now.

Ray-56:
Superseding the earlier review: the frozen Must Fix list is closed at `85217986`. The rewritten history, strict ASCII E.164 exact-destination authorization, deep output sanitization, repository validation, and 168 offline tests were rechecked successfully.

PR #252, opened 2026-08-28
feat(holdfor-post-visit-followup): add post-visit check-in skill and review board app

carollpq:
## Summary

**Contribution Area: Agent Skills (`skills/`) and User-facing Apps (`apps/`)**, with one resource-list entry per area and one safety pattern.

A practice rings an older patient a few days after an appointment. Five bounded questions, asked at her pace. The answers, plus one verbatim sentence in her own words, land in a review queue. Nothing is booked. Only after a named member of practice staff reads the queue and grants a **Release** — carrying a date range, a time of day, an appointment mode, and the exact words that may be spoken — does a second call go out to the practice's own booking line and say them.

The second call is a phone call rather than a database write because the practice's booking system is reachable only by phone. The hold time is not removed; it is served to an agent instead of to an 80-year-old.

| What | Where |
| --- | --- |
| Agent Skill | [`skills/holdfor-post-visit-followup/`](skills/holdfor-post-visit-followup/) |
| Runnable app | [`apps/python/holdfor-board/`](apps/python/holdfor-board/) |
| Decisions | [`docs/adr/0001`–`0013`](docs/adr/) |
| PRD, glossary, hold probe | [`docs/holdfor-post-visit-followup/`](docs/holdfor-post-visit-followup/), `CONTEXT.md` |
| Resource entries | `README.md` (Skills, Apps, Safety patterns) and `apps/README.md` |

### The parts that may be reusable outside this workflow

- **A Release is a bounded authority, not an approval flag.** It records who released it, the dates that may be accepted, and the exact substring that may be spoken. A reviewer may shorten the patient's sentence and can never widen it — checked server-side against the transcript turn, not trusted from the form. [`docs/adr/0003`](docs/adr/0003-a-release-grants-a-bounded-authority.md)
- **The agent accepts; the board checks the acceptance.** A receptionist will not wait while a model deliberates, so the agent accepts an offer live, and that offer is then re-read deterministically after the call against the authorisation that allowed the call. "Only inside the envelope" becomes a property of the code rather than a request made of a model. [`docs/adr/0007`](docs/adr/0007-the-agent-accepts-the-board-checks-the-acceptance.md) — added under Safety patterns.
- **A refusal is not a missed call.** The call says aloud that hanging up ends the calls for good, so `DECLINED` is filed as `declined` rather than `not_reached`, and `may_redial()` returns `False` for every outcome. The platform's own repair logic offers a retry; we decline it. [`docs/adr/0006`](docs/adr/0006-a-refusal-is-not-a-missed-call.md)
- **The call that cannot ask is the only one that can be trusted.** The check-in call asks for no date of birth, no address, nothing resembling payment, and says so aloud — because a call asking an older person to confirm personal details is indistinguishable from a scam. The identifiers reception needs are supplied by the second call, never obtained from the patient.

## Type

- [x] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

- [x] Branch name, commit messages, and PR title follow `docs/git-naming-conventions.md`. (`python3 scripts/check_branch_name.py` passes on the branch.)
- [x] No secrets, tokens, private phone numbers, call recordings, or private transcripts are included. The only API-key string in the diff is `sk-ant-not-a-real-key` in a test. No database, `.env`, or live transcript is tracked; all three are gitignored.
- [x] Real-world side effects are clearly described. See *Side effects* below and the app README.
- [x] Phone numbers are masked in documentation and test fixtures unless they are clearly fictional. The twelve seeded patients use Ofcom-reserved `+447700900xxx`; `.env.example` uses `+441632960118/119`. Handsets are masked before reaching any page — the row dict has `phone_e164` deleted, so only a mask can render.
- [x] Recurring workflows include cancellation behavior. Nothing here recurs: no scheduler, no background job, no retry. One appointment produces at most one check-in call; one Release produces at most one rebooking call. Cancelling means not pressing the button, and the app README says so rather than implying a stop command exists.
- [x] Runnable code has a dry-run, fake-server, or no-call path by default. `FakeProvider` returns a stored transcript fixture, so the whole board runs offline with no credentials. 375 tests pass with nothing configured and place no call; the suite deletes `CALLE_LIVE` from the environment for every test, so a machine set up for a live call cannot turn the suite live.

## Side effects

- **Real outbound phone calls**, only under `CALLE_LIVE=1`, only for one named appointment at a time, and only on that appointment's Due Day with recorded consent. `CALLE_LIVE` is deliberately the one setting the `.env` loader skips by name, so a settings file can never arm a real call by being present. `run-due` refuses to run live at all — fanning out is right against fixtures and wrong against a phone network.
- **Every live placement is counted** in a `live_call` row before the call is submitted, including one whose outcome came back unknown, because it may still have rung her.
- **Writes to one local SQLite file** and one local transcripts directory. Deleting both is the whole cleanup.
- **Sends transcript text to the Anthropic API**, only when `ANTHROPIC_API_KEY` is set and `HOLDFOR_EXTRACT` is not `0`. What leaves the machine is a first name and how somebody says they have been — never a surname, a date of birth, or a number, because the agent promised aloud not to ask.

CALL-E is invoked at runtime, not merely referenced: `holdfor/providers.py` shells out to `calle auth status`, `calle call start --to-phone --goal [--region --language]`, then `calle call status --run-id` on a ten-second poll. Because `calle call start` has no way to carry a result schema, a live call returns a transcript rather than filled-in fields, so `read-back` re-reads the bounded answers out of the transcript through the same validation the call's own answers would have faced. That is documented rather than worked around.

## Known limits

Named here rather than left for a reader of `holdfor/rebooking.py` to find:

- **The Envelope Match checks day and time only.** Clinician and appointment mode are carried in the Release, spoken on the call, and recorded — but `match_offer` does not re-read them, so an offer on an authorised day and time passes the deterministic check even when reception named a different clinician. The two verified constraints are real; the other two currently rest on the prompt.
- **The day matcher reads bare digits, so a spoken clock time is read as a day of the month.** "Friday at 2:30" contributes `{2, 30}` to the day-of-month set and the offer is judged `outside`. The weekday filter runs after it, so the error direction is safe — the item returns to the board rather than a wrong day standing as booked — but a good slot gets flagged. The fix is to strip clock times before reading day numbers.
- **What has run live:** check-in calls and rebooking calls have both been placed through the `calle` CLI to handsets belonging to the team, with a team member playing the patient and a team member answering as reception. No real patient, relative, or member of the public has been called at any point. The routed demo fixture `fixtures/transcripts/demo/margaret-checkin.json` is the transcript one of those calls returned, kept verbatim; the practice, the patient, and every number in the app are fictional.

## One placement question for maintainers

This branch adds a root-level `CONTEXT.md` holding the workflow's glossary — the terms in bold with a capital that the SKILL.md, the references, the PRD, and the app README all lean on. Vocabulary is load-bearing in this workflow, so the file earns its keep, but the root of a shared repository may be the wrong home for one contribution's dictionary. Happy to move it to `docs/holdfor-post-visit-followup/` and update the five files that reference it if you would prefer that.


Ray-56:
Must Fix (frozen acceptance list):

- Remove the committed verbatim real-call transcript and every real-call-derived excerpt or artifact from the current tree. Replace them with newly authored synthetic fixtures and synthetic narrative.
- Rewrite the PR's unique history so that the real-call material is no longer reachable, including content-bearing commit messages that describe or excerpt those calls. A clean squash/rebuild from current `origin/main` is acceptable. Use repository-conforming Conventional Commit subjects when rebuilding; the current unique history also contains unsupported custom `merge:` subjects.
- Move the contribution-specific root `CONTEXT.md` under `docs/holdfor-post-visit-followup/` and update its references, as long-form contribution guidance belongs under `docs/`.

The privacy requirement applies even when team members role-played fictional people and used reserved numbers.

Ray-56:
Superseding review under the community policy adopted on 2026-09-11: The earlier requirements to purge synthetic/consenting-author test material or certify external production architecture are withdrawn where applicable. The accepted basic safety list remains closed; no actual secret or unauthorized private record was established by the residual item. Maintainer cleanup: Moved the glossary under scoped docs, repaired links, and resolved the app-index conflict.

PR #262, opened 2026-08-30
feat(apps): add cortex-call-brain two-tier learning phone agent

aryanchauhanoffical:
## Summary

A CALL-E outbound medication-adherence agent that grows a persistent, curated **two-tier memory**: a private sub-brain per caller plus a shared, anonymized master brain behind a **corroboration gate** (a fact needs ≥2 distinct callers before it is trusted). Adds a **human-in-the-loop prompt-approval** step, **"call me back later" continuity**, and fail-closed **consent / quiet-hours / idempotency / budget** guards.

Contributions:
- `apps/python/cortex-call-brain/` — runnable app + Streamlit console/brain-map, dry-run and offline paths, no secrets, fictional sample numbers
- `skills/adherence-memory-callback/` — portable `SKILL.md` + references
- `docs/two-tier-call-memory-and-curation.md` — reusable safety pattern

`scripts/validate_repository.py` passes; the app runs fully offline with a dry-run / no-call path by default.

## Type

- [x] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

### DIFF
--- a/apps/python/cortex-call-brain/.env.example
+++ b/apps/python/cortex-call-brain/.env.example
@@ -1,0 +1,25 @@
+# Copy to .env and fill in. NEVER commit a real .env.
+# Gemini powers transcript understanding + embeddings. Without a key the app
+# still runs fully offline (rule-based extraction + local hash embeddings).
+GEMINI_API_KEY=
+CORTEX_GEMINI_MODEL=gemini-flash-lite-latest
+CORTEX_EMBED_MODEL=gemini-embedding-001
+
+# CALL-E CLI passthrough (the `calle` CLI holds its own OAuth token)
+CALLE_SOURCE=skills_sh
+CALLE_INTEGRATION=skills_sh_skill
+CALLE_INTEGRATION_VERSION=0.1.0
+
+# Safety / policy knobs
+CORTEX_MAX_CALLS=50               # hard cap on real calls per run (bounds blast radius)
+CORTEX_BUDGET_USD=5.00            # money spend cap; only bites when cost-per-call > 0
+CORTEX_COST_PER_CALL=0.0          # set to a real estimate to make the money budget effective
+CORTEX_HASH_SECRET=               # keyed-HMAC secret for caller source ids; blank = per-DB random
+CORTEX_FACT_PROMOTION_MIN=2       # distinct callers to promote a fact to canonical
+CORTEX_SIGNAL_ALERT_MIN=2         # distinct callers to raise a pattern to staff
+CORTEX_SIGNAL_AUTO_APPROVE_MIN=4  # distinct callers to auto-apply a prompt change
+CORTEX_MIN_RECALL_HOURS=20        # idempotency window; no re-dialing within it
+CORTEX_QUIET_HOURS=21:00-08:00    # region-local; no calls inside this window
+CORTEX_ALLOWED_DIAL=              # comma-separated EXACT E.164 numbers; non-E.164 entries ignored; empty fails closed for live --execute dials
+CORTEX_REGION=IN
+CORTEX_LANGUAGE=English
--- a/apps/python/cortex-call-brain/.gitignore
+++ b/apps/python/cortex-call-brain/.gitignore
@@ -1,0 +1,12 @@
+# Never commit secrets or the brain (raw numbers, transcripts, private summaries)
+.env
+.env.*
+!.env.example
+*.db
+*.sqlite
+*.sqlite3
+cortex.db
+
+__pycache__/
+*.pyc
+.venv/
--- a/apps/python/cortex-call-brain/README.md
+++ b/apps/python/cortex-call-brain/README.md
@@ -1,0 +1,189 @@
+# Cortex Call Brain
+
+A runnable CALL-E app that gives an outbound phone agent a **persistent, curated
+two-tier memory** — so every call makes the next one smarter, and knowledge
+learned from one caller can safely help future callers.
+
+The reference workflow is a **pharmacy / clinic medication-adherence check-in**:
+the agent calls a patient, asks how they are getting on with a medicine, and
+returns structured notes. Cortex adds memory on top of that:
+
+- a **sub-brain per caller** (private summary + open items + callback context)
+- a shared **master brain** of facts and anonymized signals learned across
+  everyone, guarded so no single caller can poison it
+- a **human-in-the-loop approval** step before the agent starts proactively
+  asking about a newly learned pattern
+
+This app is a runnable demo, not an SDK or a supported product API.
+
+## What it does on a call (side effects)
+
+- Places **real outbound phone calls** through the CALL-E CLI when run with
+  `--execute` / a live campaign. Every call is a real side effect.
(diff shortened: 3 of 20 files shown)


Ray-56:
Superseding review for head `4f34bea`: exact strict E.164 allowlist matching is fixed.

Must Fix remaining:

- Complete phone fixtures outside the standards-reserved 555-0100–0199 range remain in the current tree and rewritten single-commit history. The safety test currently accepts any 202-555 prefix, so it does not catch them. Replace all complete examples with reserved fictional values, tighten the test to the actual reserved block, and rewrite the PR history.

aryanchauhanoffical:
Thanks for the detailed review. I’ve addressed all the points in the latest push (`c101044`) and also did another security pass on my end, which caught a few more things worth tightening up.

**Your findings**

1. Live calls now require an explicit `--execute` flag. By default, nothing gets dialed, including during dry-runs or previews.
2. Phone numbers are checked against strict E.164 formatting and, if configured, the `CORTEX_ALLOWED_DIAL` allowlist before anything reaches CALL-E. The validation happens both in `place_call` and again in the campaign as a safeguard.
3. Phone numbers are now masked in console output and the dashboard. I also replaced the old sample numbers with the reserved `+1-555-01xx` range, and the old values have been removed from the rewritten commit.
4. `wait_for_result` now takes the caller identity explicitly. The previous version passed `phone=None`, so I wasn’t able to reproduce the NameError, but this is now handled explicitly anyway.
5. If polling never reaches a terminal status and there’s no transcript, the result is marked as `inconclusive`. The campaign stops at that point for reconciliation instead of learning from it or moving forward.
6. Brain-map JSON is now serialized through a helper that escapes `<`, U+2028, and U+2029, so provider or LLM-generated strings can’t break out of the inline `<script>`.

**A few additional things I hardened during my own review**

* Caller source IDs now use keyed HMAC (`CORTEX_HASH_SECRET`, or a random per-DB secret) instead of a plain hash over a small keyspace. This makes the anonymization claim in the docs actually hold. Raw phone numbers also no longer appear in graph node IDs or synthetic run IDs.
* Added a hard per-run call limit via `CORTEX_MAX_CALLS`, defaulting to 50. This is important since the money budget doesn’t really protect us when the cost per call is set to 0.
* Added a `.gitignore` for the DB and env files so caller PII doesn’t accidentally get committed.
* LLM-extracted callback context is now clause-cut, flattened, and length-limited before it can make its way into a prompt.
* Pinned `vis-network` with an SRI hash and stripped HTML from graph tooltips.

`python3 scripts/validate_repository.py` passes, and the app also runs offline with the dry-run, seed, and unit checks.

Thanks again for the thorough review and for raising the bar on this.

aryanchauhanoffical:
Thanks, both points are addressed in `a5e0129`.

1. **Authorized destination + explicit consent.** Live dialing now requires the number to be on a non-empty `CORTEX_ALLOWED_DIAL` allowlist. If the allowlist is empty, everything fails closed. This is enforced through `authorized_dial(..., require_allowlist=True)` in both `place_call` and the campaign.

   I also changed the consent handling so that providing a `--phone` or `--demo` number yourself doesn’t count as consent anymore. For those cases, the operator has to explicitly pass `--consent`. Roster files have their own per-patient consent field.

   So `--execute` by itself can no longer call an arbitrary number. The destination has to be allowlisted and there also has to be explicit consent.

2. **Ambiguous outcomes now halt the campaign.** `ERROR`, `TIMEOUT`, and `UNKNOWN` are treated as ambiguous and marked `inconclusive`. The campaign stops for reconciliation instead of learning from the result or moving on to another caller. Normal terminal outcomes like `NO_ANSWER`, `VOICEMAIL`, and `CANCELLED` are still treated as valid terminal states.

**Verified:**

* Empty allowlist + `--execute` skips every destination as unauthorized.
* `--phone --execute` without `--consent` skips as no-consent.
* `ERROR` and `TIMEOUT` return `inconclusive`.
* `NO_ANSWER` and `COMPLETED` behave normally.
* `python3 scripts/validate_repository.py` passes.
* Offline paths including dry-run, seed, and unit checks all run cleanly.

aryanchauhanoffical:
Both addressed in a6b0d05 (history rewritten to a single clean commit).

1. **Allowlist: exact, validated, no prefixes.** `CORTEX_ALLOWED_DIAL` is now parsed into entries that must each be a strictly valid E.164 number — any invalid or broad entry (e.g. a bare `+1` prefix) is discarded — and destinations are matched by exact equality only. A broad value can no longer authorize arbitrary numbers, and an empty/invalid allowlist fails closed for live calls. Verified: `+1` and `+1202` authorize nothing; `+12025550100` matches only itself.

2. **Standards-reserved fictional data + clean history.** Every sample number (app, skill examples, README, docstrings, smoke tests) now uses the reserved `+1-202-555-01xx` range; no plausible real number is paired with a name anywhere, and no `+91` numbers remain. The PR is a single rewritten commit with only reserved data.

`python3 scripts/validate_repository.py` passes; offline paths (dry-run, seed, unit checks) run clean.

aryanchauhanoffical:
Follow-up in 4f34bea (self-initiated hardening on top of the above):

- Added `tests/test_safety_invariants.py` — 23 offline tests pinning every safety property (no-call default, exact-match allowlist, explicit consent, inconclusive-halts, corroboration gate, keyed-HMAC ids, `<script>`/tooltip escaping, free-text clamping, reserved-number guard). `bandit` (0 high/med) and `ruff --select F,B,E9` are clean.
- The sub-brain `summary` is now clamped and framed as untrusted data (same as `callback_reason`), and a CLI timeout / per-patient exception halts for reconciliation instead of crashing a live run.

aryanchauhanoffical:
Fixed in 811ffd4 (single rewritten commit).

Reserved numbers: replaced the remaining out-of-block test values, and tightened the hygiene test to the exact 555-0100–0199 block — it now scans the test file itself, so any number outside the reserved range fails the suite.

A pre-push self-audit also closed two gaps:
- **Signals now use the same distinct-source gate as facts.** `bump_signal` previously counted raw reports, so one caller reporting a symptom repeatedly could reach the auto-approve threshold and change the proactive script for everyone. It now tracks distinct hashed sources (verified: one caller × N reports stays at count 1 and never auto-approves), and the dashboard's "distinct callers" label is now accurate.
- **Right-to-forget now also clears the call log** (raw number, summary, transcript), not just the sub-brain — matching the docs' "complete erasure."

Also framed the callback opener as untrusted data (parity with the summary) and documented the single-writer assumption for concurrent campaigns.

26 offline invariant tests (incl. new coverage for both fixes), `bandit` (0 high/med) and `ruff --select F,B,E9` clean, and `validate_repository.py` passes.

Ray-56:
Superseding review for head `811ffd4`: the frozen Must Fix list is closed. The rewritten single-commit history now uses only standards-reserved complete phone fixtures, and the hygiene test enforces the exact reserved block across the app, skill, and test itself. Repository validation, 26 offline tests, Bandit, Ruff, and a credential-free dry run pass; remaining pseudonymization and single-writer limitations are non-blocking follow-ups.

PR #264, opened 2026-08-30
feat(apps): add clause-check-by-phone verification app

thibaudlepan77-svg:
## What this adds

`apps/python/clause-check-by-phone`, a small dependency-free app that turns a
clause a page states quietly into one question asked out loud, and reports only
when the voice contradicts the page.

## The rule it is built on

**Never make someone repeat on the phone what the page already says.** A call is
worth placing only if its answer can contradict the page. Six clause families
are considered worth a call. Anything else raises `NothingToAsk`, which is a
result and not an error, and the common one.

## Checklist

- English-only, no secrets, no tokens, no real phone numbers. Every sample uses
  `+33000000000`, which is fictional and never dialled.
- Host and provider stated, any Python 3.9 host, calls placed against
  `POST /v1/calls` with a Bearer key. The app contains no key and no client.
- Side effects described. It places a real call to a real person, and that is
  not reversible once the phone rings.
- Cancellation described. Nothing recurring, no schedule, no retry, no stored
  state between runs.
- **No-call path included.** `dry_run.py` prints the exact task and result
  schema without dialling. A developer holding twenty free calls cannot afford
  to tune a task by placing it.
- `python scripts/validate_repository.py` passes.
- `python scripts/check_branch_name.py --branch feat/clause-check-by-phone` passes.

## Tests

`python tests_bridge.py`, fourteen witnesses, no call, no network, no key.

Three are refusals, an unknown family, a clause with no quotation, and a
malformed number. Two guard the quotation itself, because it will be SPOKEN and
text meant for the eye tolerates a broken character that a speech engine
pronounces. One guards the answer `unknown`, which must never be read as a
contradiction, since an absence of answer is not a denial.

## On safety

The call task always states, in its first sentence, that the call is automated
and placed by a software agent, and a test enforces that. It asks one question
and only one, and says so twice. It never argues and never sells. Given there
is no sandbox and no echo number, the dry-run path is the only safe way to
iterate on the wording.

### DIFF
--- a/apps/python/clause-check-by-phone/README.md
+++ b/apps/python/clause-check-by-phone/README.md
@@ -95,6 +95,21 @@ The transport is a parameter of `place` and `collect`, which is why the
 witnesses can exercise every path through this file without a key and without
 dialling anyone.
 
+### What comes back, and what does not
+
+`collect` returns the identifier, the state and the structured answer. **It
+does not return the transcript, the recording or the destination**, and neither
+does `place`. The provider hands all three over next to the answer, this
+project uses none of them, and the obvious way to look at a result is to print
+it. A default that returns them puts a stranger's voice one `print` away from a
+terminal, a shell history or a CI log. `collect(..., raw=True)` returns the
+whole payload, for a caller who has decided to.
+
+For the same reason, a refusal names the destination **masked**, keeping the
+country code and the last two digits so you can still recognise which number
+you meant. A refusal is printed, so whatever it carries is what ends up on a
+screen. A short or malformed value is hidden whole rather than half revealed.
+
 ## Side effects
 
 **It places a real phone call to a real person.** That is the whole side
@@ -168,8 +183,15 @@ answer can change something, and an answer nobody can interpret changes nothing.
 
 ```bash
 python tests_bridge.py       # 26 witnesses, no call, no network, no key
-python tests_place_call.py   # 13 more, on the file that dials, same rule
+python tests_place_call.py   # 21 more, on the file that dials, same rule
 ```
+
+Eight of the twenty-one hold the line above, that nothing a caller is likely to
+print carries a number, a transcript or a recording. They check the masked
+refusal in both directions, that it hides the number and that it still says
+enough to recognise it, that a malformed value is hidden whole, that the
+trimming does not swallow the answer the verdict is computed from, and that a
+caller who asks for the raw payload still gets it.
 
 Seven of the twenty-six refuse something, an unknown family, a clause with no
 quotation, a nationally formatted number, a country code starting with a zero,
--- a/apps/python/clause-check-by-phone/README.md
+++ b/apps/python/clause-check-by-phone/README.md
@@ -182,8 +182,10 @@ and text meant for the eye tolerates a broken character that a speech engine
 pronounces. One guards the answer `unknown`, which must never be read as a
 contradiction, since an absence of answer is not a denial.
 
-Six more guard the schema validator, and every one of them is a schema built
-to be **refused**. A validator that rejects nothing proves nothing, so each
+Six more guard the schema validator. **Five of them are schemas built to be
+refused, and the sixth checks that every family still emits a conforming one**,
+because a validator that only ever refuses is as useless as one that only ever
+accepts. A validator that rejects nothing proves nothing, so each
 conforming case is paired with a hostile one, an unsupported composition, an
 open object, an enum with no way out, a reserved field name, and a required
 field that was never declared.
--- a/apps/python/clause-check-by-phone/README.md
+++ b/apps/python/clause-check-by-phone/README.md
@@ -59,6 +59,34 @@ prepared = call_task("+33XXXXXXXXX", "students only",
 verdict = contradiction(prepared, structured_result)   # None, or one sentence
 ```
 
+## Usage, the call actually placed
+
+`place_call.py` is the runtime path. It posts to `POST /v1/calls` on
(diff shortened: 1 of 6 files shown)


Ray-56:
Superseding review for head 9690133: strict E.164, exact recipient authorization, a safe dry-run, valid fictional examples, and accurate client documentation are now present. The offline repository checks and 39 focused tests pass.

Newly discovered concrete Must Fix privacy blocker:
- The refusal path returns the full destination, and collect() plus the documented JSON-print snippet can expose the raw provider payload, including phone or transcript data. Mask the refusal output and return/document a sanitized bounded result instead of raw provider data.

thibaudlepan77-svg:
Thank you, all three were real and all three are fixed in 39a2081.

**Strict E.164 and an authorised recipient.** `call_task` now matches
`\+[1-9]\d{6,14}` with `re.fullmatch`. The old pattern took a leading `+0`, and
because it was `re.match` anchored on `$` it also took a trailing newline, so a
number read from a file passed. Both are covered by witnesses now, along with
the fifteen digit ceiling.

Authorisation is separate from format, so it is enforced separately.
`place_call.place` refuses any recipient absent from `CALLE_ALLOWED_RECIPIENTS`,
before the request is built, and an absent or empty list authorises nothing.
That check sits in the file that opens the socket rather than in the bridge,
since the bridge builds text and never dials. Three witnesses hold it, an empty
list, an unlisted number, and a list of several numbers.

**The snippets.** You are right that `+33XXXXXXXXX` was not runnable, the
bridge rejects it. Every sample now uses `+447700900123`, inside the block
Ofcom reserves for drama, `07700 900xxx`. It is valid E.164, so the snippets
run as written, and it is reserved, so it reaches nobody. The malformed-number
witnesses keep values that are not numbers at all.

**The false claim.** That one was the worst of the three and it was mine to
catch. The README said the contribution contains no client while `place_call.py`
is a credential-bearing HTTP client that can place a real call. It now says so
in the same paragraph, and says the live path is opt-in twice, once by the key
and once by the recipient list. I also corrected two counts that had drifted,
the number of refusals and the number of witnesses.

The suite is 39 witnesses now, 26 in `tests_bridge.py` and 13 in
`tests_place_call.py`, all offline, no key, no network, no call placed.

thibaudlepan77-svg:
Thank you, the privacy blocker was real and it is fixed in 8521eb7.

**The refusal path.** You are right that it returned the full destination, and
a refusal is printed, so a real number ended up in a terminal, a shell history
or a CI log. `masked()` now keeps the country code and the last two digits and
hides the rest, which is enough to recognise which number you meant and not
enough to dial it. A short or malformed value is hidden whole rather than half
revealed, because half of a wrong number is still half of a number. Two
witnesses hold it in both directions, one that the number does not appear and
one that what remains still identifies it.

**`collect` and the documented snippet.** That was the worse of the two. The
provider hands back the recording, the transcript and the destination next to
the answer, `collect` returned all of it, and the README told you to
`json.dumps` the result. This project uses none of the three. What comes back
is now bounded to the identifier, the state and the structured answer.
`raw=True` returns the whole payload for a caller who has decided to take it,
and the printed snippet says in one line what it does not print.

One detail that needed care. The verdict is computed from `structured_result`,
so I read the answer before trimming rather than after. A witness fails if that
order is reversed, otherwise the fix would have quietly emptied the only output
this project has.

The same bounding applies to a provider refusal, which used to interpolate the
whole payload into the exception message.

The suite is 47 witnesses now, 26 in `tests_bridge.py` and 21 in
`tests_place_call.py`, eight of them new and all on this. Offline, no key, no
network, no call placed. `scripts/validate_repository.py` passes.

Ray-56:
Post-merge correction for head 8521eb76:
- The frozen privacy blocker is not fully closed. The default bounding helper removes only top-level keys, so normal nested recipients/attempts can still retain full destinations and transcript turns. Return a deliberately constructed allowlist containing only call ID, status, and structured result, with a realistic nested privacy fixture.
- Newly discovered concrete functional blocker: the live request omits recipients, so the repository's fake CALL-E API rejects it and the advertised live workflow cannot place a call. Include the explicitly authorized recipient and cover the request contract offline.

The contributor branch does not permit maintainer edits, so I could not apply these bounded fixes there.

thibaudlepan77-svg:
Thank you, both were real and both are fixed in #330, which allows maintainer edits.

**The bounding helper.** You are right that it only removed top-level keys, so a payload shaped the way `apps/python/ringdown/fake/calle_server.py` shapes one came back whole, destination and transcript turns included. It now returns an allowlist, the call id, the status and the structured result, and nothing else. A list of names to strip is also wrong about every field a provider adds after the list was written, and it is wrong in the direction that leaks. Five witnesses read a realistic nested payload and assert that neither the number nor a transcript turn survives, and they fail against the previous helper.

A provider refusal needed its own path, since it carries none of those three fields and would have printed as an empty dict. It keeps the few fields that say what went wrong, shortened, with long runs of digits masked, because a provider that refuses a request usually quotes the request back and a refusal is printed.

**The missing recipient.** Also right, and it made the advertised live workflow impossible rather than merely degraded. The create request now carries `recipients: [{"phones": [...]}]` with the number that just passed the authorisation check, never another. Two witnesses cover the request contract offline, one on the recipient and one holding the body to the field names the fake server accepts.

`python3 scripts/validate_repository.py` passes, twenty-eight witnesses in `tests_place_call.py` and twenty-six unchanged in `tests_bridge.py`, still no network, no key and no call.

PR #269, opened 2026-08-31
feat(creditcall): add invoice exception call handoff

Jonny7171:
## Summary

Adds CreditCall, a focused CALL-E workflow that turns a fictional $200 invoice price exception into one disclosed, human-approved phone task. The default demo calculates and prints the call plan without placing a call.

## Safety and side effects

- requires a named human approver and recipient consent before planning
- requires explicit E.164 destination numbers
- requires a second explicit `--confirm RUN` step before starting
- discloses automation and forbids supplier impersonation
- handles carrier call screening, repeats disclosure after a person joins, and waits before continuing
- masks phone numbers in terminal output
- uses a local single-use reservation to block duplicate starts
- persists only local reservation state and a timestamp, never raw provider output
- has no schedules or automatic retries
- includes status and uncertain-start recovery commands
- never requests payment details, credentials, personal information, or a commitment

## Privacy

Only synthetic dry-run evidence is committed. The repository contains no real call artifacts, recipient data, call identifiers, or transcripts.

## Verification

- `node --test` in `apps/typescript/creditcall` (8 passing)
- `node src/cli.mjs demo`

Repository validation passes.


Ray-56:
Superseding review for head b2e41e9: all frozen Must Fix items are closed. The branch history no longer contains a real-call proof artifact or live claim, E.164 is documented, and the ledger stores only a bounded submitted-state record without raw provider results. Maintainer-assisted cleanup resolved the single README conflict with current main while preserving both app entries. Repository validation, eight offline tests, and the credential-free demo pass. No current Must Fix remains.

Jonny7171:
Addressed all three blockers in the rewritten branch history:

- removed the real-call artifact and all live-run claims from the PR history
- documented the E.164 requirement in the safety boundary and live command instructions
- stopped persisting raw CALL-E results; the ledger now stores only submitted state and a timestamp, with a regression test covering phone and transcript data

Validation is clean: 8 app tests pass, the dry demo passes, and the full repository validator passes.

PR #274, opened 2026-08-31
feat(apps): add shift safety call agent

saka2279:
## Summary

Adds Shift Safety Call Agent, a runnable Python reference app that turns fictional post-shift safety check-ins into structured records for human review.

## What it demonstrates

- deterministic Fake Provider with four fictional scenarios
- local FastAPI review UI
- explicit separation of call status and human review disposition
- preservation of unknown and unavailable values
- English CALL-E reference task and structured result schema
- optional guarded one-shot CALL-E self-call integration
- localhost-only default reviewer path with no CALL-E credentials or real call required

## Verification

- Python 3.12
- 144 tests passed in the full local verification environment
- clean-room installation passed without the optional CALL-E SDK
- official repository validator passed
- branch-name validation passed
- `git diff --check` passed

The public English CALL-E task has not been live-call validated. The default reviewer path uses fictional Fake Provider data and places no calls.


Ray-56:
Must Fix before merge:
- The README makes a specific empirical claim about a private live Japanese validation and result fields while explicitly withholding the implementation and evidence. This is materially unverifiable for a public reference contribution. Remove or clearly recast the claim as unverified background; do not add a real-call artifact.

saka2279:
Thanks for flagging this. I removed the unverifiable private live-validation claim and kept the public validation boundary limited to the Fake Provider, offline tests, and no-call reviewer path. No private or real-call artifact was added.

PR #281, opened 2026-09-01
feat(apps): add AgroVoice cocoa farmer data-collection app

Steven3-art:
## Summary

Adds AgroVoice, a customer/farmer callback app built for the CALL-E: Your Code Is Calling hackathon. It places AI voice calls to smallholder cocoa farmers in Cameroon, asks a few structured questions (village, harvest volume, logistics issues), and extracts the answers via CALL-E's native `result_schema` — never guessed or fabricated.

Note: the actual call script (`TASK_TEMPLATE` in `agrovoice_pipeline.py`) intentionally remains in French with authentic Bulu-language greetings ("Mbolo"/"Akiba"), since real Cameroonian cocoa farmers speak French and local languages. All surrounding documentation, code comments, and variable names are in English..

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

- [ ] `python3 scripts/validate_repository.py` passes.


Ray-56:
Superseding review for head `25c3d98`: preview-by-default, the manual live allowlist, per-recipient `CALL`, and stop-on-ambiguity behavior are fixed.

Must Fix remaining:

- A complete non-reserved phone example remains in the current source and older unique commits/deleted files still contain complete plausible numbers. Replace current examples and rewrite PR history. The live demo-number option also bypasses the authorization allowlist and submits example-only reserved numbers to the provider; keep them preview/mock-only or require the same exact authorization gate.
- Preview still prints arbitrary input in full, and the full provider response is printed/persisted as `structured_result_raw`, which may contain phone or transcript data. Mask preview output and minimize/redact stored provider data.
- The French/Bulu task prose remains without the explicit maintainer approval required by AGENTS.md. Keep repository-facing content in English unless that approval is granted.

Steven3-art:
Thanks for the thorough review — all five points are addressed as of the latest commits, except #5 below which needs your explicit call.
1. Dry-run default — Fixed. The app now opens on a preview-only mode (option 0) that requires no API key and places no call. A live call additionally requires the operator to type the literal word `CALL` per recipient — pressing Enter cancels.
2. E.164 + allowlist — Fixed. Replaced the ad-hoc "prepend a country code" logic with strict E.164 regex validation plus a local, gitignored `authorized_numbers.txt` allowlist — a number must pass both checks before any live call. Demo numbers now use the NANPA-reserved fictional block (+1-202-555-01xx). The old `demo_numbers.py` (which still had the non-standard fictional numbers) has been deleted rather than left stale.
3. Number/state exposure — Fixed, and I found a real bug while fixing it: the previous version was embedding the real number in both the natural-language task text and the request's `metadata` field. It's now passed only once, in the structured `recipient.phone` field CALL-E's API expects for dialing. Every other surface (console, SQLite, Excel) now stores/prints only `mask_number()`'s masked form.
4. Ambiguous outcomes — Fixed. Timeouts and ambiguous request failures now raise a dedicated `AmbiguousOutcome` exception that halts the entire batch run for manual reconciliation against the CALL-E dashboard, instead of retrying or advancing to the next recipient.
5. Non-English task text — Not yet changed, and I'd like to ask for that approval rather than assume it. The call script itself (what CALL-E actually says on the phone) is in French with two short Bulu-language phrases ("Mbolo" / "Akiba"), because the real recipients are Cameroonian cocoa farmers who speak French and local languages — an English-language call would not be understood by the people this app is built for. All repository-facing content (code, comments, README, this PR) is in English. Given AGENTS.md's approval process for this kind of exception, could you confirm whether this is acceptable, or let me know what you'd need to see for it to be? Happy to add a machine-readable `language: fr` field to the task metadata, or an explicit comment block flagging it, if that helps reviewers/tooling catch this case going forward.

Ray-56:

Remaining Must Fix only:

- apps/python: remove the demo-number live authorization bypass; every live destination must pass the explicit authorization gate.
- apps/python: mask input preview and provider-derived output rather than printing arbitrary inputs and structured_result_raw.
- Call task French/Bulu prose: obtain the maintainer's requested language exception or translate repository-facing task text per AGENTS.md.

Steven3-art:
Thanks for the updated bounded list — all 3 items addressed in the latest commit:
1. Demo-number bypass removed. There's now a single live-call path; every destination (fictional or real) must pass the same allowlist gate. No code-level exception for "safe-looking" numbers — the README now documents that operators add NANPA-reserved numbers to their own local `authorized_numbers.txt` like any other number.
2. Masking extended to preview + storage. Preview mode now masks typed input before printing it (previously echoed raw). Persisted data no longer includes the full provider response — `sanitize_state_for_storage()` keeps only `call_id`, `status`, and `structured_result`; the DB column was renamed `provider_response_safe` to reflect that.
3. Task text translated to English. Since the exception request didn't get a response in time for the hackathon deadline, I translated it rather than continue waiting. The original French/Bulu version (the one actually validated in a real successful test call) is preserved as a documented comment directly above `TASK_TEMPLATE` and in the project README, flagged clearly as the intended production language for the real target users (Cameroonian cocoa farmers) — this English version satisfies the repository-content rule but is not what I'd actually run for a real farmer call.
Let me know if anything here needs another pass.

Ray-56:
Superseding review for head d914308a under the community/demo policy. Only these current Must Fix items remain; previous items not listed here are closed:

- apps/python/agrovoice/agrovoice_pipeline.py:143–154,279–288 still copies and prints structured_result verbatim, so phone-bearing logistics text reaches terminal, SQLite and Excel. Mask the selected provider text before display/storage/export.
- apps/python/agrovoice/agrovoice_pipeline.py:75–84 still preserves full French repository-facing task prose without the requested maintainer exception. Remove/translate the prose or obtain the bounded exception.

Steven3-art:
Thanks — both items addressed in the latest commit:
1. Free-text redaction. Added `redact_phone_like_patterns()` + `sanitize_structured_result()`, applied once immediately after extracting `structured_result` from the provider response — every downstream use (console, SQLite, Excel) now works off the same sanitized copy. While implementing this I found the storage path was pulling the field from the raw `final_state` again rather than the already-sanitized value, which would have defeated the fix for the persisted copy specifically — `sanitize_state_for_storage()` now takes the sanitized result as an explicit parameter instead of re-reading it.
2. French prose removed from the file. The reference comment block has been deleted; `TASK_TEMPLATE` is English-only, with a one-line pointer to the project's Devpost story rather than reproducing the French text in-repo.
Let me know if anything else needs a pass.

Ray-56:
Superseding the earlier review: The previously requested English repository-facing guidance and public-output masking fixes are closed. No current Must Fix remains under the community/demo review policy; optional hardening remains non-blocking.

PR #284, opened 2026-09-02
feat(apps): add accessline venue accessibility verifier

BuckNorris0506:
## Summary

Adds AccessLine, a consent-first Python app that uses the CALL-E Developer REST API to verify three venue accessibility facts by phone and return schema-valid structured JSON with explicit uncertainty preservation.

## Contribution area

Apps → Python (`apps/python/accessline/`)

## CALL-E use

- `POST /v1/calls` with `recipient_result_schema`
- Poll `GET /v1/calls/{call_id}` until terminal status
- Live path is opt-in and separately gated (see Safety)
- Demo/fixture evidence is synthetic functional test data only; it does not prove real-world telephony reliability

## Safety / no-call default

- Mock CLI path is default for judges (`--mode mock`)
- Live path requires: `CALLE_API_KEY`, consent, strict E.164 destination, exact authorized destination match, and fresh per-run live intent
- Official CALL-E HTTPS origin is pinned; arbitrary base URL overrides are rejected before bearer credentials are attached
- Phone numbers are masked in normal output/artifacts; transcript bodies are omitted by default
- Automation disclosure included in call script
- No sensitive personal questions

## Validation

```bash
python3 -m unittest discover -s apps/python/accessline/tests -q


Ray-56:
Must Fix for head `15f8348`:

- The current tree and sole unique commit include an explicitly real-call result artifact and proof claim. Remove the artifact, replace it with a clearly fictional fixture, and rewrite the PR history so the real-call material is no longer reachable.
- The live path accepts an unchecked phone plus a stored consent boolean. Enforce strict E.164, exact authorized-destination matching, and clear fresh per-run live intent before the provider call.
- `CALLE_BASE_URL` may point to any origin and the bearer credential is sent there. Pin or strictly allowlist the official HTTPS CALL-E origin.
- User-facing artifacts and preview serialization expose the complete phone, and artifact serialization can include a transcript. Mask phone output and do not surface live transcripts by default.

Ray-56:
Superseding review for head `6350e1a`: the rewritten head closes the frozen real-artifact/history, live authorization, origin, and output-privacy blockers. Repository validation and 89 credential-free offline tests passed after installing the declared dependency. No current Must Fix remains.

PR #286, opened 2026-09-02
feat(apps): add casechaser open-case chasing app

agentaiventure-dot:
## Summary

Adds **CaseChaser** (`apps/python/casechaser`): a runnable app that chases an open claim, refund, warranty repair, delivery, or billing case to closure by phone with CALL-E.

Every call is planned from a case ledger and turned into a dated, quoted **commitment record**. When the promised date passes without resolution, the commitment is marked broken and the next call climbs a fixed **escalation ladder** (agent, supervisor, written complaint, regulator). Any offer, denial, or identity refusal stops at the customer: the case becomes `needs_human`, the exact words are shown, and the customer's decision is carried into the next call's history. A masked **evidence pack** (every call, quote, reference, and commitment status) is one command away for the written complaint that ends most of these cases.

Safety is enforced in code, not just in the prompt: ten named suppression reasons (quiet hours in the company's own time zone, daily and total call caps, minimum spacing, pending promise, pending decision, closed case, waiting on customer, blocked or malformed number), a closed and fully-required `result_schema` that rejects anything off-contract, an `Idempotency-Key` per case per day, masked numbers everywhere, and hard boundaries appended to every task (no payment or identity secrets, no accepting or declining amounts, no account changes, no legal threats).

Default paths are no-call: `preview` prints the exact task; `fixture` runs the real client against a local fake of the Calls API with seven terminal-call fixtures covering every workflow branch. Live mode needs `--mode live --yes` and a key in `.env`; the dashboard cannot place live calls.

## Type

- [x] New runnable app

## Checklist

- [x] Phone numbers are masked in documentation and test fixtures unless they are clearly fictional (fixtures use the reserved +1555 range).
- [x] Recurring workflows include cancellation behavior (`docs/scheduler.md`).

Tests: `python3 -m pytest` in `apps/python/casechaser` (20 tests, offline, Python 3.9+, standard library only).


Ray-56:
Must Fix for head `b8df52a`:

- The live destination gate accepts only “plus followed by digits,” has no exact authorized-destination record, and `--force` bypasses invalid/blocked destinations and every policy hold. Enforce valid region-aware E.164 plus exact authorization in live mode, prohibit live `--force`, and replace the bundled malformed live-callable numbers with standards-reserved valid fixtures.
- `CALLE_BASE_URL` accepts any or insecure origin while the bearer credential is sent there. Pin or strictly allowlist the official HTTPS CALL-E origin.
- The scheduler recipe embeds live `--yes`, so it has no fresh per-run operator intent. After an unusable or unknown result the app schedules another chase, and its idempotency key changes by day; an accepted-but-unrecorded call can therefore be duplicated by a later unattended run. Stop for reconciliation after ambiguity and require a separately authorized scheduled-run artifact.
- The dashboard can bind beyond loopback without authentication while exposing cases, transcripts, evidence, and mutation endpoints. Its list view also inserts unescaped case fields into `innerHTML`. Hard-bind it to loopback or add authentication, and escape all rendered fields.
- The claim that the closed schema rejects anything off-contract is materially false: validation accepts extra fields and wrong types for most required fields, which can then crash processing or drive case state. Fully validate the contract before state changes, or narrow the claim and fail closed.
- The current tree and both unique commits contain realistic-looking case, policy, and account identifiers rather than unmistakably fictional placeholders. Replace them and rewrite the PR history.

Ray-56:
Superseding review for head `89ecd7c`: the rewritten head closes the frozen destination, credential-origin, scheduled-intent/ambiguity, dashboard, schema, and fictional-identifier blockers. Repository validation and 27 credential-free offline tests passed. No current Must Fix remains.

PR #292, opened 2026-09-03
feat(apps): add approval-gated payout policy verifier

Kndll33:
## Summary

Adds `apps/typescript/payoutproof`, a small Node app with one declared phone-number validation dependency that converts an owner-reviewed public reward opportunity into one bounded CALL-E payout-policy inquiry.

The app:

- requires a publicly listed official business contact attestation;
- allows only enumerated public-policy questions;
- masks the phone number in preview output;
- generates deterministic packet, idempotency, and approval identifiers;
- defaults to no-call preview/plan behavior;
- requires both `PAYOUTPROOF_ENABLE_LIVE=1` and an exact approval ID before `call start`;
- treats verbal answers as non-contractual and requiring a written source.

## Verification

- `npm test --prefix apps/typescript/payoutproof` — 9/9 pass
- `python3 scripts/validate_repository.py` — pass
- `python3 scripts/check_branch_name.py --branch feat/payoutproof` — pass

No real call was placed. The sample uses a reserved fictional NANP number, and no credentials or personal data are included.

## Authorship and scope

TenK, a disclosed autonomous AI agent, created this code, tests, and documentation under the `Kndll33` controller account. The README states this explicitly. No CALL-E sponsor, payout provider, or program has endorsed the app or confirmed a payout policy.


Ray-56:
Must Fix for head `1aa2fd3`:

- The phone validator makes the leading plus sign optional, so non-E.164 destinations can reach the live command. Require a strict, region-valid E.164 number.
- `normalizeResult()` emits the raw provider object verbatim, which can expose full phone numbers or transcripts. Return only bounded, schema-validated, masked fields by default.

Kndll33:
Fixed in `4372b7a`:

- destination validation now requires a leading `+`, validates the number with `libphonenumber-js`, and rejects country/`region` mismatches;
- result normalization now allowlists status values, drops the complete raw provider payload, and returns only bounded fields plus the already-masked destination;
- added regression tests for missing `+`, region mismatch, and provider PII/transcript removal.

Verification: 9/9 app tests pass; repository validator and branch-name check pass.

Ray-56:
The two Must Fix items from the review of head `1aa2fd3` are resolved in `4372b7a`: destination validation is now strict and region-aware, and normalized output is bounded and masked without retaining the raw provider payload. The focused 9-test suite and repository validation pass; this supersedes the earlier blocker comment.

PR #294, opened 2026-09-03
feat(appointment-call-confirm): add appointment confirmation call skill

lazarusmodv1:
﻿## Summary

Adds `appointment-call-confirm`, an Agent Skill that batch-confirms
upcoming appointments by phone through CALL-E: dry-runs the list,
places one call per recipient (serially, never parallel), asks each
to confirm/reschedule/cancel, and writes a structured result back per
row.

Service businesses that book time slots (clinics, salons, repair
shops, tutoring, delivery windows) lose revenue to no-shows. The usual
fix is a staff member manually phoning tomorrow's list — repetitive
and the first thing skipped on a busy day. This automates exactly
that one workflow.

## How it uses CALL-E

Wraps the Developer API directly: `POST /v1/calls` with a
`result_schema` describing the exact structured shape we want back
(`confirmed` / `needs_reschedule` / `declined` / `no_answer` /
`voicemail` / `unclear`), then `GET /v1/calls/{call_id}` polled to a
terminal status. No mocking.

## Safety

- Never places a call without explicit `--confirm`; dry-run is the default.
- Serial calling only, one recipient at a time.
- Phone numbers masked in all output; full number only ever sent to CALL-E.
- Never guesses a `confirmed` result from an ambiguous response — reports `unclear` instead.
- Requires a separate `--authorized-numbers` file before any live call; a number not listed is never dialed.
- Full contract in `references/safety.md`.

## Testing
cd skills/appointment-call-confirm
python3 scripts/place_confirmation_calls.py --in assets/sample_appointments.csv
prints the dry-run with no calls placed and no dependencies required.
Add `pip install -r requirements.txt`, `--confirm`, and
`--authorized-numbers` (plus a real CALL-E key and consented number)
to place a real call.

## Type

- [x] New skill

## Checklist

- [ ] Recurring workflows include cancellation behavior. *(N/A — this skill runs on-demand, not on a recurring schedule; see "When not to use" in SKILL.md.)*


Ray-56:
First complete review for head fd9862c. Frozen Must Fix list:\n- Add a documented dependency/install path so a clean checkout can run the default credential-free dry run; the script currently imports an undeclared third-party package at startup.\n- Replace plausible full phone/contact examples with region-valid standards-reserved fictional data and rewrite the PR history.\n- Strictly validate each destination as region-bound E.164 and require that it is authorized before any live request.\n- Pin or allowlist a trusted HTTPS CALL-E origin before sending the bearer API key.\n- Halt the entire batch after a create ambiguity, missing call ID, polling timeout, or other uncertain provider outcome; do not advance to another recipient until reconciled.\n- Document cancellation clearly, including that interruption prevents future calls but may not cancel an already accepted in-flight call, plus the supported cancellation path.

lazarusmodv1:
﻿Thanks for the thorough review — addressed all six in `33fd353`:

1. **Undeclared dependency** — `requests` is no longer imported at
   module scope; it's lazily imported only when `--confirm` runs, with
   a clear error pointing to the new `requirements.txt`. Dry-run now
   needs zero installs on a clean checkout.
2. **Fictional numbers + history** — replaced with the NANP-reserved
   555-0100-0199 block, correctly formatted E.164 (my first attempt
   was actually malformed — good catch). Force-pushed an amended
   commit so the real number isn't in reachable history.
3. **Strict E.164 + authorization** — added per-region digit-length
   validation before any request, and a new required
   `--authorized-numbers <file>` gate, separate from `--confirm`. A
   number not explicitly listed is never dialed.
4. **Pinned CALL-E origin** — the script now refuses to send the
   bearer key anywhere except `https://api.heycall-e.com`, even if
   `CALLE_BASE_URL` is overridden.
5. **Halt on uncertain outcomes** — local/network errors, missing call
   IDs, and poll timeouts now halt the entire batch immediately
   instead of continuing. Clean terminal outcomes (including a normal
   `declined`/`no_answer`) still continue — only genuinely unknown
   outcomes stop things. Also switched to a deterministic idempotency
   key so a naive re-run after a halt has a shot at CALL-E
   deduplicating rather than double-dialing.
6. **Cancellation** — checked CALL-E's documented endpoints directly;
   there is no cancel/delete endpoint today. `references/safety.md`
   now says so explicitly: interrupting the script stops future calls
   in the batch but cannot cancel one already created, and there is
   currently no self-serve cancellation path.

Let me know if any of these need a different shape.

Ray-56:
Correction to my merge review: the rewritten contribution and squash now on main are sanitized, and the bounded region-binding fix in `b3b3dde` passes its four offline regressions. However, GitHub still serves the pre-rewrite commit `fd9862c`, including the old contact example in `skills/appointment-call-confirm/assets/sample_appointments.csv`. I should have verified that object before merging.

The historical privacy Must Fix remains: complete GitHub-side removal or dereferencing of the affected pre-rewrite PR commits/refs (GitHub Support may be needed), then verify the old commit and file URLs no longer resolve. Do not repost any contact values. The setup, authorization, origin, ambiguity-halt, and cancellation fixes remain accepted.

lazarusmodv1:
Flagging directly: I opened a separate PR (#310, under my other GitHub
account Lazvyn) for what became an extended review/hardening cycle on
this same appointment-call-confirm skill. I've since copied all of
that hardening back into this PR, which I'm treating as my one and
only submission for this skill. #310 is now closed. Wanted this on
the record rather than have two open threads for the same underlying
work.

PR #297, opened 2026-09-03
Add RingBack - appointment phone-campaign agent with first-yes cascade (Python, dry-run by default, bilingual FR/EN)

Ludo321:
…e (Python, dry-run by default)

## What this app does

RingBack is a phone-campaign agent for solo professionals (physiotherapists,
barbers, music teachers…). It keeps a small appointment book (SQLite) and drives
batch calls from a local web UI: build the list, review it, start it by hand,
cancel any queued call before it runs, and get one structured outcome per call
plus the transcript. Outcomes are written back to the appointment book.

**Five kinds of campaign**, each with its own conversation rules:

| kind | what the agent does |
|---|---|
| missed appointment | calls back someone who did not come, offers a new slot |
| appointment reminder | confirms attendance, offers other dates if not |
| appointment confirmation | confirms a booked slot |
| appointment rescheduling | moves N appointments out of a range — **everybody is called** |
| **freed slot (cascade)** | one slot to fill: calls down a list and **stops at the first yes** |

The cascade is the interesting one. A cancelled appointment frees a slot; RingBack
calls candidates one by one and the first firm yes ends the campaign — nobody else
is disturbed. Optionally it shifts the following appointments up, so the freed time
is not merely filled but recovered.

Python 3.12, **standard library only** (no dependency to install).

**The product is bilingual, French and English — one button in the banner**, no
reload, no JavaScript required. The switch carries everything, not only the
screens: the agent's briefing sent to CALL-E, spoken dates, the voice itself
(`region`/`locale` → `FR`/`fr-FR` or `GB`/`en-GB`), the demo data set, and every
error message. Reviewers who do not read French can evaluate the whole product,
real call included, in English.

## Compliance with the contribution rules

- **Dry-run by default ✔** — the default call client is a full simulator (no
  network access, scripted conversations, fake numbers). The real client sits
  behind **three independent locks**, none reachable from the web UI:
  1. it refuses to construct without the `CALLE_API_KEY` environment variable;
  2. the scheduler's `dry_run` flag (default `True`) must be lifted via the
     `--appels-reels` command-line flag;
  3. an explicit confirmation (typing `APPELER` at a keyboard prompt) is required
     at **every** launch — never persisted.
  Tests prove that while any lock holds, **zero HTTP requests leave the machine**.
- **Masked phone numbers ✔** — numbers never appear in clear in the UI, the logs,
  or the audit file (`+33 6 •• •• •• 42`); masking is enforced in the data-access
  layer and asserted throughout the test suite. The number is also deliberately
  never embedded in the task text sent to the agent (recipient field only).
- **Cancellation logic ✔** — any queued call can be cancelled before execution
  (`Planificateur.annuler`), from the UI or the API; cancelled calls are recorded
  as such and excluded from outcome tallies.
- **Side effects documented ✔** — two on-disk artifacts, both documented in the
  README: `donnees/ringback.db` (SQLite, created on first run, demo rows only when
  empty — delete to reset) and `donnees/audit_appels_reels.jsonl` (one line per
  attempted REAL call: timestamp, masked number, status — written on success and
  failure alike).

## How to run

```
python lancer_serveur.py          # web UI on http://127.0.0.1:8770 (simulation)
python -m unittest discover -s tests -v   # 1202 tests, all green
```

## Bilingual, French and English

The UI ships in both languages — one button in the banner, no reload, no
JavaScript required. The product itself is written in French (its users are
French solo professionals); English is applied to the **finished page**, at the
two lines where HTML becomes bytes, so the French product is never traversed
and cannot regress.

The switch carries everything, not just the screens: the agent's briefing sent
to CALL-E, spoken dates, the voice (`region`/`locale` → `GB`/`en-GB`), the demo
data set, and every error message. Reviewers who do not read French can run and
evaluate the whole product in English, including a real call.

## Testing

1202 unit tests (stdlib `unittest`): database layer incl. masking, on-disk
persistence, no-show detection, input validation, CSV import, web server exercised
over real HTTP, and the real client integration-tested against a **local fake
CALL-E API** (`POST /v1/calls` + polling + `result_schema`) — timeouts, failure
statuses, and non-conforming results rejected **before** any database write. The
real API is never contacted by the test suite.


Ray-56:
Must Fix for head a453461d:

- Remove plausible full contact identifiers from examples, fixtures, exported views, and every unique commit, then rewrite the affected PR history using reserved fictional data. The UI/CSV path must not expose full destinations.
- Keep repository-facing content English-only; the contribution currently contains extensive French-only prose, comments, reports, and interface source.
- Do not send the CALL-E bearer credential to an arbitrary configured origin. Restrict the live API endpoint to an approved secure origin or fail closed.
- Correct the materially false safety/validation claims: the committed report records a failed real-client guard while claiming all checks passed, and the documented masked-only audit can persist raw provider content.

Ludo321:
> Must Fix for head [a453461](https://github.com/CALLE-AI/awesome-phone-call-agents/commit/a453461d210083b3d6fa933e705e9986eea23493):
> 
>     * Remove plausible full contact identifiers from examples, fixtures, exported views, and every unique commit, then rewrite the affected PR history using reserved fictional data. The UI/CSV path must not expose full destinations.
> 
>     * Keep repository-facing content English-only; the contribution currently contains extensive French-only prose, comments, reports, and interface source.
> 
>     * Do not send the CALL-E bearer credential to an arbitrary configured origin. Restrict the live API endpoint to an approved secure origin or fail closed.
> 
>     * Correct the materially false safety/validation claims: the committed report records a failed real-client guard while claiming all checks passed, and the documented masked-only audit can persist raw provider content.

@Ray-56 
Thank you for the review — all four points were real defects, and two of them were things none of our own nets could see. I appreciate the vigilance; it made the product measurably safer.

Here is what changed.

1) Plausible contact identifiers

Every demo, fixture and example number now comes from the six roots ARCEP reserves for fiction (01 99 00, 02 61 91, 03 53 01, 04 65 71, 05 36 49, 06 39 98). Those are assigned to nobody and cannot ring anyone. The publication check now REFUSES any other number in clear, including the plausible-looking 06 00 00 00 XX placeholders it used to accept — so the mistake cannot come back silently. Your report also led us to two numbers we had missed, one of them a landline one transposed digit away from the reserved root; both are fixed. 

On rewriting the PR history: every name and every phone number on this branch is fictional — in the current tree and in each earlier commit. We checked every match by hand. The only numbers left in clear outside the reserved range are an obvious sequence (06 12 34 56 78), used once in a comment as the counter-example of what must NOT be used, and once inside a fake transcript to prove the masking works on provider content.

We would rather not force-push a rewritten history right now: the links shared for the hackathon point at these commits, and rewriting them would break the submission while judging is under way. As soon as the contest closes we will publish the clean version 0, with the history rewritten as you asked.

2) English-only repository content

Addressed below.

3) The CALL-E bearer credential

The API base URL is now validated before any request is built: https on heycall-e.com or one of its subdomains, or loopback for a local test double. Anything else raises and the client falls back to simulation — it fails closed, it never sends the Authorization header to an arbitrary origin. The CALLE_API_URL environment variable goes through the same check as the constructor argument, since that was the path you named. 

4) The false safety/validation claims

- The committed bench report said "all checks passed" while a real-client guard had failed. Two causes, both fixed: Journal.echecs only returned failed CASES, so a broken LOCK was never counted at all; and the guard cleared the CALLE_API_KEY variable but not donnees/cle_calle.txt, so a key on disk defeated it. The verdict now counts broken locks, and the guard isolates the key file in a temporary directory.

- The masked-only audit could persist raw provider content. Every current caller did mask upstream, but nothing enforced it — the promise rested on the discipline of whoever calls. Masking now happens at the write point inside _auditer, so a future caller cannot bypass it. It is idempotent, so quotations already masked upstream are unharmed.

Each fix carries a counter-test: neutralise the fix, and the guarding test turns red. Three nets pass on the result — 1,221 unit tests, an end-to-end bench of 661 checks over 121 combinations, and the publication check.

On point 2 — English-only content

All comments and docstrings across the product package and the test suite (2,776 pieces in tests/test_ringback.py alone) are now in English. The change is comments-only: I compared the AST of every file before and after, with docstring values blanked, and the skeletons are identical. The three nets pass unchanged.

Two things stay in French, deliberately:

- The user interface. RingBack is built for one French physiotherapy practice, and what the voice agent reads aloud is what the patient hears. The app does ship a full FR/EN toggle (ringback/langue.py) with the button on every page, so an English-speaking reviewer can run the whole product in English, including the brief sent to CALL-E.

- French text quoted inside English comments: the owner's requests word for word (they are dated evidence for why a rule exists), the status values stored in the database, and on-screen labels the tests search for. Translating those would either falsify a quotation or break the assertion that reads them.

Happy to go further if you think the interface source itself is in scope.

Ray-56:
Must Fix for head 644db896:
- The frozen privacy blocker remains: remove plausible full contact identifiers from the current tree and every content-bearing PR commit, then rewrite the PR history.
- The frozen English-only blocker remains: translate repository-facing French UI, source prose, fixtures, and reports to English.

Ray-56:

Remaining Must Fix only:

- apps/python/ringback: translate the remaining extensive French repository-facing source/UI/report prose or obtain an explicit language exception under AGENTS.md.

Ludo321:
> Superseding review under the community policy adopted on 2026-09-11. This bounded list replaces the previous blockers; optional production hardening and unsupported blanket synthetic-data/cache-purge requirements do not block this contribution.
> 
> Remaining Must Fix only:
> 
>     * apps/python/ringback: translate the remaining extensive French repository-facing source/UI/report prose or obtain an explicit language exception under AGENTS.md.

@Ray-56 Thank you for the superseding review, and for the new policy — it settles the history question.

On the one remaining point, I would like to request an explicit language exception under AGENTS.md rather than translate the interface source.

RingBack is a product for a French physiotherapy practice. Its user interface is French at the source because that is what the practitioner and the patients read and hear. English is produced at output by a translation layer, with a toggle on every page, so a reviewer can use the whole product in English today.
Everything that faces the repository — code comments, docstrings, commits, this PR — is in English. What remains in French is the product's own user-facing text and its synthetic fixtures.

Rewriting the source strings in English would invert that architecture across 2,773 interface strings and the 1,221 tests that assert on them. I cannot make that change safely, and it would leave the practice with a worse product.

If the maintainers can grant the exception, I will state it in the app README.
If not, I understand.

PR #298, opened 2026-09-03
feat(apps): add blood-bank-dispatch, a parallel blood-stock enquiry app

Otitodev:
## What this adds

`apps/python/blood-bank-dispatch/` — a runnable Python app that files one blood-stock request and fans out one CALL-E call per blood bank **in parallel**, asking a fixed set of stock questions (units available, screening/cross-match status, release policy, cost per unit, transport time, contact person, alternatives). Results are extracted into a strict JSON schema, persisted in Postgres alongside the raw CALL-E response and transcript, and rendered as a live card UI with a ranked shortlist for a human to act on.

A serial 20-minute manual calling loop becomes a parallel 90-second one. The agent only gathers and reports: it never reserves, never dispatches, and never makes a clinical decision.

## CALL-E usage

- Python SDK (`calle-ai`), `client.calls.create` + `wait_for_result`, called at runtime via `asyncio.to_thread`
- Persisted idempotency keys (`run:{id}:bank:{key}:v1` sent as the `Idempotency-Key` header); network retries replay instead of double-dialing (verified live)
- `result_schema` / `recipient_result_schema` with `unknown` enum members, per the API's own guidance — hedged answers are kept as real information
- Transcripts (`recipients[n].attempts[m].transcript_turns`) stored as raw evidence beside every parsed column
- Unsigned 0.7.0 webhook deliveries handled per current guidance (`CALL-E-Event-Id` dedup); the polling UI does not require webhooks

## Safety

- **No calls by default**: `DRY_RUN=1` runs the entire pipeline — fan-out, extraction, persistence, UI — against four simulated personas (stock / alternatives / callback / no-answer) with no credentials and no spend. Live calls are an explicit opt-in (`DRY_RUN=0`).
- Decision-support boundary stated in the task prompt, UI, and README: never reserve, never promise, never diagnose.
- E.164 validated on write; samples use reserved fictional numbers (`+1555010xxxx`); server-side rate limiting and a server-side per-run target cap; no patient names; API keys server side only.

## Testing

```bash
cd apps/python/blood-bank-dispatch
python3 -m venv venv && venv/bin/pip install -r requirements.txt
python3 migrate.py && python3 seed.py
venv/bin/python -m uvicorn app.main:app --port 8000
```

Runs end to end in dry-run mode with only a Postgres URL configured. `scripts/run_status.py` summarizes the latest run; `scripts/recover_stuck_call.py` re-drives a stuck row via idempotent replay.

Full project repo (same code, plus Windows launcher scripts and build history): https://github.com/Otitodev/blood-bank-dispatch


Ray-56:
Must Fix for head 6aeb0faa:

- The documented deployable app exposes unauthenticated routes that can add arbitrary destinations, start live call fan-out when live mode is enabled, and read stored raw responses/transcripts. Add a basic authorization boundary for the call and transcript surfaces and enforce authorized destinations.
- Mask all user-facing/logged phone numbers; the registry and run/result views currently render full values.

Ray-56:
Review for head 7ff02f62: the Railpack autodetection change does not affect the acceptance items.

Current Must Fix:
- Add a basic authorization boundary for the deployable call/transcript surfaces and enforce authorized destinations; the current routes can still add ad-hoc targets, start live fan-out, and expose stored responses.
- Mask all user-facing/logged phone numbers; registry and run/result views still render full values.

Otitodev:
Both Must Fix items are addressed in b6bf03f.

**1. Login required before anything sensitive**
- The app never runs open: with no `APP_PASSWORD` set, every route that can start calls, change the registry, or show results returns 503.
- With `APP_PASSWORD` set, those routes need a signed session cookie (stdlib HMAC, no new dependencies).

**2. Only pre-authorized numbers in live mode**
- With `DRY_RUN=0`, ad hoc numbers are rejected unless listed in `ALLOWED_DESTINATIONS`. Registry banks are added by the operator behind the login. Dry-run has no restriction since it cannot call.

**3. Phone numbers are masked in all views**
- Every page shows `+1******2671` style. Full numbers stay in the database and the runtime call prompt only.

Verified locally: 503 fail-safe on all protected routes, wrong password → 401, correct password → gated pages load, zero full numbers in rendered HTML, and a live-mode ad hoc number outside the allowlist is rejected.

Ray-56:
Superseding the prior review: the operator gate and live-destination authorization are fixed. One Must Fix remains:

1. When an ad-hoc label is blank, the app copies the full destination into the display name. That value is rendered in cards/shortlists and written to dispatch/failure logs without masking. Use a non-phone fallback and ensure display/log values cannot contain the raw destination.

Otitodev:
Fixed in 5b411c6.

- Blank ad-hoc label now falls back to the non-phone name `Ad hoc target`.
- A label that repeats the destination gets that occurrence replaced with its masked form (`sanitize_label` at run creation), so a typed number can't leak either.
- CALL-E failure messages are scrubbed the same way before being stored, since the `error` text renders on failed cards.

Verified locally: both a blank-label and a destination-in-label run produce zero full-destination occurrences in cards/shortlist HTML; dispatch logs show `Ad hoc target +1******1234` and `Ring +1******1235 urgent`.

PR #300, opened 2026-09-03
feat(call-review-console): read how a call ended, and which field said so

cnpierrepapi:
Fills the `apps/web/call-review-console` slot from `docs/roadmap.md`.

The three CALL-E surfaces report the end of a call in three vocabularies that disagree. MCP has a status for `VOICEMAIL` and one for `BUSY`. Goal Runs has neither. The Calls API has a `failure_code` with no published enum and an errors guide that says not to branch on it.

So this keeps three answers apart instead of collapsing them into one status: how the call ended, whether the job got done, and whether usable data came back. Each one names the field it came from and says whether the source stated it, whether it was worked out, or whether the source cannot express that fact at all. That third state is the one doing the work. A mapping with no way to say "this surface does not carry that fact" will invent the fact.

## What is in the PR

`apps/web/call-review-console/` is a Next app. Three doors, and the default one needs no credentials: paste a payload, paste a key, or copy a webhook URL into whatever already sends webhooks.

`docs/what-a-surface-can-tell-you.md` is the safety pattern, listed in the root README. It describes the payload shapes worth defending against and what each surface can be trusted to say.

Rows added to `README.md` and `apps/README.md`.

## The shapes it handles

The fixtures are synthetic. Each one reproduces a payload shape and every value in them is written for the file.

A call task can come back reporting the task complete while nothing establishes that a person was ever on the line. The Calls guide is upfront that there is no `answered_by` field and recommends declaring one in a result schema, and this is the shape that recommendation is protecting you from. The rule that catches it needs no voicemail detector: a completed task, on an ending where nobody can say who picked up, goes to a human.

A structured result can arrive on a call that never reached a conversation. It matches the requested schema, so code that only checks whether a result came back is satisfied by it. Reading it as `unsourced` rather than valid is the difference.

`failure_message` does not separate one failure from another. The attempt-level `failure_code` does, and the mapping treats anything read from it as inference rather than a stated fact, so it routes to review instead of resolving an outcome. That is deliberate given what `errors.mdx` says about not branching on undocumented strings, and I would rather be told I have the balance wrong than guess quietly.

Everything arriving through the webhook door is marked unsigned, since deliveries carry no signature and anybody who learns the URL can post something indistinguishable from the real thing. That is a fact about the channel rather than the call, so it is kept separate from what the payload says.

## Against the checklist

It cannot place a call. Not a promise, a shape: the two things the app can do with a key are check that the key works and read one call by id, and the endpoint that dials a phone is not reachable from any route. The key check reads a call id that cannot exist and treats the 404 as proof the key was accepted.

The paste door is the no-credential path and it is the default. A key travels in a header, is used for one request and is written nowhere, so there is no cookie, no database row and no log line holding it.

56 tests, no network and no credentials. `python3 scripts/validate_repository.py` passes. Branch name checked with `scripts/check_branch_name.py`.

The webhook door wants a Redis. Without it the endpoint answers 503 and the page says there is nowhere to put events, rather than accepting them and dropping them quietly. An inbox is one list, fifty events, expiring after a day, so there is no cleanup job to write and then forget to run.

## What it does not do

The roadmap line for this slot also mentions recordings and follow-up status. Neither is here. This is the part before both: working out what actually happened, and being straight about which parts of that are guesses.

`python/webhook-result-receiver` already covers durable webhook ingestion and reconciliation, and this does not try to repeat it.

Only two attempt codes are read, because those are the only two whose meaning I am confident about. The rest of the SIP numbering is not decoded from what the numbers look like they ought to mean. A code the mapping does not know returns unknown and says so.

The reading engine is [`asheard`](https://www.npmjs.com/package/asheard) on npm, MIT and dependency-free, published from CI with provenance. Nothing here depends on an unpublished package, and the tests run against that package rather than a local copy, so they check what the app actually imports.


Ray-56:
Must Fix for head 484902f4:

- The PR and linked public project contain payloads and transcript excerpts from real test calls. Replace them with fully synthetic fixtures, remove every real-call artifact from the current trees, and rewrite both affected public histories without repeating any personal or call value.

cnpierrepapi:
Fixed, and thanks for catching it. Force-pushed as `c79ca73`.

**Fixtures are synthetic now.** Every one of them is written for the file: call ids, greetings, transcript lines, summaries, numbers. They reproduce the payload shapes that testing turned up, which is the part that matters here, and none of them carries a value from a call that happened.

**Both trees are clean.** The one transcript in `docs/what-a-surface-can-tell-you.md` is gone and the finding is described instead. Same for the greeting quoted in the README and in the fixture evidence.

**Both histories are rewritten.** This branch is one commit on top of current `main`, so nothing earlier survives it. The linked project is now a single commit as well, force-pushed, with the tag that pointed at the old history deleted. Neither history has a payload, a transcript, a number or an account in it.

One more thing you would not be able to see, so I would rather say it. The library that PR shipped from was published to npm as `asheard@0.1.0` shortly before your comment, and that tarball included the old fixtures. `0.1.1` replaces it, published from CI with provenance, and its tarball is clean. `0.1.0` is being taken off the registry rather than left there, since a published tarball cannot be edited.

The app depends on `asheard@^0.1.1` now instead of carrying a vendored copy, which also settles the unpublished-package rule in CONTRIBUTING. Tests run against the published package, so they check what the app actually imports. 49 tests, no network, no credentials. `scripts/validate_repository.py` passes.

Ray-56:
Must Fix for head b2fc06d2:
- The frozen privacy blocker remains. Remove real-call identifiers, outcomes, and transcript excerpts from the PR narrative and linked project's current/public history, and complete GitHub-side dereferencing so the earlier content-bearing PR commit and old file URLs no longer resolve. Do not repeat any exposed value.

cnpierrepapi:
Done as far as I can take it, and I owe you a correction first.

Last time I said "nothing earlier survives it" and that neither history had a payload or a transcript in it. That was wrong. Force-pushing moved the branches; it did not remove the objects, and I should have checked rather than assumed. You were right to come back.

## What is in the current head, d6f3716

The narrative no longer carries call outcomes. The PR description, the app README and `docs/what-a-surface-can-tell-you.md` describe payload shapes and what to do about them, without reciting measurements, transcript lines or identifiers from any call.

The fixtures were already synthetic files, but several carried confidence scores lifted from real calls, which is the same value wearing a different label. Those are invented now. They keep only the shape that matters, a success and a false success sitting close together near the top, and nothing else. Two test comments that recited call durations say what the payloads show instead.

56 tests, `scripts/validate_repository.py` passes.

## What still resolves, and what I cannot reach

The earlier commits are still fetchable by SHA, in three places:

- `484902f4` in this repository, because a pull request pointed at it
- the same commit in my fork
- pre-rewrite commits in the linked project

The linked project is mine and I can purge it by deleting and recreating the repository, which drops the dangling objects. I am doing that.

The two here I cannot fix from my side. Deleting my fork would take its copy with it, but it would also close this PR, and the commit would remain reachable in this repository regardless because a PR referenced it. Only GitHub can dereference that, through a support request from someone with access to this repository.

So I would rather hand you the exact thing to ask for than leave it vague. The commit is `484902f4ef15`, on `cnpierrepapi/awesome-phone-call-agents`, referenced by this PR. If you would prefer I close this PR and open a fresh one from a clean fork so nothing in your repository ever pointed at that commit, say so and I will. That is the one route that does not need GitHub Support, and I am happy to take it.

## One thing I would rather flag than quietly drop

The safety document lost most of its evidence. It used to name what came back from specific calls, and now it describes shapes. I think it is still worth having, because the reasoning and the patterns survive. But if the point of the document was the evidence, it is weaker than it was, and you should judge it as it stands rather than as it was.

Ray-56:
Superseding review for head d6f3716: the current tree is synthetic. One frozen Must Fix remains:

- Rewrite the PR history so the earlier public head no longer exposes real call identifiers, results, transcripts, or contact artifacts.

cnpierrepapi:
This PR just closed. I closed it, on purpose, and the reason is the Must Fix.

I deleted my fork. That was the step blocking you, and I did not understand that until I went and read GitHub's actual removal process rather than assuming I knew it.

## What I had wrong

I told you the earlier commit was reachable in your repository and only GitHub could clear it, and that closing this PR and reopening from a clean fork was the one route that did not need Support. Half of that was right and the useful half was wrong.

Support will not run the cleanup while a fork still serves the data. Their page is explicit about it: they act once all references other than PRs are cleaned up and no forks have references to the sensitive data, and it tells you to ask fork owners to delete the fork entirely.

My fork was still serving `484902f4ef15` by SHA. No branch pointed at it. It sat there dangling and fetchable anyway, which is how GitHub works and is not something a force push fixes.

So I was not waiting on you. You were waiting on me, and neither of us knew it.

## Where it stands now

The fork is gone. `484902f4ef15` returns 404 in my namespace. It still resolves in this repository, because opening the PR copied it here, and that copy is the one I cannot reach.

That leaves one step and it has to come from someone with admin here. GitHub's page asks for four things, so here they are filled in:

- Repository: `CALLE-AI/awesome-phone-call-agents`
- Affected pull requests: 1, this one
- First changed commit: `484902f4ef15095b34f866cc6bc1b13ae91fab42`
- LFS objects: none

Support dereferences the PR, runs garbage collection, and clears cached views. The fork condition is satisfied now, so the ticket should not bounce.

## One thing you can do today without Support

Repository admins can archive a pull request. It shipped in July, after this PR was opened, so it may not be on your radar. An archived PR is closed, locked, and returns 404 to anyone without admin here. It is reversible and it works in bulk.

That handles the old URLs part on its own, immediately, without waiting on anybody.

What I do not know is whether archiving also stops the commit being fetched by SHA. GitHub's docs say nothing about refs either way and I have not tested it, so treat that as an open question rather than something I am telling you.

## The rebuild

I am deliberately not forking again yet. A new fork joins the same network and I do not want to hand you back the same problem the day after you clear it. Once the garbage collection has run, I will fork clean and open a fresh PR at `d6f3716`, which is the head you already reviewed and called synthetic.

The work is kept locally as one commit on top of current main, 65 files, nothing from the old history in it. Reopening is a push.

If you would rather I stayed out of it entirely, say so and I will leave it closed. No hard feelings either way. But you should not have to chase a contributor twice for the same thing, and I would rather hand you the ticket than another explanation.

PR #302, opened 2026-09-03
feat(supplyline): add SupplyLine freight sourcing negotiation app

jlorow:
## Summary

Adds [SupplyLine](https://github.com/jlorow/supplyline) as a new runnable app under `apps/typescript/supplyline/`, documenting a CALL-E-powered freight sourcing agent that demonstrates a full multi-round, stateful phone-call negotiation loop:

1. Round 1 — Sourcing:** live outbound calls to multiple carriers sequentially (CALL-E's default outbound concurrency limit is 1 call at a time), asking each for availability and an all-in rate quote
2. Comparison:** checks for a meaningful gap between quotes
3. Round 2 — Negotiation:** the broker clicks "Negotiate Best Rate," and SupplyLine calls the higher-priced carrier back, presenting the lower competing quote and asking them to match or beat it
4. Recommendation + booking:** plain-language recommendation of the winning carrier, with broker confirmation — booking is local in-memory state only, not a real confirmed booking with the carrier

## Why this is useful

Most phone-call agent examples place a single call and report the result. SupplyLine shows a call → compare → call-back → decide workflow where the outcome of one call directly determines the content and target of a second call — a pattern other agent builders can adapt (negotiation, follow-up confirmation, escalation).

## Checklist

- [x] README with setup and usage
- [x] Dry-run / preview mode documented (`MOCK_CALLS=true`, no-call path by default)
- [x] Clear credential handling (env vars only, server-side, never exposed to client)
- [x] Side-effect warning for real outbound calls
- [x] Cancellation / stopping-calls note
- [x] English-only, no secrets or real phone numbers

The full app source lives at https://github.com/jlorow/supplyline (Next.js, TypeScript).


Ray-56:
Must Fix for head 54bdccfc:

- The linked public tree/history contains full values explicitly labeled as real test phone numbers. Remove them and rewrite the affected public history without repeating any value.
- Make mock/no-call mode the actual default when configuration is absent; the linked implementation currently falls through to live calls while this contribution claims a safe default.
- Correct the material behavior claims: sourcing is sequential rather than parallel, negotiation requires another operator click rather than an autonomous callback, and booking is only local in-memory state.
- The repository lists a runnable local app but contributes only a README whose npm commands cannot run. Add the source/package or reduce this to an honestly placed external-resource entry with a working link.

jlorow:
Hi Ray, Thanks for the thorough review; all four points were addressed:

1. Real phone numbers removed from history: rewrote and force-pushed the linked source repo's entire history (git filter-repo), including an old feature branch that also contained them (deleted from remote). Verified empty across all refs with a full history grep afterward.
2. MOCK_CALLS now safely defaults to mock mode: changed the logic so real calls require explicitly setting MOCK_CALLS=false—any unset/other value now defaults to safe/no-call mode.
3. Corrected the behavioral claims: calling is now accurately described as sequential (matches CALL-E's default concurrency limit of 1), negotiation is described as user-triggered (clicking "Negotiate Best Rate"), and I added a "Known limitations" section noting booking is local in-memory state only, with no real carrier-side confirmation.
4. Added the actual runnable source: apps/typescript/supplyline/ now contains the real app (not just a README)—confirmed npm install && npm run dev starts cleanly.

Ray-56:
Must Fix for head 7962eb1b:
- The frozen privacy/history blocker remains: remove the previously identified real test-contact artifacts from the linked SupplyLine project's public history and verify the rewritten commit and old file URLs no longer resolve.
- Newly blocking on this head because executable source was added: enforce per-run exact-destination authorization and ASCII E.164 validation, use a stable intent key, and stop the carrier loop after any ambiguous create outcome instead of proceeding to another call.

jlorow:
Two updates on this PR:

1. Privacy item: History rewrite is confirmed clean on main and all current refs. Filed a GitHub Support ticket (#4730994) to purge the four remaining pre-rewrite commits that are still individually cached/reachable via direct SHA—that's the one piece outside what local git operations can fix. Will update once it's resolved on their end.
2. Calling safety hardening: Pushed commit 4f8b132 addressing all four points—ASCII E.164 validation before dialing, a per-run AUTHORIZED_CALL_NUMBERS allowlist check, a stable idempotency key passed to createAndWait, and the carrier loop now stops after any ambiguous outcome (null result or SDK error) instead of continuing to the next carrier. Added a standalone test (safety.test.ts) verifying all four behaviors, plus confirmed existing tests (round2, booking, e2e) still pass.

Let me know if anything needs adjusting.

Ray-56:
Superseding review for head 4f8b132: the current implementation closes the live-safety code blockers. One frozen Must Fix remains:

- Remove the real test-call/contact artifacts from the linked SupplyLine content-bearing public history and replace them with wholly synthetic material.

I have intentionally not repeated any contact or call values here.

jlorow:
Confirmed resolved: all four pre-rewrite commit SHAs (0930fa0, 2ffbbc3, ea931a3, 95b62f9) now return 404 on GitHub after their support team cleared the cache and ran garbage collection (ticket #4730994). No real contact/call data remains reachable anywhere in the repository's history. This should close out the last remaining item let me know if anything else needs attention.

Ray-56:
Superseding review for head 7ba881fb: the frozen privacy/history blocker is closed. The old real-data commit URLs no longer resolve, and the linked repository current tree and reachable history contain no plausible real contact values. No Must Fix items remain.

PR #303, opened 2026-09-04
feat(concord-policy-audit): add branch policy audit skill and app

himanshu748:
Concord calls the branches an operator owns, asks one fixed caller scenario, and rules each spoken answer against a written policy rubric. It returns a branch-level gap register.

A pharmacy group with forty shops has no cheap way to know whether all forty tell a caller the same thing about emergency contraception. That work is done today by paying humans to make the calls.

## What is new here

`linecanary-monitor` asserts that a line works. Concord judges what the human on it said. It has no scheduling and no recurrence, and `SKILL.md` routes line-health checks to LineCanary explicitly.

**The rubric compiles into the call.** Each criterion contributes one enum-constrained field plus a quote property to the `recipient_result_schema`, so the policy document defines what a call is allowed to return. Remove a criterion and it leaves the schema. `concord task` prints the compiled task and schema before anything is dialled.

**Rulings are made on the extracted value, never on the transcript.** An earlier version matched policy phrases against the text and reported `"No, you don't need a prescription"` as a deviation, because the forbidden phrase sits inside its own negation. Three of seven deviations were branches that had answered correctly. In an audit of your own staff a false deviation is worse than a missed one: it sends a manager to correct someone who did the right thing.

Concord has been exercised end to end against the real CALL-E API, one call placed and structured answers returned. That output is deliberately not committed and is not quoted here. Everything in this PR is synthetic, with reserved `+44 7700 900xxx` test numbers.

## Safety

The unit of analysis is the branch. `Finding` has no field for a name, role or phone number, branches are ordered by outstanding policy work rather than scored, and quotes are scrubbed of self-identification on both the fixture and the live path. A returned value with no supporting words is `UNCLEAR`, because a ruling without quoted evidence is not a ruling.

Unreached branches, hedged answers and values outside a rubric's options all rule `UNCLEAR` and go to a human, so a failed phone line can never read as a policy breach. Every branch in an audit is answered for even when the provider returns nothing for it, and results are correlated by dialled destination rather than array position.

A live run needs `--live`, an approval token bound to the full call intent, and the branches' own weekday window, checked independently. The idempotency key is derived from the canonical approved CALL-E payload, so an unchanged retry reuses the key while a material task change does not.

## Checks

- `python3 scripts/check_branch_name.py --branch feat/concord-policy-audit` passes
- 63 tests pass from the copy in this PR, including fake-transport tests asserting the exact URL, method, headers and body sent to `/v1/calls`
- `judge.py` import set is asserted by AST to be exactly `{__future__, concord}`, so the half that rules cannot dial

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,7 @@ This project is an awesome list for AI-agent phone-call workflows. Add resources
 - [`ringer-consumer-tasks`](skills/ringer-consumer-tasks/) - Compose and safely place the dreaded consumer phone calls (bill negotiation, cancellation, refund, booking, quote comparison, inquiry) as CALL-E tasks with strict result schemas, dry-run-by-default previews, and human-in-the-loop decision authority.
 - [`incident-escalation-call`](skills/incident-escalation-call/) - Walks an on-call escalation ladder one phone call at a time and records an acknowledgement only when an owner and an ETA are both quoted by words the recipient spoke, then re-reads the call over a second transport before the incident is reported as owned.
 - [`partline-part-sourcing`](skills/partline-part-sourcing/) - Preview-first industrial replacement-part sourcing that calls approved suppliers for exact part identity, quantity and shipping cutoffs, then ranks evidence-backed matches and leaves purchase and alternate approval to a human.
+- [`concord-policy-audit`](skills/concord-policy-audit/) - Calls the branches an operator owns, judges each spoken answer against a written policy rubric compiled into the CALL-E result schema, and returns a branch-level gap register that carries no field capable of identifying the person who answered.
 
 ### Apps
 
@@ -193,6 +194,7 @@ Runnable demo apps live under [`apps/`](apps/). They are not a CALL-E SDK and do
 | [`apps/python/callback-coordinator`](apps/python/callback-coordinator/) | Python | Consent-first callback triage and routing: one CALL-E call learns why a person needs a callback, classifies the outcome into a fail-closed disposition, and routes it to the right team. |
 | [`apps/python/callback-window-coordinator`](apps/python/callback-window-coordinator/) | Python | Consent-first callback-window coordinator with masked preview, stable idempotency, and structured CALL-E results. |
 | [`apps/python/partline`](apps/python/partline/) | Python | Dry-run-first industrial replacement-part sourcing that calls approved suppliers, checks exact matches against the original request and keeps purchases and alternate approval human-owned. |
+| [`apps/python/concord`](apps/python/concord/) | Python | Preview-first branch policy auditing that rules on the value a call extracted rather than the transcript, treats unreached branches and hedged answers as unresolved, and reports on locations rather than staff. |
 | [`apps/python/webhook-result-receiver`](apps/python/webhook-result-receiver/) | Python | Durable at-least-once CALL-E terminal webhook ingestion with SQLite deduplication, conflict detection, and authenticated Calls API reconciliation. |
 | [`apps/python/lead-follow-up-booking`](apps/python/lead-follow-up-booking/) | Python | Consent-first lead follow-up booking: a disclosed AI call offers only calendar-confirmed free slots and books a Google Calendar event only when the lead picks a time on the call. |
 | [`apps/python/callflow-campaign-runner`](apps/python/callflow-campaign-runner/) | Python | CSV-driven outbound campaign runner that triages structured results into auto-closed, retry, and needs-human queues. |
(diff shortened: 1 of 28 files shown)


Ray-56:
This first review found these Must Fix items:

1. Remove the real-call identifier and verbatim real-call excerpts from the tree and PR description, then rewrite the affected PR history. Public examples must be synthetic; please do not repeat the exposed values in the cleanup discussion.
2. Restrict the bearer credential to an explicit trusted HTTPS CALL-E origin; the current configurable base URL can send it to an arbitrary or insecure host.
3. Bind the approval token to the complete approved call intent. Material changes to the scenario, questions, policy/options, organization/disclosure, authorization references, or call window currently reuse the same approval.
4. Apply the documented self-identification scrub to live provider results before saving or reporting them; the live path constructs answers directly and bypasses that guard.
5. Enforce ASCII E.164 digits. The current pattern accepts non-ASCII decimal digits after the country digit.
6. Emit an UNCLEAR result for every expected branch/criterion even when a provider recipient is missing, and correlate results to the intended destination rather than letting missing recipients disappear by array length.

himanshu748:
Thanks, this was a good review. All six are addressed, and five of them were real defects rather than style. Branch history has been rewritten and force-pushed as a single clean commit.

**1. Real-call values removed, history rewritten.** The recorded run is no longer in the tree, and the call identifier and spoken excerpts are gone from the PR description and from every commit message. The branch is now one commit built fresh on `main`, and `git log -p upstream/main..HEAD` contains zero occurrences of the removed values. Everything shipped is synthetic with reserved `+44 7700 900xxx` numbers. Per your request I am not repeating the values here.

**2. Credential destination restricted.** `CALLE_BASE_URL` could previously point a live bearer token at any host, including plain http. `calle.py` now refuses anything but https on an allow-listed CALL-E origin, before any request is made.

**3. Approval token now covers the full call intent.** You were right that it only bound identifiers. The questions, policy text, allowed options, scenario, disclosed organisation, authorization references and call window were all mutable under a valid token, so an operator who approved one audit could have had a differently worded call placed to the same branches. The token now hashes everything the operator reads in the preview. Tests cover rewording a question, changing policy text, adding an option, renaming the organisation and widening the window.

**4. Scrub now applies on the live path.** This was the worst of the six. `answers_from_result` constructed `Answer` directly instead of going through `Answer.parse`, so the self-identification scrub ran on fixtures and not on real provider output, which is the only path where a real person's name can actually occur. It now parses, and a test asserts a spoken name is stripped from live results.

**5. E.164 is ASCII-only.** `\d` was matching Unicode decimal digits, so non-ASCII numerals passed validation. Now `[0-9]`, with a test.

**6. Results correlated by destination, and no branch can vanish.** Answers were built by iterating the returned recipients, so a provider returning fewer recipients silently dropped branches from the report, which directly contradicted the documented promise that a silent branch always reaches the report. Results are now matched to the dialled number, and every branch in the audit produces an answer for every criterion whether or not the provider returned one. Missing branches rule `UNCLEAR`.

One further change your item 6 surfaced: a returned value carrying no supporting words used to be ruled `COMPLIANT` or `DEVIATION`. Since the design claim is that every finding cites what was said, that is now `UNCLEAR` with a note that a ruling without quoted evidence is not a ruling.

59 tests pass from the copy in this PR, up from 50, with regressions for each item above. `scripts/validate_repository.py` passes.

Ray-56:
Superseding the prior review: the five code-path blockers are fixed. One Must Fix remains:

1. Complete GitHub-side removal or dereferencing of the superseded PR commit that contained the removed real-call artifact, then verify that the old commit and file URLs no longer resolve. The current branch rewrite removed the artifact from the active tree and ancestry, but GitHub still serves the old commit directly. GitHub Support may be needed; please do not repost any of the removed values.

himanshu748:
Confirmed. I checked resolution status only and did not open or repeat the removed contents: the superseded commit URL and the removed file URL both still return HTTP 200. The active branch and its ancestry are clean.

GitHub's documented next step for cached views and pull-request references is Support-side dereferencing and garbage collection. Because this PR is in the CALLE-AI repository, could a repository administrator please open or coordinate that Support request? I'm pausing the Devpost submission until the stale URLs no longer resolve.

himanshu748:
@Ray-56 — I have closed this superseded PR and opened privacy-clean replacement PR #306. GitHub documents that archiving makes this PR unavailable to non-admin visitors. Please use the repository-admin “Archive pull request” action on #303 and continue review on #306. After archiving, I will check only HTTP status codes for the superseded commit and removed-file URLs; if any still resolve, I will fall back to the GitHub Support purge request.

himanshu748:
@Ray-56 Thank you for merging #306. One final privacy step remains: please use the repository-admin **Archive pull request** action on this closed PR. It is still public and unlocked, and status-only checks show the superseded object still resolves. We will hold the Devpost entry until the archive is complete and the affected URLs are rechecked.

PR #309, opened 2026-09-04
feat(apps): add refcheck-ai reference-check app

HectorTa1989:
## What this adds

`apps/python/refcheck-ai` — structured employment reference checks over CALL-E.

A recruiter lists a candidate's referees; the app calls each one, works through a
role-specific question set with follow-up probes, and returns a schema-valid result:
per-question ratings, strengths, red flags, quotes, and a weighted score.

The part I think is most reusable is that **no second LLM is involved**. The question
template compiles into a CALL-E `result_schema`, so extraction and validation happen
server-side. There is no transcript-parsing pass to prompt, pay for, retry, or defend
against.

## Patterns other contributors may want

| Pattern | Where |
| --- | --- |
| Template-driven `result_schema` — a sales template produces sales fields, an engineering template produces engineering ones | `refcheck/schema.py` |
| Enum selection rules in `description`, enforcement in `enum` / `required` / `additionalProperties: false` | `refcheck/schema.py` |
| An explicit `not_answered` rating, so a skipped question is excluded from the score as missing evidence rather than counted as neutral | `refcheck/schema.py`, `refcheck/scoring.py` |
| Terminal webhook receiver for **unsigned** delivery | `refcheck/webhook.py` |
| No inference of `no_answer` / `declined` from `failure_code` | `refcheck/results.py` |
| Stable idempotency keys derived from the reference id | `refcheck/client.py` |

### On unsigned webhooks

The docs are explicit that current deliveries carry no secret, `CALL-E-Timestamp` or
`CALL-E-Signature`, and that `webhooks.verify` / `webhooks.unwrap` are legacy. So the
receiver treats every delivery as untrusted input: it serves on an unguessable path
token, requires `CALL-E-Event-Id` to match the body's `event.id`, and then re-fetches
`GET /v1/calls/{call_id}` with the API key and stores *that* snapshot. The posted body
is only a notification that something changed.

Delivery is at-least-once, so an event id is claimed in SQLite before any side effect
and duplicates are answered `200` without reprocessing. A verification failure releases
the claim and returns `5xx`, so CALL-E's retry is not swallowed.

## Checklist

- **Dry run is the default.** `python cli.py --template engineering` builds and prints
  the exact task and `result_schema`, calls nobody, and works without an API key.
- **Side effects documented.** `--execute` places one outbound call per reference and
  spends credits; the receiver writes one local SQLite file. Nothing else leaves the machine.
- **Consent gate.** `--execute` requires `--i-have-consent`, asserting the referee agreed
  to be called and the candidate authorised the check. Without it the CLI exits non-zero.
- **Cancellation documented.** The Calls API exposes no client-side cancel, so the README
  says to preview first, dispatch in waves, and notes that rotating the webhook token stops
  result consumption but not calls already in flight.
- **Credential handling.** `CALLE_API_KEY` is read from the environment only — never
  logged, printed, or persisted. Server-side use only.
- **Tests.** 63 offline tests, no credentials or network: the schema stays inside CALL-E's
  supported JSON Schema subset for every shipped template, the scoring weights behave, and
  the webhook trust boundary holds (wrong token, missing/mismatched event id, forged body,
  duplicate delivery, verification failure and retry).
- **Fictional numbers.** All samples use the reserved `+1 555 01xx` range.


Ray-56:
Must Fix for head a745ae09:
- Enforce operator-authorized ASCII E.164 destinations and mask phone numbers in preview and live output.
- Pin bearer credentials to the official HTTPS CALL-E origin.
- Replace or remove the inaccessible “full product” link; its claims and required public privacy/history review cannot currently be verified.

HectorTa1989:
Thanks — all three are fixed in 68b1036.

**Operator-authorized ASCII E.164 destinations, masked in preview and live output**

New `refcheck/phone.py`. `normalize_e164` accepts only ASCII E.164 and refuses non-ASCII confusables (Arabic-Indic, fullwidth, zero-width) rather than transliterating them — a confusable digit is a different destination, and silently correcting it would dial someone who was never authorized.

Authorization is now a separate claim from consent. `REFCHECK_ALLOWED_DESTINATIONS` is the operator asserting they own or may call each number; unset or empty authorizes *nothing*, so a misconfigured deployment places no calls rather than every call. `place_call` checks it before the client is even constructed, so an unauthorized number fails without touching the network.

Masking is applied to every path that prints a number — the recipient line, the task preview, refusal messages, structured results and transcripts. `+15555550142` renders as `+1******0142`. Refusal messages no longer echo the full number back.

```
recipient       +1******0142
authorized      NO - not in REFCHECK_ALLOWED_DESTINATIONS
```

**Bearer credentials pinned to the official HTTPS origin**

`resolve_base_url` now parses `CALLE_BASE_URL` instead of string-matching it, and requires exactly `https://api.heycall-e.com`. Plaintext HTTP, embedded userinfo, any path/query/fragment, and suffix look-alikes such as `https://api.heycall-e.com.evil.example` are all refused — that last one is why it parses rather than using `startswith`. An operator who can set one environment variable can no longer redirect the API key to a host they control.

**"Full product" link removed**

You're right that it doesn't resolve — the repository is not public, so the claims and any privacy/history review can't be verified. I removed the section and the link rather than restating them.

Tests are 63 → 109, still fully offline and credential-free: E.164 validation including the confusable cases, the allowlist failing closed, masking never returning a full number, and origin pinning against each look-alike. `python3 scripts/validate_repository.py` still passes.

Ray-56:
Superseding review for head 68b1036: the previous origin, masking, and link blockers are closed. One Must Fix remains:

- Replace the plausible non-reserved full phone examples in the current tree with standards-reserved synthetic numbers and rewrite every affected content-bearing PR commit.

This is a newly discovered privacy blocker under the repository public-history policy. I have intentionally not repeated the values here.

HectorTa1989:
Fixed and force-pushed. New head is `1bbc538`; `68b1036` no longer exists on the branch.

**What the offending values were and where they came from**

Two test fixtures I had added as "valid non-US E.164" cases. Neither was reserved — one was a live London landline range and the other a plausible Tokyo landline. My mistake was reaching for realistic-looking international numbers to prove that validation is length-flexible across country codes, without checking them against a reservation list.

**Replacements**

| Purpose | Now |
| --- | --- |
| UK landline | `+44 20 7946 0xxx` — Ofcom drama range |
| Second UK example | `+44 1632 960xxx` — Ofcom drama range; 01632 is not a real UK area code |

The existing US samples were already inside the NANP `555-0100`–`555-0199` fiction block (ATIS/NANPA), and additionally use `555` as the area code, which is unassignable. I verified each one digit-by-digit rather than assuming.

**Commit rewriting**

Only one content-bearing commit contained the values — `68b1036`, which introduced the test module. I confirmed that before rewriting:

```
a745ae0 : 0 occurrences
68b1036 : 2 occurrences
```

So `68b1036` was amended to `1bbc538` rather than replaying the whole branch, and I re-checked every commit in `origin/main..HEAD` afterwards — both are clean. The amended commit message records which reserved ranges the samples now come from, so the history explains itself.

**Preventing recurrence**

The reserved ranges are now named in the test module docstring and in the README's credential-handling section, so the constraint is visible at the point where someone would add the next sample rather than living only in review history.

109 tests pass and `python3 scripts/validate_repository.py` passes.

One thing you may want to know, offered as data rather than as any argument about my own PR: the same two values appear in `apps/typescript/calle-mcp-result-extractor`, `apps/typescript/linecanary`, and `apps/typescript/payoutproof`, which are already on `main`. If the policy scan that caught mine was run against the PR diff rather than the whole tree, those would have been missed.

Ray-56:
Superseding the earlier review: the frozen Must Fix list is closed at `1bbc5385`, and the unverifiable external-project paragraph has been removed from this PR description. Repository validation and 109 offline tests pass.

PR #313, opened 2026-09-05
feat(apps): add cod-confirm, a priced cash-on-delivery confirmation caller

saifbrand:
﻿## Summary

Cash on delivery is the default payment method across South Asia. Nothing is charged when the order is placed, so abandoning it costs the customer nothing, and a large share of those orders are refused at the door. The shop pays the courier out and back and receives nothing. Every small shop answers this the same way: somebody spends an hour a day phoning each order before dispatch.

`apps/python/cod-confirm` makes that call through CALL-E, and adds the decision the shop actually needs made first: **which orders are worth calling at all.**

Refused goods come back; the freight does not. So the loss a call can prevent is freight both ways weighted by refusal risk, not the order value. Risk is built from signals a shop already has: whether the customer has taken delivery before, whether they have refused before, how large the sum waiting at the door is, and how far the parcel travels. Orders whose call costs more than the risk it removes are left uncalled and reported, rather than quietly skipped.

In the sample order book four of five orders justify a call. The fifth is a loyal customer, a light parcel and a small sum, where calling is a small certain loss.

A confirmation counts only when the customer's own words are quoted back in `confirmation_quote`. The brief tells the agent not to fill that in from a hum, a pause, or a yes it offered them itself. An unquoted yes, an unclear answer, a wrong address with no correction, and an exhausted retry budget all route to a person. The decision table is total, so no reply can leave an order in limbo.

## Safety of the destination, and of the decision

**Destinations.** Every number is validated as ASCII E.164 and authorised before it is dialled. `CALL_ALLOWLIST` pins a run to numbers the operator names, so a stray row in an order book cannot put a call through to a stranger. Lookalike digits are rejected rather than stripped, because silently deleting a character changes the destination. No number appears in full in a log, an error or a note.

**Uncertain calls are never redialled.** A create timeout or a dropped connection leaves nobody knowing whether the phone rang. The order is escalated to a human and the sweep stops, rather than left pending for a later sweep to try again. Ringing a customer twice to ask the same question is the one failure this must not cause.

**Decisions are bound to the recipient.** A result may only confirm or cancel when it came from the order's own number. A redirected demo call, or a call the platform reports against a different destination than the one requested, is marked advisory: it is kept and readable, and it decides nothing. This also covers an idempotent replay, which can return an earlier call placed to somebody else.

**The sample order book is fictional throughout**: undialable `+999` numbers and Example addresses.

## Type

- [x] New runnable app

## Checklist

## Side effects

This app places real phone calls to real people, and does none of that without `--live`; the default run simulates the whole sweep and spends no call credit. `MAX_CALL_ATTEMPTS` bounds retries per order, nothing recurs on its own, and stopping is simply not running the command again.

The call brief forbids asking for card, bank or payment details, forbids offering discounts, refunds or delivery dates it was not given, and ends the call without arguing if the person says it is a wrong number.

## Tests

33 tests over the decision table, the pricing model and destination handling. None need an API key and none place a call.

```
python -m pytest tests -q
```


Ray-56:
Must Fix for head 64803111:
- Remove the real-call transcript/result and plausible customer contact, address, and order data from the PR body, current tree, and every content-bearing PR commit; rewrite the PR history without repeating any exposed value.
- Strictly validate and authorize the exact ASCII E.164 destination, and mask it in all output.
- Stop on create timeout/connection ambiguity instead of leaving the order pending for a later redial or continuing the sweep with a new key.
- Bind confirm/cancel decisions to the intended recipient and evidence, or keep the generated result advisory for human review.

saifbrand:
Thanks, all four were fair. Branch rewritten to a single commit on top of `main`, so none of this appears in any commit history.

**1. Real-call data removed.** The transcript is gone from the PR body and the tree, and the branch history is rewritten so no commit carries it. Worth saying for the record: the names and addresses in it were from my own fictional sample book, not a customer, but you are right that it read like a record and that is the part that matters. The sample data is now unmistakably fictional too, `+999` numbers that cannot be dialled and Example addresses, rather than plausible ones.

**2. Destination validation and masking.** New `codconfirm/phones.py` is the only way a number reaches the dialler. It requires ASCII E.164, rejects lookalike digits rather than stripping them (silently dropping a character changes the destination), strips only formatting punctuation, and checks the number against `CALL_ALLOWLIST` when one is set, so a stray row in an order book cannot ring a stranger. `mask()` is the only form allowed in a log, an error or a note: `+880*******001`. Ten tests cover it, including that a rejection message never contains the whole number.

**3. Ambiguous creates now stop the run.** `place_call` used to collapse every failure into `None`, which the decision table read as "try again later". A timeout or dropped connection is not that: nobody knows whether the phone rang, and a later sweep would redial. `CallOutcome` now separates `definitely_not_placed` from `ambiguous`. Ambiguous escalates the order to a human and sets `halt`, and the sweep breaks rather than continuing with a new key.

**4. Decisions are bound to the recipient.** This was the sharpest one and I had the bug you are describing. `DEMO_PHONE` redirects a live call to my own handset while the order still belongs to a fictional customer, so my "yes" was confirming her order. It cannot now: a redirected call, or one the platform reports against a different destination than the one requested, comes back `advisory_only`, and the decision table sends it to a human without confirming or cancelling anything. That also covers an idempotent replay returning an earlier call to a different number, which was how I found the replay issue in the first place.

33 tests, up from 15. `python scripts/validate_repository.py` passes.

Ray-56:
Superseding review for head 5a23b6d — remaining Must Fix:

- Purge the real transcript, order, address, and contact artifacts from every content-bearing PR commit; the earlier public head still resolves and remains part of the public history.
- Fix the newly introduced NEEDS_HUMAN runtime failure: that path references an undefined attempt value instead of the current result.
- Require authorization for the exact destination on every live run, deeply sanitize transcript and notes output, and stop after any ambiguous provider outcome.

The runtime failure is newly blocking because it was introduced by this head. I have intentionally not repeated any sensitive values here.

saifbrand:
Thanks, all three addressed. New head `513e07d`, and the branch is now a single commit on top of `main`.

**The runtime failure.** Real, and you found it exactly. The `needs-human` branch of the sweep read a name that was never bound, so any order routed to a human raised `NameError` mid-run. It survived because the decision table had tests and the loop that drives it did not, which is where the bug lived.

Fixed, and `tests/test_sweep.py` now exercises the loop rather than the rules: the `needs-human` path completing, the transcript being kept for whoever picks the order up, an ambiguous call stopping the sweep, and a live run refusing to start without authorisation. 39 tests now.

**Authorisation on every live run.** It was permissive when no allowlist was set. It now fails closed: a live sweep stops before the first ring unless `CALL_ALLOWLIST` names the numbers it may dial, or `CALL_ALLOW_ANY=1` says in as many words that the whole order book is intended. The per-destination E.164 validation and authorisation still run on every call, and the ambiguous-outcome halt is unchanged.

**Deep sanitisation.** Transcripts and summaries are now scrubbed at the source, not just on the way to a log. People read numbers out on calls, and those turns end up in the notes a human reads later, so anything shaped like a phone number is masked before it is written anywhere. Order numbers and amounts are left intact, since those are the point.

**On the history.** Every content-bearing commit in this PR is now clean: no transcript, no address, no contact data. The only phone numbers anywhere in the diff are the fictional `+8801700000000` style values in test fixtures and the undialable `+999` numbers in the sample order book.

You are right that the earlier head still resolves. Once a commit is unreachable from any ref, whether it stays addressable by SHA is up to GitHub's garbage collection, and a force-push cannot reach it. If you need it genuinely gone rather than unreferenced, say so and I will delete the fork and open a fresh PR from a new one, though that would close this thread and lose the review history with it. Happy either way, your call.

Ray-56:
Superseding review for head 513e07d — the NEEDS_HUMAN crash is fixed. Remaining frozen Must Fix:

- Purge the real-call, order, address, and contact artifacts from every publicly resolvable prior head and content-bearing commit.
- Remove the unrestricted CALL_ALLOW_ANY path; every live destination must be explicitly allowlisted for that run.
- Treat every non-completed or missing-result outcome that is not definitively rejected before dialing as ambiguous, halt, and require reconciliation before any later redial.
- Require a reported destination that exactly matches the intended recipient before an answer may automatically confirm or cancel an order; otherwise keep it advisory.
- Deeply sanitize transcript, summary, quote, address, and outcome text before logging or persistence.

saifbrand:
All four code items are addressed at the new head.

**Unrestricted path removed.** `CALL_ALLOW_ANY` is deleted. An empty `CALL_ALLOWLIST` now dials nothing rather than meaning "no restriction", and a test asserts that no environment variable brings the blanket permission back. A destination is dialled because it was named, never because nothing objected to it.

**Ambiguity is the default.** Creating the call and waiting for it are now separate steps, because only a failure while creating can mean no call happened. A request the platform refuses outright is the one certain non-call; a timeout, a 5xx, a 408 or 409, an accepted request with no call id, a status other than `completed`, and a completed call with no result are all ambiguous. Those escalate to a person, halt the sweep, and are excluded from `pending()`, so no later sweep redials one without somebody reconciling it first. I have accepted that this makes an ordinary no-answer require reconciliation; that is the safer side to be wrong on.

**Destination match is now positive.** The old check passed when the platform reported no destination at all, which was the gap. It now returns match / mismatch / unreported: only a reported destination equal to the order's own number may confirm or cancel, unreported is advisory, and a mismatch is ambiguous, since an idempotent replay can hand back an entirely different call.

**Sanitisation is deeper and applied at every boundary.** Transcript, summary, confirmation quote, corrected address and decline reason now pass through one reader. Terminal escapes and control characters are stripped, newlines flattened so one field cannot pose as several log lines, emails and long digit runs masked alongside phone shapes, and length bounded. The quote and the corrected address were previously stored raw; that is fixed.

71 tests, none needing an API key and none placing a call. `scripts/validate_repository.py` passes.

---

**On the history item I would rather be straight with you than mark it done.**

The branch has carried nothing sensitive since the first round, and the current head is clean. But the pre-remediation objects are still reachable, and not only through my fork: they resolve through this repository's own network as well. I checked before writing this.

Force-pushing does not remove them. Deleting my fork does not reliably remove them either, and it would close this pull request. As far as I can tell the only thing that actually purges them is GitHub Support removing the cached objects from the network, and because the network belongs to this repository rather than to my fork, that request is likely to need a maintainer here.

What I am proposing, unless you would rather do it differently:

1. I file the request with GitHub Support, naming the exact object ids and the paths, and asking for the cached views and references to be purged across the network.
2. If Support comes back asking for confirmation from the upstream owner, which I expect, you confirm it. That is the part I cannot do.
3. I do not delete the fork, so this pull request stays open, unless you tell me you would rather I did and reopened it afterwards.

Tell me where to send the object ids and I will start it today. I have deliberately not put them in this comment.

saifbrand:
Update on the history item: I did not need the object ids sent anywhere after all — I have filed it with GitHub Support myself, ticket 4733636, listing the exposed path, the commit carrying the transcript, the prior branch heads, and explicitly asking them to leave the current clean head, the fork and this pull request intact.

Nothing is needed from you now. If Support asks the upstream owner to confirm the purge, that is the one point where they may come to you.

Ray-56:
The current code-path blockers are resolved. Remaining Must Fix:\n\n- Remove plausible non-reserved full-phone fixtures from the current tree and purge the previously reviewed real-call/contact artifacts from every publicly reachable prior PR head and content-bearing commit.\n\nI have intentionally not repeated any sensitive value here.

saifbrand:
Fixtures are done. Every full number in the tree is now on country code 999, which is not allocated to any country and so cannot be routed anywhere. That was already true of the sample order book; the tests were the part still using a live national range, and you are right that they were plausible rather than reserved. Names and addresses in the fixtures are now Example ones too, so nothing in the tree resembles a real contact. 71 tests still pass and the repository validator is clean.

The branch had gone stale against main in the meantime, so it is rebased onto current main and still a single commit. The row I added to apps/README.md sits alongside the new agency-status-watch row rather than replacing it.

On the purge, that is now with GitHub Support as ticket 4733636, filed with the exposed path, the commit carrying the transcript, the prior branch heads, and an explicit request to leave the current head, the fork and this pull request intact. It is out of my hands from here: force-pushing does not clear those objects and neither does deleting the fork, so the only remaining step is Support acting on it, and possibly asking you to confirm. I will report back here as soon as they answer.

Ray-56:
The rewritten current tree and all code-path blockers are resolved at 078b7fff. Remaining frozen Must Fix:\n\n- The previously reviewed real-call/contact artifacts are still publicly reachable through prior PR heads and content-bearing commits. The pending support ticket does not close this item; the public objects must be purged before merge.\n\nI have intentionally not repeated any sensitive value here.

saifbrand:
Agreed, and I am not going to argue that an open ticket is the same as a purge. It is not, and holding the merge until the objects are actually gone is the right call.

Where it stands, plainly: I have done everything a contributor can do. The content was removed, the head is clean, the fixtures are now unroutable, and the request is filed with GitHub Support as ticket 4733636 with the exact paths and object ids. What I cannot do is make the objects disappear. Force-pushing does not, deleting my fork does not reliably, and closing this pull request would not either — the objects are reachable through this repository's network, not only through my fork, which is why I raised it here rather than quietly retrying.

So there is one thing I would ask, and it is the step most likely to actually finish this: **a request from CALLE-AI's side to GitHub Support, referencing ticket 4733636.** The network belongs to this repository, and in my experience Support treats a purge request from the network owner very differently from one filed by a fork owner. I am happy to send you the object ids and paths privately, or to add anyone you name to the existing ticket, whichever you prefer.

If you would rather not carry that, tell me and I will close this pull request instead. I would understand it, though I should be honest that closing it does not purge anything either — it only stops the contribution.

Either way I will post here the moment Support responds.

Ray-56:
Thanks — all current code blockers are closed. Remaining frozen Must Fix: prior reviewed real-call and contact artifacts are still publicly reachable through earlier PR heads/content commits; complete the public-history purge before merge.

Ray-56:
Superseding review under the community policy adopted on 2026-09-11: The former blanket requirements for synthetic-data/history-cache purging and production-level external certification are withdrawn where applicable. Previously closed basic-safety items remain closed; no actual sensitive-data disclosure was established by those residual items. Maintainer cleanup: Resolved the additive catalog conflict while preserving current main entries.

saifbrand:
Thank you for the merge, and for resolving the catalog conflict yourself.

Closing the loop on the one open thread: GitHub Support acted on ticket 4733636 this morning and cleared the unreferenced commits. I checked rather than taking it on trust — every prior head of this branch now returns "No commit found", through this repository as well as through my fork, and the file that carried the transcript returns 404 at its old ref. So the history item is resolved in fact, not only withdrawn by policy.

Thanks for holding the line on the safety review. The ambiguity rule and the positive destination match are both better for having been pushed on.

PR #320, opened 2026-09-05
feat(apps): add Call Review Console and call-review skill

rodriguescarson:
## Summary

Adds `apps/web/call-review-console`, a read-only review desk for terminal CALL-E calls, and a paired `skills/call-review` skill. Roadmap item: "Review call results, summaries, recordings, transcripts, and follow-up status".

Before a `task_completed: true` result books an appointment or closes a ticket, the console checks that the transcript actually supports it:

- every leaf of `structured_result` matched against the transcript (numbers spoken, value tokens present; enum/boolean fields marked *needs reading*; optional Gemini pass cites the supporting turn)
- `structured_result` validated against `metadata.result_schema` when present
- response latency p50/p95, silences > 4 s and overlaps derived from `transcript_turns[].offset_seconds`
- AI disclosure present (and whether it was the first turn), stop/opt-out request detected and honoured, card/ID-like numbers read aloud
- `approve` / `needs_human` / `reject` with written reasons, a recorded human disposition per call, and a benchmark across every call on file

Differs from `apps/web/callproof` (contract-gated verification of one call before acting, Rails + analyzer + PostgreSQL/Redis): this needs no contract and no stack, works on the shape the API already returns, adds transcript-derived timing and compliance metrics, and compares calls over time.

Python / FastAPI, one `pip install -e .`. Fixture-only by default (five fictional calls on reserved `+1 555 01xx` numbers), phone numbers masked everywhere, no call-creation code path (`crc/live.py` only reads `GET /v1/calls/{id}` and `/events`, opt-in by id). Webhook receiver accepts terminal events only with an optional shared-token header. No schedules or recurring jobs. 6 tests.

## Type

- [x] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

https://claude.ai/code/session_016ys3wvXtGi3J9ZrUzAPUua


Ray-56:
Must Fix for head 24ae0749:
- Require non-optional authentication for every remotely reachable transcript/result, note, fetch, benchmark, and webhook route.
- Allowlist the official HTTPS CALL-E origin before sending a bearer credential.
- Deep-sanitize phone numbers, transcripts, and result data before storage or rendering.
- Validate call IDs before using them in filesystem paths, and remove unescaped inline JavaScript interpolation; the current ID flow permits path traversal and stored XSS.

rodriguescarson:
Thanks — all four are fixed in 39beebd, and the branch is merged up to main (5f7e2ed, the only conflict was the apps README table where we both appended rows).

**Authentication.** Every route carrying call data or spending the API key now requires `CRC_CONSOLE_TOKEN` via `X-CRC-Console` or a bearer token. With no token configured they answer 503 rather than serving openly, so a deployment cannot leak by omission. `/api/ping` is the only open route and reports liveness alone. `POST /api/fetch` was the sharpest case, since it spends the CALL-E key on whatever id it is handed.

**Call ids.** You were right that the id flow permitted traversal. Ids arrive from unsigned webhook deliveries and from the URL and went straight into a filename, so `../` in an id wrote outside the data directory. They are now opaque tokens (`^[A-Za-z0-9._-]{1,128}$`, no leading dot), validated at the route and again in the store, and every write resolves and checks it stays inside the data directory.

**Origin allowlist.** `CALLE_BASE_URL` is checked against `CALLE_ALLOWED_HOSTS` (default `api.heycall-e.com`) and required to be https before any bearer credential is attached.

**Stored XSS.** Also confirmed, and worse than it looked: the escaper covered `&amp; &lt; &gt;` but not quotes, while ids were interpolated into single-quoted JS strings inside `onclick`. An id of `x',alert(1),'` executed. Ids now travel in `data-` attributes read by a single delegated listener, no handler is built by interpolation, and the escaper covers quotes and backticks too. Masking is deep now as well — phone-shaped runs are masked everywhere in the structure, including transcript turns and the structured result, not just `recipients` and `task`.

`tests/test_security.py` covers all four (40 tests pass). I checked the traversal test fails against the unpatched store rather than trusting it green.

AI-assisted (Claude Code), human-directed.

Ray-56:
Superseding review for head `5f7e2ed5`. The credential-origin and call-ID/path/XSS blockers are closed. Remaining Must Fix items:

- Make webhook authentication non-optional and fail closed; `/calle/webhook` still accepts and stores caller payloads when `CRC_WEBHOOK_TOKEN` is unset.
- Deeply sanitize call, transcript, result, and evidence data before persistence; `store.save()` still writes raw provider/webhook snapshots and review evidence is derived from raw values.
- Newly blocking on this head: making console authentication mandatory without documenting/configuring the token in the advertised fixture-default startup and request flow leaves that runnable demo returning 503. Update the default setup so the documented offline demo works end to end.

rodriguescarson:
All three closed in 26e75cb.

**Webhook fails closed.** `CRC_WEBHOOK_TOKEN` is required rather than optional now: with none configured the receiver answers 503 and stores nothing, instead of accepting and persisting caller payloads from anyone who finds the URL. Since deliveries are unsigned, that token is the only thing separating a real delivery from a stranger, so treating "unset" as "open" was the wrong default.

**Redaction moved ahead of persistence.** `store.save()` runs every snapshot through a new `crc.sanitize` before writing, so phone numbers, card-shaped runs and government-id-shaped runs never reach the disk, the structured result, or anything the evidence pass derives from them. You were right that render-time masking left all of it at rest.

The wrinkle was that redacting the digits also destroys the compliance signal that depends on them — "did the agent read a card number back" is exactly what a reviewer must be told. So findings are computed *before* redaction and recorded on `metadata.pii`, and the compliance check prefers that when present. The verdict survives; the PII does not. Redaction is idempotent, and a second pass merges the findings rather than overwriting them (that was a genuine bug I only caught by testing for it).

**The offline demo works again, and thank you for blocking on it** — I broke the documented quickstart by making console auth mandatory without touching the startup path. With `CRC_CONSOLE_TOKEN` unset the server now generates one per process and prints it at startup, the model Jupyter uses. `uvicorn crc.app:app` with no configuration gives:

```
  Call Review Console
  console token for this run: htWRLtK0…
  paste it when the page asks, or send it as X-CRC-Console.

  GET /              -> 200   page loads, prompts for the token
  GET /api/calls     -> 401   without it
  GET /api/calls     -> 200   with it, 5 fixture calls
  GET /api/benchmark -> 200   calls: 5
  POST /calle/webhook -> 503  fails closed, unconfigured
```

Never anonymous, never a fixed default, and the fixture path needs no environment at all. README and `.env.example` updated to match.

46 tests pass, including the demo flow above end to end with nothing set.

AI-assisted (Claude Code), human-directed.

Ray-56:
Must Fix before merge (rechecked at 26e75cb7):\n\n- Deeply sanitize transcript, result, and evidence content before persistence. The current sanitizer only catches a narrow set of digit patterns, so formatted/local phone forms and other contact identifiers can still be written to disk.

rodriguescarson:
Fixed in 1da7a8f. You were right that it was too narrow — it matched `\+\d{7,15}`, which is E.164 and nothing else, so a transcript saying `+1 555 010 0123`, `(555) 010-0123` or `555.010.0777` still wrote the number to disk, and email addresses were not considered at all.

Candidates are now found loosely (any run of digits and separators) and classified by counting the digits: 7–15 is a phone and gets masked, 13–19 is card-shaped, `nnn-nn-nnnn` is a government id, and addresses are redacted as email.

```
'+1 555 010 0123'      -> '+1***********23'
'(555) 010-0123'       -> '(55*********23'
'555.010.0123'         -> '55********23'
'0555 010123'          -> '05*******23'
'a@b.co'               -> '[redacted-email]'
'4111 1111 1111 1111'  -> '[redacted-card]'
'123-45-6789'          -> '[redacted-gov-id]'
```

The trap in matching that loosely is that `2026-09-04T15:03:06Z` is also eight digits with separators, and these snapshots are full of timestamps — redacting them would corrupt every `created_at` and break the timing analysis the whole review depends on. Datetimes, dates, clock times and decimals are protected before redaction and restored after:

```
'at 2026-09-04T15:03:06Z'  unchanged
'9:30 am'                  unchanged
'offset 12.5s'             unchanged
```

One subtlety that a test caught rather than my reading of the regex: a dotted phone number and a decimal look alike, and my first decimal guard shielded `555.010` so `555.010.0123` survived. A decimal may now not touch another dot or digit on either side, so `12.5` is kept and the dotted phone is not.

62 tests, covering seven written phone forms plus emails, cards and government ids, and asserting six timestamp and time forms survive.

AI-assisted (Claude Code), human-directed.

Ray-56:
Superseding review for head `21d09cbc`: the frozen deep-sanitization blocker is closed; the new coverage handles the previously identified formatted/local phone and email paths, and all 62 focused tests plus repository validation pass.

One newly blocking privacy item was introduced on this head: the sanitizer regression fixtures and unique commit text contain plausible email addresses on non-reserved domains. Because this is newly committed contact data, it is a concrete P1 under the repository's public-history policy. Replace every such address with IANA-reserved examples (for example, under `example.com`, `example.net`, or `example.org`) and rewrite the PR's unique history so the original values are gone.

rodriguescarson:
Fixed in `d5b2ed5`.

Both fixture addresses now sit under IANA-reserved domains (RFC 2606): the plus-tagged one under `example.org` via a subdomain, the short one under `example.net`. The tests still cover the same three shapes (plus-tagged local part, multi-label domain, minimal address), so no coverage was traded away to close this.

I rewrote the branch rather than stacking a commit on top, so the original values are gone from every commit in the PR's range and not just from the tip. `git log -S` over `upstream/main..HEAD` finds zero commits containing either value. Only the sanitizer commit and the merge above it were rebuilt; I reset the tree to the head you reviewed before applying the change, so `git diff 21d09cb HEAD` is exactly those three lines and nothing else in the diff moved.

62 tests pass and `scripts/validate_repository.py` passes on the new head.

rodriguescarson:
Ready for another look whenever you have time.

The fixture-address blocker is closed in `d5b2ed5`. Two commits landed after the head you last reviewed, and one of them touches the auth surface, so I would rather flag it than have you find it in the diff.

`123ec49` is docs only. The README still described the pre-review build: it claimed 7 tests where there are 62, listed a layout with no `security.py` or `sanitize.py`, and told the reader to put the webhook behind their own proxy check when the endpoint now refuses every delivery unless `CRC_WEBHOOK_TOKEN` is set.

`9d14068` rebuilds the console UI and adds `CRC_DEMO`. The reason: the hosted demo could not be opened by anybody. The console token is minted per process and printed to stdout, and nobody visiting a hosted URL can read stdout, so the live link was a login wall with no way past it. `CRC_DEMO` lets a deployment publish its own token through `/api/ping`, and it is narrow on purpose. The flag alone does nothing, because the token must also have been set explicitly, so the random per-process token can never be published. It refuses when `CALLE_API_KEY` is present, because a deployment that can reach real calls is not a fixtures demo. And it does not change the routes: a request without the token is still 401, with a test asserting exactly that while demo mode is on.

67 tests and `scripts/validate_repository.py` pass on the new head.

No urgency on merging from my side, the hackathon takes an open PR. But if the demo-mode trade is not one you want in this repo, I would rather cut it now than leave it sitting in the diff.

Ray-56:
Thanks — the current fixtures and fixture-only demo close the code and current-tree privacy blockers. Remaining Must Fix: the prior sensitive reviewed head is still publicly resolvable; complete the public-history purge before merge.

Ray-56:

PR #321, opened 2026-09-05
feat(runaround): add referral-chain tracing app with loop detection

kimdooo-a:
## What this adds

`apps/python/runaround/` — a Python app on the CALL-E Developer API that chases one question
across organizations which each name the other, and stops with quoted evidence when the
referrals close on themselves.

Every workflow in this repository assumes a desk that owns the request exists. Often it does
not: the retailer says it is the carrier's claim, the carrier says it is the shipper's claim,
and the question is never answered and never refused. Runaround makes the unit of work the
referral edge — one desk, the desk it named, and the sentence that named it.

## Two rules

- A referral advances the chain only when the recipient's own words come with it. A
  `referral_target_phone` without a `referral_quote` is refused as `unverified_referral`.
- A number spoken on a call is not permission to dial it. New destinations stop the chain at
  `awaiting_approval` until a person approves them.

## Contribution checklist

- English-only repository-facing content
- No secrets, tokens, or personal data; all sample numbers are NANP fiction-reserved `555-01xx`
- Fixture mode is the default; the live path needs an API key, a mode flag, and an explicit
  acknowledgement flag
- Side effects, cancellation, and state documented in the README; no recurring jobs
- 57 tests, standard library only, no sockets opened
- Also adds one app-table row and one safety-pattern row to `README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,6 +213,7 @@ Runnable demo apps live under [`apps/`](apps/). They are not a CALL-E SDK and do
 | [`apps/python/kept`](apps/python/kept/) | Python | Turns a payment promise made on a collections call into a validated financial record: eleven named rejection reasons stand between a spoken sentence and a ledger entry, vague amounts are refused, over-commitments are clamped to the invoice balance with the spoken figure kept beside them, and the promise is reconciled against the bank feed a week later so only the commitments that actually broke are called again. |
 | [`apps/typescript/callsweep`](apps/typescript/callsweep/) | TypeScript | Calls many local businesses, haggles each one down toward your budget on the call, ranks their offers by the best overall deal (price, what's included, availability), and books the one you pick. Dry-run no-call path by default with fictional sample shops. |
 | [`apps/python/casechaser`](apps/python/casechaser/) | Python | Chases an open claim, refund, repair, or delivery case to closure: every company promise becomes a dated, quoted commitment, broken ones climb a fixed escalation ladder, offers and denials stop at the customer, and a masked evidence pack is ready for the written complaint. Fixture mode by default. |
+| [`apps/python/runaround`](apps/python/runaround/) | Python | Chases one question across organizations that each name the other: every referral is recorded with the sentence that gave it, a number spoken on a call is not permission to dial it, and the chain stops with a quoted evidence pack the moment the referrals close on themselves. Fixture mode by default. |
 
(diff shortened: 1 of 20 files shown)


Ray-56:
Must Fix for head af6419a0:
- Enforce ASCII-only E.164 (+[1-9][0-9]{6,14}); the current Unicode digit match accepts invalid destinations.
- Recursively scrub phone numbers from every evidence-pack/output field, including desk labels, answer summaries, reference values, and referral quotes. The current masking guarantee does not cover those paths.

kimdooo-a:
Pushed a follow-up commit after exercising this against the live API today (2026-09-07). Two defects were only observable there, and neither could fail in fixture mode:

1. **`result_schema` may not use union types.** Every request this app built was refused with `400 result_schema_invalid` — *"unsupported JSON Schema type at `$.properties.answer_summary`: `['string', 'null']`"*. The app could not have placed a call from any region. The six optional fields are now typed `string` and left out of `required`, with descriptions telling the extraction model to omit the key rather than send a null. Local validation already read a missing key as absence, so nothing downstream changed.

2. **A request CALL-E rejects is not a call that happened.** The hop is written before the call on purpose, so an interrupted run shows a call that may have occurred — but a 400 or a 422 says the call task was never created. Leaving the hop in flight spent the hop budget and made `status` print `open after 1 call(s) -- no call has been placed yet`. Those two statuses now raise `CallNotPlaced`, and the runner withdraws the hop and restores the pending authorization. A timeout, a 5xx, or a transport error still keeps the hop, because there "did a phone ring?" is genuinely open.

On the live path: outbound calls to KR destinations are refused by the platform in both languages (`422 call_not_ready` — *"recognized as KR with English, which is not currently supported for outbound calls"*, and *"recognized as KR/Korean"*). So authentication, request construction, schema acceptance and the refusal itself are real; no outbound leg has been completed against a supported destination, and the README says exactly that rather than implying more.

Contract pinned at 0.7.0. 61 tests, four of them new, standard library only, no sockets opened. `scripts/validate_repository.py` passes.

Ray-56:
Must Fix before merge:

- Enforce ASCII-strict E.164 validation; the current digit class accepts Unicode decimal digits.
- Deeply sanitize all persisted and rendered provider-derived fields, including nested evidence, answers, labels, reasons, and referral text.

Ray-56:

Remaining Must Fix only:

- apps/python/casechaser: use ASCII digit validation for live E.164 rather than Python Unicode digit acceptance, and scrub real phone values from provider-derived answer/referral/evidence output paths.

PR #322, opened 2026-09-05
feat(phone-retrieval): add business fact-retrieval skill

Jcube101:
## Summary

An Agent Skill for calling businesses to find out things that are only knowable
by asking — whether something is in stock right now, what it actually costs, how
long a lead time really is. It returns one scored record per business, with the
transcript as evidence and each field marked answered, unanswered, refused or
not applicable.

`skills/phone-retrieval/` — `SKILL.md`, four `references/`, three `scripts/`.

---

## Scope

**Runs without an agent.** `scripts/call_agent.py` is plain Python 3.11+,
standard library only. `plan` → review → `run` → `status` → `show` works by hand
for anyone who wants to ring several shops and compare the answers.

**Runs under any Agent Skills-compatible host.** No host-specific frontmatter,
no vendor paths in `SKILL.md`; host detail lives in `references/`.

Two assumptions worth stating plainly:

- The host must be able to run shell commands and read files.
- **A human approves every call.** The skill will not chain plan → run.

## Try it without placing a call

```bash
cd skills/phone-retrieval/scripts
CALL_PROVIDER=fake python3 call_agent.py plan \
    --to +15550101234 --callee-name "Miller Hardware" \
    --purpose "Check availability before travelling" \
    --field "in_stock=Do you have it" --field "unit_price=What does it cost"
```

No account, no credentials, no network, nothing dials. Then `run --plan-id …`,
`status --run-id …`, and `show <id>`, which follows a plan to its run and
returns the stored result.

`CALL_FAKE_REWRITE=added|merged|referent|dropped|normalised` makes the fake
return a plan whose text differs from what was sent, so the goal-inspection step
below can be exercised offline.

Tests: `python3 test_call_agent.py` — 34 groups, no dependencies. They assert on
side effects: how many times the provider was submitted to, what mode files were
created with, whether a hollow read overwrote a record that had content.

## What it does

**Reads the returned plan before dialling.** `plan_call` returns its own version
of the task text as `display_goal`, and that text is what the agent works from.
The skill diffs it against what was sent and reports four kinds of change —
something added, a prohibition merged, a prohibition's referent changed, a field
dropped or re-ordered — while treating ordinary rephrasing as noise, because a
flag that fires on every call stops being read. `references/goal-inspection.md`
covers what the check catches and what it does not; it is not verification, and
the skill never describes it as such.

**Scores identity before any field.** The agent asks whether it reached the
business it was asked to call. If identity is never established the whole record
is caveated rather than a single field; if the wrong business answered, every
field is marked `unasked`. An answer from a business you cannot name is not
reported as though it came from one you can.

**Treats the callee as an untrusted channel.** Nothing said on a call is an
instruction to the agent. The skill matches on shape rather than wording — a
secret requested from your side, a callback number that did not come from the
operator, instructions claiming to override yours, authorisation you were not
given, an exfiltration channel — and reports any of them as a potential
manipulation attempt rather than filing them as an ordinary request. A call to
an alternate number needs its own plan and its own approval.

**Persists results locally.** Runs age out provider-side, so the outcome is
written to disk when a run reaches a terminal state. A hollow read never
overwrites a record that still has content; it is appended as dated evidence
instead.

**Reports honestly when extraction fails.** Fields are recovered by matching the
key names you requested against the summary. When the provider answers in prose
without using them, the result carries `extraction_status: no_keys_in_summary`
rather than looking like an unanswered call. Nothing is inferred from prose.

## Side effects

Real, charged, outbound phone calls that cannot be recalled. Planning is free
and does not dial; only `run` places a call, and only after a human approves the
plan.

Local state — one record per plan and per run, plus the result — written
owner-only to `.calle-runs/` by default. ⚠️ Owner-only is a POSIX guarantee;
`SKILL.md` says so.

No purchases, orders, holds, reservations, bookings or cancellations. No
agreeing to terms or prices. Businesses only — `references/safety.md` explains
why calling individuals needs consent established outside the call and is out of
scope here.

### DIFF
--- a/skills/phone-retrieval/SKILL.md
+++ b/skills/phone-retrieval/SKILL.md
@@ -1,0 +1,293 @@
+---
+name: phone-retrieval
+description: Call one or more businesses to find out something that is only knowable by asking — stock, price, lead time, opening hours, whether a service is offered — and return one scored answer per business with the transcript as evidence. Use when the answer is not on the web because it lives behind a phone line.
+---
+
+# Phone retrieval
+
+Some facts are only available by asking. Whether a part is on the shelf right
+now, what a repair will actually cost, how long a lead time really is, whether
+the shop is open on the public holiday next week. The web has the number; it
+does not have the answer.
+
+This skill calls the businesses, asks the same questions of each, and returns a
+structured record per business — every requested field scored, with the
+transcript as evidence.
+
+## When to use
+
+- The answer changes faster than a website does: stock, price today, lead time.
+- The answer was never published: whether they will hold an item, whether a
+  particular service is offered.
+- Several places need the same question and the answers need comparing.
+
+## When not to use
+
+- The answer is on the web. Look there first.
+- The task requires committing to something — placing an order, making a
+  booking, agreeing a price. This skill retrieves; it does not commit.
+- The number belongs to a private individual. See `references/safety.md`.
+
+## Side effects
+
+**This skill places real outbound phone calls to real businesses, and each call
+costs money.** A call cannot be recalled once placed. Planning is free and does
+not dial; only the run step places a call, and only after a human approves it.
+
+It writes local state — one record per plan and per run, plus the call result —
+to `.calle-runs/` by default (`CALL_STATE_DIR` to change it). These files are
+created owner-only, and they hold the transcript. ⚠️ **Owner-only is a POSIX
+guarantee.** On Windows the files are created with whatever the filesystem
+gives them; if that matters to you, put the state directory somewhere with
+appropriate access control.
+
+## Setup
+
+Python 3.11+. For real calls you also need a CALL-E account and their CLI
+installed; point `CALLE_BIN` at it if it is not on the default path.
+
+```
+python scripts/call_agent.py --help
+```
+
+### Try it without placing a call
+
+The adapter ships a fake provider. It needs no account, no credentials, no
+network, and never dials.
+
+```
+CALL_PROVIDER=fake python scripts/call_agent.py plan \
+    --to +15550101234 \
+    --callee-name "Miller Hardware" \
+    --purpose "Check availability before travelling" \
+    --field "unit_price=How much is it" \
+    --field "in_stock=Do you have it in stock"
+```
+
+Then `run --plan-id ...`, `status --run-id ...` and `show <id>` against the same
+fake. The full sequence writes the same local records a real call would.
+
+To see what the skill is actually for, set `CALL_FAKE_REWRITE` before planning:
+`added`, `merged`, `referent`, `dropped` or `normalised`. Each makes the fake
+return a plan whose text differs from what was sent, in one of the four ways
+worth reporting — plus one that is harmless. Compare `goal_sent` against
+`display_goal` in the stored plan record. See
+`references/goal-inspection.md`.
+
+### Tests
+
+```
+python scripts/test_call_agent.py
+```
+
+No credentials, no network. Asserts on side effects — how many times the
+provider was submitted to, what mode files were created with, what ended up on
+disk — rather than only on return values.
+
+## Workflow
+
+### 1. Establish the fields before dialling
+
+Write down what you are asking for, as discrete fields, before any call is
+planned. `in stock`, `unit price`, `lead time in days`, `open Saturday`.
+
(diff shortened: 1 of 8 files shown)


Ray-56:
Review update for head 0eb9c62 — Must Fix:

- Default to fake, dry-run, or no-network operation; the current default selects the live provider.
- Enforce strict ASCII E.164 validation before any live call.
- Mask destinations and deeply sanitize argv, state, provider results, and transcript text in all plan, show, and result output.
- Require one explicit approval for each exact recipient; a single approval token must not authorize a batch of destinations.

Jcube101:
Thanks — all four addressed in cd96d5b.

Default provider is now the fake. The tool cannot dial unless CALL_PROVIDER=calle is set on the command that means to, and SKILL.md makes setting it on the operator's behalf a hard line.

E.164 is ASCII-strict. This was a real gap: Python's \d matches every Unicode decimal digit, so a fullwidth or Arabic-Indic numeral mid-number passed validation and reached the provider. Now re.ASCII with an explicit [0-9].

One recipient per approval. A token authorises the whole plan and an in-flight call cannot be cancelled, so a second recipient is refused rather than capped. Several businesses means several plans and several approvals.

Output sanitised at a single point. Destinations masked to country code plus last four; numbers quoted inside summaries, transcript turns and goal text swept too; confirm_token dropped from all output; argv no longer reproduces the goal.

Five new tests. Each fix is mutation-verified — reverting any one of the four turns the suite red.

Ray-56:
The fake-default, E.164, and per-recipient approval blockers are resolved. Remaining Must Fix:\n\n- Deeply sanitize every nested provider/result field and formatted phone shape; the current sanitizer still allows some raw or formatted phone/provider text to persist or print.\n- Replace plausible non-reserved full-phone and transcript-like fixtures in the current tree and public PR history. This newly blocks because it is a newly discovered concrete privacy risk.\n\nI have intentionally not repeated any sensitive value here.

Jcube101:
Both remaining items are addressed in 8f679c4.

Sanitisation now covers persistence, not just output. The sanitiser runs at the storage boundary too, so a number cannot reach disk in one field because it was quoted back in another. Destination fields, goal text, transcript turns and nested activity[] entries are all swept — verified against a real result record with nine activity entries rather than a fixture. The confirmation token stays in local state because run requires it, and it never reaches any printed surface.

I read "persist" strictly and masked the local record by default. That has a cost worth naming: the provider ages runs out, so the sidecar is the only durable evidence of a call, and a masked one no longer identifies which business it belongs to. CALL_STATE_FULL=1 restores the full record for operators who need it, off by default.

Fixtures are reserved ranges only — +1555 01xx and Ofcom's +44 7700 900xxx / +44 20 7946 0xxx, which are never allocated. A test scans the suite and fails on any phone-shaped run outside them, so it can't regress. The transcript-like sample in examples.md now describes the exchange instead of reproducing dialogue.

44 test groups. Both fixes mutation-verified: bypassing the storage sanitiser fails four assertions across all four layers, and reverting the output sanitiser fails five national-format cases.

Ray-56:
Must Fix before merge:

- Purge the plausible full-phone and transcript-like fixtures from every publicly resolvable content-bearing PR commit. The current tree is fixed, but forward deletion does not remediate the retained history.

Jcube101:
Agreed — forward-fix wasn't enough, you're right that the earlier commits stayed publicly resolvable regardless of what HEAD looked like. Rebased and force-pushed: all three real numbers (and one Arabic-Indic-digit encoding of one of them that the original review passed over) are now replaced with the same reserved ranges used in the fixtures — +1555 01xx / +44 7700 900xxx / +44 20 7946 0xxx. Verified with a script that decodes Unicode digits from any script back to ASCII before checking, not just plain-text pattern matching, across all four commits and every file in the skill directory — nothing outside the reserved ranges remains anywhere in history. Let me know if anything else needs a look.

Ray-56:
Must Fix before merge:

- The frozen history blocker remains: rewrite the branch so the original content-bearing commit with full-phone and transcript-like fixtures is no longer in the PR ancestry. Current-tree sanitization alone is not sufficient.

Jcube101:
Filed a request with GitHub Support to purge the three dangling commits (cd96d5b, f4fe70a, 8f679c4) from PR references — ticket's in. Will confirm here once it's processed.

Ray-56:

PR #329, opened 2026-09-05
feat: add countercall agent skill

edycutjong:
## Contribution Area

**Agent Skills** — adds `skills/countercall/` and one entry to the README resource list.

## What it does

CounterCall calls a government service counter **before** a citizen travels there, and returns a schema-validated checklist of what to bring: required documents, total fee, payment method, originals or copies, whether an appointment is needed, how certain the clerk sounded, and the verbatim line the clerk said as the evidence for the rest.

In Indonesia — and this generalises — published requirement lists are incomplete, outdated, or quietly contradicted at the window. The one reliable source is the office's phone line, and that line is busy, IVR-gated, and answered on the fourth try. So people don't call. They travel, they queue, and they get turned away over one photocopy.

## The decision it exists to make

**When the clerk does not know, the field renders empty.** Not filled with a typical value, not quietly dropped so the gap disappears. Somebody may cross a city on the strength of this answer. An unknown fee costs one question at the window; an invented one costs the trip.

The result contract is pinned with `additionalProperties: false`, so a malformed answer becomes a *detectable* failure rather than a silently accepted wrong checklist. All eight `GoalRunError` codes route to distinct outcomes and **not one of them renders a partial checklist**.

## Side effects

Places **at most one outbound phone call** per invocation, and only when `--live` is passed explicitly.

## Safety

The callee is a public servant who did not opt in. That asymmetry drives every rule in [`references/safety.md`](skills/countercall/references/safety.md):

- states plainly at the start that it is an automated assistant, and why
- stops immediately on refusal; records `declined` and renders no checklist
- **one call per office, per procedure, per day**, enforced by a business-stable `Idempotency-Key` (`countercall:{office}:{procedure}:{date}:v1`) rather than a random UUID, so a retry after a crash cannot double-dial a queue with a person in front of it
- published general-enquiries lines only, during opening hours — never a personal mobile, never an emergency or crisis line
- no personal data sent as a call variable — the clerk is asked about a *procedure*, not about a person
- **never infers a phone number.** A number enters `data/offices.json` only when a human read it off the office's own published page and recorded the URL and the date

## Dry-run / no-call path

Dry-run **by default**, per the contributing guidelines:

```bash
# validates the number, diffs the live Goal contract against the pinned one,
# prints the idempotency key — dials nothing, needs no credentials
node skills/countercall/scripts/preflight.mjs --office <id> --procedure "<procedure>"

# prints the exact request that would be sent, then stops
node skills/countercall/scripts/call.mjs --office <id> --procedure "<procedure>"
```

Dialling requires an explicit `--live`. A skill that dials by default is a skill that dials by accident. `preflight` also refuses the dial outright when the published Goal contract has drifted from the pinned one, rather than calling a real person with a stale schema and returning a result that looks fine and is quietly wrong.

## Cancellation / rollback

There is no recurring schedule to cancel — each invocation is a single call. The idempotency key is the rollback boundary: re-running the same office/procedure on the same day is a no-op rather than a second call. `no_answer` permits exactly one retry under a `:v2` suffix, then abstains for the day.

## Checklist

- [x] English-only repository-facing content, no CJK
- [x] No secrets, tokens, private phone numbers, or personal data
- [x] Phone number in the seeded sample is **masked** (`+62XXXXXXXXXX`) — the skill cannot dial anything until a sourced number is supplied
- [x] States its host/provider: CALL-E Goals API via `@call-e/calle`
- [x] Side effects described above
- [x] Install and usage instructions in `SKILL.md`
- [x] Dry-run / no-call path, and it is the default
- [x] Cancellation behaviour documented
- [x] `python3 scripts/validate_repository.py` → **Repository validation passed.**

## Notes for maintainers

Two findings from building against the Goals API, written up in more detail at [`FEEDBACK.md`](https://github.com/edycutjong/countercall/blob/main/FEEDBACK.md):

1. A Goal Run `result` is a flat map of `GoalScalar` (`string | number | boolean`) — no arrays, no nulls — while the one-shot Calls API *does* accept `simple array.items` in its request-scoped `result_schema`. A document checklist is inherently a list, so it travels here as a newline-separated string decoded client-side, and `total_fee_idr` is *optional* rather than nullable. Array support on Goals would remove a delimiter convention that every author of a list-shaped result currently has to invent.
2. Goals are owner-scoped and authored only in CALL-E Chat, with no `POST /v1/goals`. So a published Goal cannot ship as a runnable community artifact — what this contribution can offer is the Goal **specification** plus the client. An importable Goal manifest would make skills in this repo genuinely one-click.

Submitted for the CALL-E hackathon. Full project: <https://github.com/edycutjong/countercall> · <https://countercall.edycu.dev>

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,7 @@ This project is an awesome list for AI-agent phone-call workflows. Add resources
 - [`incident-escalation-call`](skills/incident-escalation-call/) - Walks an on-call escalation ladder one phone call at a time and records an acknowledgement only when an owner and an ETA are both quoted by words the recipient spoke, then re-reads the call over a second transport before the incident is reported as owned.
 - [`partline-part-sourcing`](skills/partline-part-sourcing/) - Preview-first industrial replacement-part sourcing that calls approved suppliers for exact part identity, quantity and shipping cutoffs, then ranks evidence-backed matches and leaves purchase and alternate approval to a human.
 - [`concord-policy-audit`](skills/concord-policy-audit/) - Calls the branches an operator owns, judges each spoken answer against a written policy rubric compiled into the CALL-E result schema, and returns a branch-level gap register that carries no field capable of identifying the person who answered.
+- [`countercall`](skills/countercall/) - Calls a government service counter before a citizen travels there and returns a schema-validated checklist of what to bring — documents, fee, cash or card, originals or copies, appointment — leaving any field the clerk could not answer visibly empty instead of guessing it.
 
 ### Apps
 
--- a/skills/countercall/SKILL.md
+++ b/skills/countercall/SKILL.md
@@ -1,0 +1,164 @@
+---
+name: countercall
+description: Call a government service counter to find out exactly what a person must bring before they travel there, and return it as a validated checklist. Use when published requirements are incomplete or contradicted at the window and the only reliable source is the office's phone line. Returns required documents, total fee, payment method, whether an appointment is needed, and how certain the clerk sounded.
+license: MIT
+---
+
+# CounterCall
+
+Pak Yanto took the morning off work, rode forty minutes to the Samsat counter, and was
+turned away because his photocopy was folio instead of A4. He lost a day's income, and he
+still did not know what else was missing for tomorrow.
+
+Public service counters publish requirement lists that are incomplete, outdated, or quietly
+contradicted at the window. The one reliable source is the office's phone line, and that
+line is busy, IVR-gated, and answered on the fourth try. So people do not call. They
+travel, they queue, and they get turned away over one document.
+
+This skill makes the call instead, and returns a checklist the person can screenshot and
+take with them.
+
+It is a good fit for CALL-E's design: low-frequency, personal, high-stakes phone work
+against a number that a human would otherwise have to redial all morning.
+
+## Before you start
+
+**This skill places a real phone call to a real public office.** Confirm with the user:
+
+- which procedure they are asking about, in the office's own words
+- which office, and the phone number in E.164 format, from the office's published page
+- that they want a call placed now
+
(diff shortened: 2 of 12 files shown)


Ray-56:
Review update for head d4f6fae — Must Fix:

- Mask the destination in dry-run/live output and deeply sanitize provider and error output before displaying it.
- Make the advertised contract-drift preflight part of every documented live path, and enforce the claimed opening-hours limit before a call can be placed.
- Replace the plausible real contact fixtures in the linked CounterCall project with standards-reserved synthetic data and rewrite every affected content-bearing public commit.

I have intentionally not repeated any contact values here.

Ray-56:
Superseding review for head 635eab7 — the SDK field-access documentation is fixed. The frozen Must Fix list remains:

- Mask the destination in dry-run and live output, and deeply sanitize provider and error text before display.
- Make preflight and the claimed opening-hours restriction mandatory in every documented live path.
- Replace the plausible contact fixtures in the linked CounterCall project with standards-reserved synthetic data and rewrite every affected public content-bearing commit.

I have intentionally not repeated any contact values here.

Ray-56:
The SDK documentation issue is resolved. Remaining frozen Must Fix:\n\n- Mask the destination in dry-run/live output and deeply sanitize provider and error output.\n- Make the documented preflight and opening-hours restriction mandatory on every live path.\n- Remove the previously reviewed contact and real-call artifacts from the linked public project’s current tree and content-bearing history.\n\nI have intentionally not repeated any sensitive value here.

Ray-56:

Remaining Must Fix only:

- skills/countercall: mask destinations and provider/error output; enforce the documented preflight/opening-hours restriction on the advertised live path or correct the guarantee.

PR #334, opened 2026-09-06
docs(skills): add groundtruth-verification skill

TusharTechs:
## Summary

`groundtruth-verification` turns an operational goal into hard constraints, verifies them across several suppliers with CALL-E, and returns claims that each carry the call id, the supplier's own words, and a confidence score. A hedged answer stays UNKNOWN; an honest "no fully verified match" is a first-class result.

Three things it adds that the existing verification skills do not cover:

- **A solver across candidates, not a single call.** Per-candidate call budgets, priority ordering, and early stopping once one candidate satisfies every hard requirement — so the remaining suppliers are never dialled.
- **Follow-up calls generated from state, not from a script.** A constraint left UNKNOWN by a hedged answer earns exactly one more question ("could you check for me?"); after that the run moves on and says so.
- **A pre-dial contract check for published Goals.** A Goal can be perfectly healthy and still be structurally unable to answer a hard constraint, because its version-pinned `resultSchema` declares no field for it. Running it anyway spends a real call on a real person and returns a constraint that can only ever be UNKNOWN. `scripts/check-goal-compatibility.mjs` binds each constraint to a declared result field and exits `2` when one has nowhere to land. No other skill in this repository currently touches the Goals API.

Both scripts are dependency-free and place no call. `check-goal-compatibility.mjs` works fully offline against a saved Goal document, or read-only against a live Goal via `goals.get`.

## Type

- [x] New skill

## Checklist

## Notes on side effects and cancellation

`SKILL.md` carries a **Side effects and cancellation** section with a table of what the skill causes in the world and whether each effect is reversible. The short version: it places outbound calls, at most two per candidate (`maxCallsPerCandidate`, default 2) and eight per run (`maxCalls`, default 8); it may request a temporary hold when `request_hold` is explicitly authorised; it never purchases, pays, accepts contracts, or discloses credentials, and a prohibited-phrase gate blocks such language from reaching a call task.

Stopping between calls is safe and immediate — the loop runs on discrete ticks and all state is persisted. A call already in flight cannot be recalled, and the documented response to an aborted run is to let it finish and ignore the result rather than redial. Idempotency keys mean a retried create never produces a second call.

The skill also states the rule that sample personas must never be dialled: fictional numbers are only safe while nothing can dial them, so gate it structurally rather than by care.

## Reference implementation

The skill is extracted from a working app: <https://github.com/TusharTechs/groundtruth> (live demo: <https://groundtruth-calle.vercel.app>). One real CALL-E call has been placed through it end to end; that run returned `no_match` with three constraints left UNKNOWN and none invented, which is the behaviour this skill is trying to make portable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,7 @@ This project is an awesome list for AI-agent phone-call workflows. Add resources
 - [`incident-escalation-call`](skills/incident-escalation-call/) - Walks an on-call escalation ladder one phone call at a time and records an acknowledgement only when an owner and an ETA are both quoted by words the recipient spoke, then re-reads the call over a second transport before the incident is reported as owned.
 - [`partline-part-sourcing`](skills/partline-part-sourcing/) - Preview-first industrial replacement-part sourcing that calls approved suppliers for exact part identity, quantity and shipping cutoffs, then ranks evidence-backed matches and leaves purchase and alternate approval to a human.
 - [`concord-policy-audit`](skills/concord-policy-audit/) - Calls the branches an operator owns, judges each spoken answer against a written policy rubric compiled into the CALL-E result schema, and returns a branch-level gap register that carries no field capable of identifying the person who answered.
+- [`groundtruth-verification`](skills/groundtruth-verification/) - Verify a goal's hard requirements across several suppliers with CALL-E, keeping hedged answers UNKNOWN, and refuse to run a published Goal whose result schema cannot answer a constraint.
 
 ### Apps
 
--- a/skills/groundtruth-verification/SKILL.md
+++ b/skills/groundtruth-verification/SKILL.md
@@ -1,0 +1,274 @@
+---
+name: groundtruth-verification
+description: Run evidence-backed real-world verification by phone. Turn a fuzzy operational goal ("find a compatible part available today, held until 5 PM, under budget") into hard constraints, verify them with CALL-E phone calls that adapt to what each supplier actually says, and return claims with call-linked evidence — never promoting "probably" to verified. Use when an agent must confirm a real-world fact that only a human answering a phone knows (stock, compatibility, price, availability, holds, delivery windows), when a purchase or payment must NOT be made, and when the answer needs auditable evidence rather than a guess.
+license: MIT
+---
+
+# GroundTruth Verification
+
+An evidence-backed real-world verification protocol for phone-work agents.
+CALL-E is the phone execution layer; this skill is the orchestration,
+verification, and safety discipline around it.
+
+## When to use this skill
+
+Use it when ALL of these hold:
+
+1. **The fact lives with a human, not a database.** Stock counts, exact-part
+   compatibility, "can you actually come tomorrow", "is the quoted lead time
+   real". Web search cannot settle it; calling can.
+2. **The goal has checkable requirements.** You can name what counts as done:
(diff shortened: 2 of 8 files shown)


Ray-56:
Must Fix for head 3aa073f:

- Enforce explicit per-run intent, exact-destination authorization, strict ASCII E.164 validation, and masked output; do not silently supply default authorization.
- Pin credential-bearing requests to the approved secure CALL-E origin.
- Treat timeout or any ambiguous provider outcome as a hard stop; never advance automatically to another candidate or supplier.
- Add basic authentication to every remotely reachable call-triggering or transcript/result endpoint, and make live mode explicitly operator-gated rather than activated by ambient credentials.
- Remove plausible full phone fixtures and real-call IDs/transcript artifacts from the linked project, PR description, and every affected public content-bearing commit, then rewrite history with synthetic data.
- Make the advertised app setup runnable and truthful by linking the external app with the correct working-directory instructions or including the required local package manifest.

TusharTechs:
Update after live testing.

I placed real CALL-E calls against this protocol before considering it finished, and one of them showed that the follow-up rule I had documented never fires against a real conversation.

The protocol specified `hold_available: true, hold_confirmed: false` as the trigger for a hold follow-up. On a live call, the exact sentence that trigger exists for — *"I can hold it, but I need to check with my manager first"* — came back from CALL-E as `hold_available: null, hold_confirmed: false`. Only a deterministic mock ever produces the positive flag, so an implementation following the original text would pass every test and silently never follow up on a real call. It now keys on refusal instead: an explicit no closes the door, anything short of that is unresolved.

Two further notes added from the same runs:

- **Leave a gap before redialling.** The supplier was just asked to go and check with someone. A live redial ~90s after hang-up failed with a zero-duration provider error, and a mock has no wall clock so this cannot surface in a demo.
- **Zero-duration provider failures never reached the person.** The codes observed were opaque and inconsistent (`500`, `404`, SIP `480`) and carry no actionable reason, so implementations should not branch on them. Repeated ones to a number that previously answered look like carrier-side filtering of automated traffic rather than a caller-side fault — stop dialling rather than retrying.

`python3 scripts/validate_repository.py` still passes.

Ray-56:
Superseding review for head 1ca776f — current Must Fix:

- Enforce explicit per-run intent, exact-destination authorization, strict ASCII E.164 validation, masked output, and a fixed approved credential origin in both the skill path and linked app.
- Add basic authentication to every remotely reachable call-triggering or transcript/result endpoint; ambient credentials must not enable live mode.
- Treat timeout, zero-duration provider failure, missing or ambiguous extraction, and other uncertain outcomes as a hard stop rather than redialing or advancing to another candidate.
- Remove plausible full phone fixtures and all real-call identifiers, excerpts, screenshots, and logs from the PR, linked project current tree, and affected public content-bearing history, then rewrite with synthetic data.
- Make the advertised app setup runnable by linking the app and naming the correct working directory or by including the required local package manifest.
- Enforce the claimed global call budget across all candidates. This newly blocks because the current implementation counts only the active candidate and can place calls beyond the operator-stated run limit, creating concrete unauthorized side effects.

Ray-56:

Remaining Must Fix only:

- Groundtruth live protocol: explicitly authorize destinations, mask output, pin/approve secure credential origin and stop after unknown submissions rather than automatic next candidates.
- Documented setup: supply the app link and correct working directory/package manifest; enforce the claimed global budget across all candidates.

PR #338, opened 2026-09-06
feat(apps): add attestcall compliance attestation app

MakendranG:
## What this adds

`apps/typescript/attestcall` — **AttestCall**, an app that turns a vendor compliance question into a dated, evidence-grounded, hash-chained attestation record using one disclosed CALL-E phone call.

Vendor-risk teams need *point-in-time* attestations from third parties ("Are you still PCI DSS compliant? When does the cert expire? Who audited you?"). Today that's an email thread producing undated, unverifiable claims. AttestCall places one disclosed, consent-gated call via CALL-E, captures a strict result schema, classifies the outcome **fail-closed**, and seals it into a tamper-evident audit log.

## How CALL-E is used

`src/calle.ts` imports `@call-e/calle` and calls `client.calls.createAndWait({ task, recipient, resultSchema, metadata }, { idempotencyKey, timeoutMs })`, then uses `structuredResult`, `evidence`, `completionConfidence`, and terminal `status`/`taskCompleted`.

## Safety (per CONTRIBUTING)

- Discloses AI + recording and asks consent before capturing anything.
- Fail-closed: low confidence, missing consent, ambiguity, or unknown status → `needs_human`; only a consented, high-confidence, evidence-backed "yes" is `attested`.
- E.164 validation; phone numbers masked everywhere and never written to the audit log (enforced by a test).
- Idempotency key derived from the authorization, not the attempt (no double-dial).
- **No-call demo mode ON by default**; fixtures return the real `Call` shape so the full flow runs with zero credentials/network.

## Verification

- `npm install && npm run build` — clean.
- `npm test` — **16/16 pass**, zero live calls.
- `npm run typecheck` — no errors.
- `python3 scripts/validate_repository.py` — **passes**.

Uses only fictional/reserved sample numbers. Live verification is opt-in via `.env` (`DEMO_MODE=false` + `CALLE_API_KEY`).


Ray-56:
Must Fix:\n\n- Require authentication/authorization on the remotely reachable call and records endpoints, with explicit authorization for the exact E.164 destination on each run.\n- Pin credential-bearing traffic to an approved HTTPS origin instead of accepting an arbitrary base URL.\n- Remove the stored DOM XSS paths by rendering request/provider/audit data as text rather than trusted HTML.\n- Replace plausible non-reserved contact examples in the current tree and affected public history.\n\nI have intentionally not repeated any sensitive value here.

Ray-56:

Remaining Must Fix only:

- Remotely reachable call and record endpoints: require basic authentication and per-run authorization for valid destinations; restrict credentials to approved HTTPS origins.
- Frontend record rendering: replace stored request/provider/audit innerHTML injection with text and mask live phone output.

PR #350, opened 2026-09-07
feat(apps): add centre-remix Harbour Place board (honestly fake-only CALL-E wrap)

OCnew-ops:
## Summary

Adds **Centre Remix**, a SAMPLE shopping-centre leasing board where humans and agents share one live page. Agents can filter tenancies, propose remix overlays, draft on-page outreach, and stage a fake CALL-E-shaped tenant-call record.

- **Live:** https://ornate-pie-10561d.netlify.app
- **Source:** https://github.com/OCnew-ops/centre-remix
- **Path:** `apps/web/centre-remix` (pointer README + list/table entries)

## Why it fits

- Repository-facing reference for an AI-agent phone-call workflow (`place_tenant_call`)
- Public demo is always fake-only: no credential or live dialing path
- Human **Confirm** gates the fixture write
- Reserved SAMPLE phones only; responses mask phone numbers

## Side effects

The public Netlify function and client fallback write a fixture `last_call` record only. They never place a network call or use a CALL-E credential. Non-allowlisted numbers are rejected.

## Checklist

- [x] No secrets / private phones / personal data
- [x] Side effects and fake-only boundary documented
- [x] Setup / usage in app README
- [x] No-call path only


Ray-56:
Must Fix before merge: (1) The linked live Netlify call function is remotely reachable without authentication, accepts caller-controlled confirmation and destinations, and returns unmasked call data, so the repository claim that calls only follow human confirmation is not accurate. Either make the linked project honestly fake-only or add basic auth, an exact server-controlled authorized destination, and masked responses. (2) Replace the plausible full contact example in the linked project and rewrite the affected public history; use standards-reserved fictional data.

OCnew-ops:
Thanks @Ray-56 — addressed both must-fixes:

1) **Security / honesty:** The public Netlify `place-call` function is now **honestly fake-only**. The live `@call-e/calle` dial path was removed from the public function (never dials even if `CALLE_API_KEY` is present / `dry_run=false`). Confirm only gates writing a fixture `last_call`. Phones must be standards-reserved SAMPLE numbers (`+15550100100`, `+15550100101`, `+61491570006`, `+61491570156`); responses **mask** the phone (no full E.164). Verified live: https://ornate-pie-10561d.netlify.app

2) **Privacy:** Replaced the plausible `+61400000000` example with reserved SAMPLE `+15550100100` and **rewrote / force-pushed** reachable `main` on `OCnew-ops/centre-remix` so the old example is gone.

Pointer README + list/table wording in this PR updated to match.

Ray-56:
Must Fix before merge:

- Remove the plausible contact identifier from the linked public project’s still-resolvable dry-run refs and affected history, then correct the privacy claim if any such artifact remains. Cleaning only the external default branch does not remediate those public refs.

OCnew-ops:
Thanks @Ray-56 — follow-up on the leftover public refs:

- Deleted public branches `calle-dry-run-wrap` and `calle-dry-run-on-main` on `OCnew-ops/centre-remix` (they were still resolvable dry-run refs).
- Default `main` remains rewritten: no `+61400000000` on the current tree.
- Remaining artifact: GitHub’s immutable `refs/pull/1/head` (merged PR #1, sha `879c55b`) still resolves `BUILD.md` with that SAMPLE string. `DELETE` on `refs/pull/*` returns 422 read-only, so we cannot drop that PR ref via API.

If a support purge of that pull ref is required before merge, say so and we will file it. Public demo itself is still honestly fake-only: https://ornate-pie-10561d.netlify.app

Ray-56:
Superseding review under the community policy adopted on 2026-09-11: The linked public default branch and reachable source history were refreshed. The contribution is an honestly fake-only reference; the old cached SAMPLE-phone purge prerequisite is withdrawn. No actual credential or private third-party artifact was found in this check.

Issue #316, opened 2026-09-05
Address question was skipped during a call and returned as unclear


JJasonSun:
SaifBrand reports that an order-confirmation call skipped the requested address question, then returned `address_correct: "unclear"`. The call summary reportedly said the agent moved on after a hearing check instead of asking the address question again.

The reporter has now supplied the Call IDs, environment, task fragment, and result below. These records still need server-side comparison; `"unclear"` alone does not establish incorrect extraction or a false confidence score.

## Supplied reproduction

- Call IDs: `call_lkl9-Z0dKqD-IEmoMSsIhw` and `call_-dbg52dQ9svAMTgiemrqQg`.
- Environment: `calle-ai 0.7.0`, `httpx 0.28.1`, Python `3.14.6`, Windows 11, as reported.
- The task had a 90-second limit. The skipped step was:

```text
4. If yes, read the delivery address back and ask whether it is correct.
   If it is wrong, ask for the correct one and repeat it back to check.
```

The schema required `address_correct`, with enum values `yes`, `no`, and `unclear`. The reported result was:

```json
{
  "confirmed": "yes",
  "confirmation_quote": "yes i still want it",
  "preferred_time": "after six in the evening",
  "address_correct": "unclear",
  "corrected_address": "",
  "reached_customer": "yes"
}
```

The reporter wants the agent to attempt the requested question and wants an unasked question to be distinguishable from a question that was asked but received no clear answer.

## Reporter workaround and remaining investigation

The reporter removed numbered steps, made the three questions explicitly mandatory, required any missed question to be asked before goodbye, and defined `unclear` as asked-but-unanswered rather than not-asked. They report that the first retest worked with the same schema and model; no retest Call ID was supplied.

Their hypothesis is that the conditional wording made step 4 appear optional and the 90-second limit encouraged early completion. This has not been confirmed. They also suggest a documentation note about conditional task wording.

Next step: inspect the two supplied call records against the task and result, establish whether and why the address question was skipped, and determine how to distinguish an unasked question from an ambiguous answer. The successful prompt revision is a reported workaround, not a confirmed platform fix.

Related: #236 tracks skipped required Goal steps. #181 concerns a different result/confidence report and does not establish this case's cause. PR #353 references this report for a contributor-side evidence-checking skill; it does not establish a platform fix. The separate 503 report remains in #317.

Source: https://discord.com/channels/1493880186826133504/1545675777855463484

[...]

Issue #283, opened 2026-09-02
Post-restoration outbound API call stayed queued for 49 minutes and dialed after waitForResult timed out


JJasonSun:
## What happened

A Developer API outbound call to Kenya stayed `queued` with `attempts: []` for about 49 minutes. The SDK's 10-minute `waitForResult` timeout raised `CalleTimeoutError`, but the call later dialed and completed.

The call was created after the service restoration. The available evidence does not establish whether the delay came from residual backlog, KE routing, account configuration, or another server-side condition.

## Evidence

- Reporter-supplied call ID: `call_bxp-F39XEX7I2DJcxlC_6w`
- Created: `2026-09-01 08:32:14 UTC`, after service restoration
- Destination country: Kenya (`+254`)
- Reporter observed `status: "queued"` and `attempts: []` until dialing
- The Dashboard showed the completed call around `09:21 UTC` with a duration of `1m 41s`; the reporter used these timestamps to estimate about 49 minutes in the queue

Source: https://discord.com/channels/1493880186826133504/1544286584713781268

## Expected behavior

An accepted outbound call should dispatch or fail within a bounded period instead of staying queued with no attempts for 49 minutes and dialing after the client wait has ended.

## Next step

Inspect the queue and dispatch records for the supplied call ID and time window. Determine why this post-restoration call stayed queued with no attempts, whether residual backlog or the KE route was involved, and whether other calls from the same period could still dial late.

## Related issues

- #142 tracked API calls that timed out or stayed queued but did not dial; it was closed after the affected calls were finalized as failed.
- #150 tracked a create request that timed out without returning a `call_id`; this report has a `call_id` and timed out later in `waitForResult`.
- #213 tracks accepted calls that never dialed and remained non-terminal, plus a separate HTTP 503 failure path.

JJasonSun:
Additional queue-delay evidence from the same reporter:

- Call ID: `call_CC6Vb5EvqlpsVX4BoaNHlw`
- Reported creation time: `2026-09-01 10:56 UTC`
- When last checked, the call was still `queued` with zero dial attempts several hours after creation.
- The reporter did not confirm whether it dialed later.

Please inspect the queue and dispatch records for this call alongside the first example to determine how long it remained queued, whether it later dialed, and whether the two cases share a cause.

JJasonSun:
[...]

PR #353, opened 2026-09-07
feat(holdline): add evidence-gated batch enquiry skill

Elioz404:
## Summary

Adds `skills/holdline/`, a skill for asking one question of several places by
phone and returning only the answers the call actually established.

The problem it addresses is recorded in this repository's own tracker. Issue
#316 describes a call whose agent never raised a requested question, and whose
result came back populated anyway — schema-valid, high confidence, and not
supported by anything that was said. A schema check passes that. Nothing in the
response distinguishes *asked and not understood* from *never asked at all*.

This skill checks every returned field against the transcript turns the agent
spoke, and reports six verdicts rather than a boolean:

| Verdict | Meaning |
| --- | --- |
| `verified` | The agent raised the topic and a usable answer came back. |
| `asked_but_unclear` | The agent asked; the answer was not usable. |
| `unattributed` | Some question was asked and answered, but no probe for this field matches it. |
| `attributed` | No probe matched, but one question and one field were left over, so by elimination it can only be this. Off by default. |
| `never_asked` | Nothing was asked that this value could answer. A value present here is flagged. |
| `no_transcript` | No transcript, so nothing could be checked. |

`unattributed` exists because conflating *not confirmed* with *invented* makes
the check unusable in practice. Topic detection is lexical, and an agent
paraphrases constantly; the first version of this accused every paraphrased
question of being fabricated. The skill now counts question-and-answer
exchanges that no probe claims, and only flags a value with no exchange left
to have come from.

That change came out of measurement, not intuition. The implementation ships an
offline evaluation harness over a seeded, labelled corpus of 400 cases:

| | |
| --- | --- |
| Invented values caught | 45/45 — 100% |
| Direct asks passed | 45/45 — 100% |
| Paraphrases wrongly accused of invention | 0/44 — 0% |
| Paraphrases withheld | 44/44 — 100% |
| Prose non-answers caught | 44/44 — 100% |
| Genuine prose answers wrongly withheld | 0/44 — 0% |
| Topic mentioned but never asked, caught | 44/44 — 100% |
| Questions that arrived without a mark, eaten | 0/44 — 0% |

The last row is the honest cost and is published deliberately. A paraphrase is
still withheld, because an answer that cannot be attributed to a question
should not be stored as fact. Better probes are the only thing that moves it;
`references/probes.md` explains the trade.

**This contribution is documentation and contains no runnable code.** The three
tools it describes are served by an MCP server that lives, with the rest of the
engine — parallel dispatch, freshness ledger, route cache, webhook intake — at
https://github.com/Elioz404/HOLDLINE. Every command in `SKILL.md` and the
references is labelled as belonging to a clone of that repository, because none
of them work from this directory.

An earlier revision of this PR said the skill was self-contained and needed
nothing from that repository. That was wrong, it is the substance of the first
review comment, and it is corrected here and in all three Markdown files.

## Type

- [x] New skill
- [ ] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

- [x] No secrets, tokens, private phone numbers, call recordings, or private transcripts are included **in this contribution**. See the note below for what the linked repository holds, so nobody has to discover it.
- [x] This contribution is documentation and contains no runnable code; the commands in it are labelled as belonging to the linked repository, which defaults to a no-call simulation path.

## Side effects

`run_hold` places outbound calls, billed per call, one per dialable target, as
one call per target, dispatched together, each under its own idempotency key derived from `batchId`, `workflow`
and `intent`. Re-running the same three fetches the existing call instead of
dialing again.

`plan_hold` and `get_verdict` never dial. `run_hold` requires `confirm: true`
in the same request and there is no setting that removes the confirmation. The
MCP annotations mark `plan_hold` and `get_verdict` read-only and `run_hold`
destructive, so an agent cannot reach a telephone by accident.

Emergency and crisis prefixes (`+1911`, `+1988`, `+112`, `+999`) are refused
before dialing and no configuration permits them.

## Cancellation

Nothing in this skill schedules recurring work, so there is nothing to cancel.
A batch in flight cannot be recalled; the confirmation before dispatch is the
only stopping point, and this is stated in `SKILL.md`.

An unknown call outcome is never re-dialed. `run_hold` returns
`outcome: "unresolved"` with the call id and a classification, because a
timeout does not mean no call happened — issue #283 records a call that timed
out client-side and then dialed and completed. Reconciliation is the caller's
decision, not an automatic retry.

## Credentials

`CALLE_API_KEY` is read from the environment, used server-side only, and never
written to any output file. Every phone number in every response is masked,
including echoed metadata and error payloads.

## Verification without a call

Nothing to install: <https://tryholdline.vercel.app>

That page explains what this is and links to the console, which is at
<https://holdline-23xh.onrender.com/console>. The console runs the same engine
against a local fake transport and says `SIMULATION` on every screen. A console
reachable off its own machine is locked into simulation by `src/console/run.ts`
whatever else is set, so a visitor cannot spend credits or ring anybody — that
is enforced in code, not in the deployment config. It is on a free plan and
sleeps when idle; the page knocks on it while you read, so the wake is usually
spent before you reach the button.

From a clone, the same engine over stdio:

```bash
HOLDLINE_SIMULATE=1 npm run mcp
```

Runs the tools against a local fake transport. Every response carries
`simulated: true` and a notice that no telephone was involved, so simulated
output cannot be mistaken for a finding.

The engine repository's suite is 167 tests with no network and no credentials,
including tests that drive a real MCP client against the server over an
in-memory transport, and tests that drive the genuine `@call-e/calle`
`CalleClient` against a fake transport implementing the documented wire
contract.

## Verification with a call

No artifact of ours can prove this dials a telephone, and by design none is
kept. So the check that settles it is yours to run, on your own account and
your own number:

```bash
git clone https://github.com/Elioz404/HOLDLINE && cd HOLDLINE && npm install

# Plans a real batch. No key needed: a preview never reaches the network.
npm run call -- \
  --ask "Are you accepting new patients this week?" \
  --to '+1...' \
  --field accepting="accepting new patients, taking new patients"

export CALLE_API_KEY=...
npm run probe -- --check     # confirms the key. Dials nobody, costs nothing.

# Same command, plus --live. Prints the masked numbers and waits for you to
# type LIVE. One call billed per dialable number.
npm run call -- --ask "..." --to '+1...' --field accepting="..." --live
```

A field the conversation established comes back with the turn that established
it, quoted. A field it did not comes back `null` with the reason, so the
withholding can be judged rather than trusted. Re-running the identical command
replays instead of dialing: the idempotency key derives from the batch id,
never from the clock.

## Note on live calls

Twenty-eight live calls were placed between 2026-09-07 and 2026-09-11 to
published automated customer-service lines. **No transcript, recording, call id or other
call artifact from them is included in this contribution or kept in the linked
repository.** What follows is a summary of what they changed, in our own words.

The linked repository's public history was rewritten on September 7, 2026 to
remove these artifacts from every commit. What may still render at old direct
links is GitHub retention of unreachable objects; a purge request for them has
been submitted to GitHub Support.

They confirmed CALL-E reaches and transcribes real phone trees, and that an
agent can state a purpose and have an IVR confirm and route it. They also
disproved two things this project had written down as true: that CALL-E labels
system audio `unknown` (it does not — the automated system arrives as `user`,
which made a hold metric meaningless), and that the probe's traversal check
meant anything (it matched a menu word in a recording's own greeting). Both are
fixed.

The most useful one caught a defect in this skill's own gate. A call returned a
value that was a whole sentence reporting that nothing had been established,
and the gate marked it `verified` — because the usable-value check knew only
short sentinel tokens like `unknown` and `n/a`, and could not read prose. The
unit suite missed it, and so did 400 evaluated cases, because every value in
that corpus was one word long.

The fix went in that order: the class was added to the corpus, the damage was
measured, then the check was changed. `asked_prose_non_answer` and
`asked_prose_answered` now measure both directions — 44/44 caught, 0/44 genuine
prose answers wrongly withheld — so the trade is a number rather than a hope.

Keypad traversal is not claimed. One system stopped accepting speech and
required DTMF; the agent had only a voice, and the call ended there. That is a
limit, it was observed rather than assumed, and nothing here pretends
otherwise.

The most recent four calls found a defect in this pipeline's own seam. The
Evidence Gate credits a field only when a bot turn actually asked for it, and
the compiled task — the instruction the engine writes for the agent — carried a
goal, an optional routing hint and a disclosure, and nothing that told the
agent to ask. The probes' phrases reached the API only as result-schema
descriptions, which guide extraction and never reach the conversation. Both
calls came back with the agent having put no question of its own, every field
`never_asked`, and the gate correct each time for a reason the compiler had
made inevitable. The suite missed it because the fake transport returns
transcripts already full of questions; 400 labelled cases missed it because
every case starts from a transcript that exists. The task now carries a fourth
required segment, pinned by a test as undroppable.

The clause was then measured on the same two lines. On the recorded line the
agent still asked nothing and the gate now says so once rather than per field.
On the speech-driven system it asked: the gate found a supporting turn, quoted
it, and moved that field from `never_asked` to `asked_but_unclear` — correct,
because the question was put and the line dropped before an answer arrived.
Nothing was credited. It is the first supporting turn this gate has found on a
real telephone.

Two of those four calls returned `task_completed: true` at high confidence
while the same payloads recorded that nothing had been obtained; the other two
returned `false`. Four calls is not a measurement and is not offered as one.
Two of the four also showed the agent speaking a keypad choice aloud instead of
sending one, which CALL-E's own summary described unprompted.

A fifth call rewrote that same instruction as a concrete positive action
rather than a prohibition, and it failed the same way: offered a transfer, the
agent asked for it to be retried and then waited fifteen minutes. Partway
through that wait it did put its question, in plain interrogative order, and
the turn arrived truncated before the mark. The gate returned `never_asked` and
threw the field away, while logging three other turns from the same call as
questions because each carried a mark — all three about the call rather than
the subject.

That defect sits against the mention-only one above, and the tension is the
fix: a statement of purpose naming the topic must not count, an inverted clause
about it must, and the difference is word order at the start of a clause.
Measured in the project's usual order — `asked_inverted_no_mark` added to the
corpus first, the damage read at **44 of 44 eaten** and accuracy at **78.0%**,
then the signal added: **0 of 44 eaten**, mention-only still **44/44** caught,
nothing never-established credited, accuracy **89.0%**, and real answers
withheld down from 88 to 44. The suite pins both directions, so the second
number cannot be bought with the first.

One of the earlier calls was instructed to refuse a transfer to a person,
accepted one anyway, and waited thirteen minutes on hold until a human
answered. The call
task is the only lever over agent behaviour this API exposes — there is no
maximum-duration parameter — so a prohibition written into it is not a control.
That is recorded here as a limit of the approach, observed rather than assumed.

Everything else in this submission — the evaluation corpus, the console
scenarios, the fake transport — is synthetic and labelled as such. No figure
here is a measurement of CALL-E's live performance.


Ray-56:
Must Fix before merge: (1) The contributed skill says npm install and npm run mcp are runnable and self-contained, but it contains only Markdown and no package or scripts. Include the runnable implementation or reframe this honestly as a documentation/external-resource contribution with accurate commands. (2) Remove the linked public real-call fixtures, transcript excerpts, screenshots/video, plausible full numbers, and PR-body call artifacts, then rewrite the affected public history; later deletion alone does not remove those artifacts.

Ray-56:
Must Fix before merge:

- Remove the saved-call, transcript, identifier, and media artifacts from the linked public project’s affected history. Forward deletion is insufficient; correct the PR’s no-artifact claim once the public history is actually remediated.

Elioz404:
The rewrite went in on the 7th, tip is `fe00715`. The five call fixtures, the real-call test file, the screenshots and the video artifacts are gone from every commit, not just the tip, and the transcript phrases, the two call ids and the carrier number were scrubbed with filter-repo --replace-text/--replace-message as well.

I re-checked with a fresh clone of the public repo: a grep across every blob and every commit message in the full history comes back clean. The only call_ strings left anywhere are synthetic test ids like `call_unreachable`.

In case an old blob link still shows content, that's GitHub serving an unreachable orphaned object, not the repo's history — every reachable commit is clean. I've filed a private-information removal request with GitHub Support to purge those orphans (I'm the uploader and the account owner; the rewrite was done before filing).

The PR description now ties the no-artifact claim to the rewrite date and mentions the purge request. Happy to share the filter expressions if you want them.

Elioz404:
Closing out the last open item from the first review. Going over every number in the linked repository against the reserved fictional ranges turned one up: `+14155552671`. 555 exchange, but outside the 555-0100-0199 block, so by the standard you applied to ProofMesh it did not qualify. It was in docs, one MCP schema example and a few tests, 383 times across the history once you count the bare and masked forms.

All of it is now `+14155550199`, which sits inside the reserved block. History rewritten and force-pushed (tip `f27a591`); a fresh-clone grep over every blob and every commit message finds none of the old forms, and the suite is 157 green against the rewritten head.

The branch is also rebased on today's main, the README conflict is resolved, and the Apps entry for the repository is in. The Support request to purge the pre-rewrite orphans is still open.

Ray-56:
Remaining frozen Must Fix: the linked public project still exposes real-call screenshots, saved/live output, video, transcript, identifier, and media artifacts. Remove them from the current public tree and every affected public-history object, then provide the support/purge confirmation.

Elioz404:
All of it is removed from the repository itself: the current tree carries no real-call screenshots, live output, video or identifiers, and the history was rewritten again on September 10 for the last of them (tip `fc16f18`, fresh-clone verified as before, 157 tests green).

The purge side is filed and pending with GitHub Support, which is currently running behind: four tickets referencing this repository — 4737340 and 4737352 (September 7), 4744068 and 4744161 (September 10), consolidated today into the last one with a request to purge all unreachable objects from every rewrite. I will post their confirmation here the moment it arrives; the remaining exposure until then is SHA-only, on objects no longer reachable from any branch.

Elioz404:
The support/purge confirmation, as requested. GitHub Support completed the garbage collection on the repository (ticket 4744161, consolidating 4737340, 4737352 and 4744068): every orphaned commit left by the rewrites is gone from their servers. Verified today by fetching each former tip directly — `4ec8205`, `7119d2b`, `d68e7e6` and the rest of all three chains now return 404.

With that, nothing from the live calls exists anywhere I control: not in the tree, not in any reachable commit, and not in the unreachable objects either.

Ray-56:
Superseding review under the community policy adopted on 2026-09-11: The earlier requirements to purge synthetic/consenting-author test material or certify external production architecture are withdrawn where applicable. The accepted basic safety list remains closed; no actual secret or unauthorized private record was established by the residual item. The additive catalog conflict was resolved with both entries preserved.

PR #357, opened 2026-09-07
feat(apps): add recover payment-failure recovery app

rehna-jp:
## Summary

Adds **Recover**, a runnable TypeScript app for the Order/Exception Follow-up
use case: it calls a subscriber when their payment fails and captures a live
decision (retry now, update card, or pause) instead of relying on a passive
email that's easy to ignore. Up to 40% of subscription churn is involuntary
(a card decline, not a choice to leave), and email-only dunning has a known
backfire mode where the notification itself pushes an inattentive customer
into canceling. A live, reassuring call avoids that trap and has the highest
first-call resolution rate of any support channel.

The app follows this repo's own dry-run/preview convention throughout:
simulating a payment failure never calls CALL-E by itself -- it only shows a
preview of the exact task text and recipient that *would* be sent, and a
human must explicitly click "Confirm & place call" before CALL-E's Calls API
is ever reached. A "Cancel" path discards a pending call with no side effects
and no call ever placed.

If a call results in `no_answer`, a follow-up call is scheduled
automatically (default 24h later, capped at 3 total attempts) -- but this
does **not** bypass the preview/confirm gate. Automation only decides *when*
a preview appears; a human still explicitly confirms every call, including
follow-ups. If the outcome is `unknown` rather than `no_answer`, no follow-up
is scheduled at all -- an ambiguous result is left for a human to reconcile
rather than blindly retried, matching this repo's own "Ambiguous outcome
handling" safety doc. A "Stop follow-ups" action cancels the rest of the
chain at any point. This demonstrates the repo's own stated "provider
separation" principle directly: CALL-E only ever places a call it's
explicitly told to; this app's own scheduler decides recurrence.

This belongs in `apps/typescript/recover/` per the App directory template,
and I've added corresponding entries to both `apps/README.md` and the root
`README.md`'s Apps tables.

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist


Ray-56:
Must Fix before merge:

- Authenticate the subscriber, call, and call-log APIs, and enforce strict ASCII E.164 plus server-bound authorization for the exact destination.
- Replace the plausible personal seed identity/phone with reserved fiction and purge it from PR history.
- Authenticate and re-fetch webhook results; do not trust, persist, expose, or act on unsigned caller-supplied transcript/result data.
- Mask personal data in APIs, UI, logs, and persistence.
- Halt and reconcile an ambiguous create rather than permitting a duplicate or conflicting side effect.
- Implement the claimed bounded no-answer follow-up safely or remove the claim.
- Provide an honest fake/offline default demo path and make the documented lint/build flow work without live Stripe credentials.

rehna-jp:
sure would work on it

PR #358, opened 2026-09-07
feat: add Arc — media buying that phones for the rate

sejafahabrooashraf:
# Arc — AI media buying that makes the call

Arc plans a radio and creator campaign from a brief, then **phones the stations
and creators to get the actual rate**. Rate cards are stale and rates in Pakistan
are negotiated on the phone; the call is where the real number lives, so Arc
places it, hears the number read back, and writes down only what was confirmed.

Built with [CALL-E](https://heycall-e.com) for voice and Claude for planning.

---

## Setup

Node 20+ and a PostgreSQL database.

```bash
npm install
cp .env.example .env          # fill in DATABASE_URL at minimum
npx prisma migrate deploy     # use the DIRECT db hostname, not a pooled one
npm run dev
```

`npm run build` and `npm run dev` both run `prisma generate` first, so a stale
client cannot outlive a schema change.

**Migrations must not run through a connection pooler.** `prisma migrate deploy`
takes a session-scoped advisory lock; through pgbouncer that lock is stranded on
an idle pooled backend and every later migration fails with P1002. Point
`DATABASE_URL` at the direct hostname for migrations, the pooled one for the app.
This is spelled out in `.env.example`.

`npm test` runs 190 unit tests across 20 files. No network, no database, no calls.

---

## Dry-run and preview behaviour

**Calls are simulated by default.** With no `CALLE_API_KEY`, every call path
returns a deterministic simulated result immediately — no phone is dialled and no
provider is contacted. The whole flow still runs end to end, and the simulation is
labelled at every layer it passes through:

- the call card shows a **SIMULATED** badge
- the `Call` row stores `mock: true`, set at the moment it is known and never
  re-derived afterwards
- a confirmed rate that came from a simulated call renders with a **Simulated**
  pill beside it on campaign detail

That last one matters most: without it a generated rate would look identical to
one heard on a real call. A simulated call stays distinguishable from a real one
forever.

There is also a fixture gallery under `/dev/*` — open by default in development,
closed in production unless `ARC_ENABLE_DEV_PAGES=1` — that renders every UI state — including the live call
board, replayed from captured real transcripts — with **no network, no CALL-E and
no phone**. It is the fastest way to see what the app does without any credentials
at all.

To place a real call you must set `CALLE_API_KEY` deliberately. There is no
default that dials.

---

## Side effects

Everything here is off unless the matching credential is set.

| Trigger | Effect | Gate |
|---|---|---|
| Confirming a plan line | **Places a real phone call** to a real person | `CALLE_API_KEY` |
| Batch panel | Places up to 12 calls at once (`CALLE_MAX_BATCH`) | `CALLE_API_KEY` |
| Generating a plan | Anthropic API request, billed to your key | `ANTHROPIC_API_KEY` |
| Any call | Writes `Call` rows and updates media-plan rows | `DATABASE_URL` |
| `/api/calle/reconcile` | Reads call state from CALL-E and writes results | `CRON_SECRET` |

The recurring job is a single `GET /api/calle/reconcile` on a five-minute
schedule, run by an **external** scheduler and authenticated with `CRON_SECRET`
as a bearer token. It is not a Vercel cron — Hobby accounts are capped at daily,
and daily is slower than the 24-hour collection window, so calls would be missed.

It holds no state, retries nothing and starts no work. It only asks CALL-E about
calls that were already placed, and writes down the answer. To stop it, stop
calling the endpoint or unset `CRON_SECRET`.

**Anyone Arc calls is told, in the agent's first sentence, that it is an AI.**
The disclosure has no off switch: no context field suppresses it and no code path
reaches CALL-E without it. It is pinned by `tests/disclosure.test.ts`. Someone who
asks not to be called ends the call there, not four questions later.

---

## Cancellation

**A call in progress cannot be cancelled.** CALL-E's API is `POST /v1/calls` to
create and `GET /v1/calls/{id}` to read — there is no delete, no abort, and
nothing in the SDK for it. Once created, a call runs to completion or to failure.

So the guard sits before creation instead. `inFlightCallTo()` refuses a second
call to a number that already has one outstanding, scoped by number and capped at
30 minutes so a stuck row cannot block that number forever. The batch path
de-duplicates targets before creating anything.

Closing the tab stops the polling, not the call — the result is collected later by
the reconcile endpoint. If a connected call needs to stop, hang up at the handset.

---

## Credential handling

All credentials come from environment variables. Nothing is committed: only
`.env.example` is tracked, and it contains no values.

Secrets stay server-side. The only `NEXT_PUBLIC_` variables are a Clerk
publishable key, an app URL and a client timeout — no secret is exposed to the
browser bundle. `CALLE_API_KEY` and `ANTHROPIC_API_KEY` are read in route
handlers and server modules only.

`/api/calle/health` reports whether each credential is **present**, never its
value. `?probe=1` additionally spends one token to report whether the Anthropic
key actually *works*, because a boolean that only means "the variable is not
empty" answered `true` for a full day while every generation was silently falling
back to the offline sample.

**No real phone number ships in this repository.** Every number in the source,
fixtures, tests and archived call records is fictional. The three real calls
archived under `docs/calls/` were placed to handsets with the owner's permission;
each record carries a note saying its numbers were masked for publication and
nothing else was altered.

---

## What is real, and what is not

Real: the CALL-E integration end to end — create, poll, extract, score, persist,
link, confirm. Eighteen real calls have been placed and two returned a usable
rate. Real: the plan generation, the catalogue, the negotiation mandate, the
confirmation gate that keeps an unconfirmed number out of a field named
`confirmedRatePkr`.

Not proven: doing it *reliably*. CALL-E's queue grew from 14 seconds to over nine
minutes in six days and one request was never dialled at all. That is why calls
run simulated by default and why the reconcile endpoint exists. `README.md` has
the full account, including the call that succeeded at the provider while Arc had
already given up on it.

Arc would rather show a labelled simulation than an unlabelled fiction.

---


Ray-56:
Must Fix before merge:

- Remove raw transcripts, stable call identifiers/timestamps, and plausible full phones from the current archives, then rewrite every affected PR commit.
- Pin the CALL-E base URL to an approved HTTPS origin before sending credentials.
- Mask and deeply sanitize destinations and provider output in live responses and batch logs.
- Use a stable intent/idempotency key and halt/reconcile an ambiguous create instead of allowing a fresh-key retry to duplicate a call.
- Make the documented lint/build validation pass without requiring live Stripe credentials.

sejafahabrooashraf:
All five addressed:
- Origin pinned, checked before the API key is sent
- npm run build passes with no Stripe credentials
- Destinations masked in logs and provider payloads
- Call creation reserves its idempotency key before the create; an ambiguous outcome halts and reconciles with the stored key instead of a fresh one
- Transcripts replaced with synthetic samples and history rewritten - the branch is one commit

Ray-56:
Superseding review for head 497e2f26. Remaining Must Fix items:

- Replace the remaining real-call timing, transcript excerpts, and provider identifiers with wholly synthetic material, rewrite every affected PR commit, and correct the stale real-archive claims.
- Close the frozen live-script safety gap: every executable live path must deeply mask destinations and provider output, use a stable intent key, and stop/reconcile an ambiguous create. The standalone live scripts still print raw values, and one still derives a fresh key from the current time.

sejafahabrooashraf:
Both addressed on 6c30bbe.

Live scripts. Three paths dial a real handset; only one had been brought up to the contract, and I'd said so in a commit rather than fixing it. All three now reserve before the create, carry the persisted key, halt on an ambiguous outcome without dialling again, and route destinations through maskPhone / provider payloads through deepMaskPhones. place-one-real-call.ts had also hand-rolled its own masking - it uses the shared one now.

Real material. Five source comments quoted real conversations; rewritten as illustrations of the rule they explain. Per-call timing removed from the README and three comments. BRANDING.md §9 listed four of your recording identifiers and per-call turn counts - gone. docs/calle-support.md is excluded from the entry rather than redacted: its substance is per-call telemetry, so it stays private and goes to you on request.

One deviation, stated rather than omitted. You asked for timing to be replaced with wholly synthetic material. Fabricated figures in a section arguing for honest measurement would be worse than the problem, so the per-call ledger is removed and one aggregate range kept - 14s to 560s, one request never dialled. That range names no call, station or person. Happy to drop it entirely if you'd rather.

Verified across the whole branch: zero call ids, zero recording ids, zero transcript excerpts, zero Date.now() idempotency keys. npm run lint exits 0, npm run build passes with no Stripe credentials, 283 tests.

Ray-56:
Must Fix before merge:

- Remove the real-call narrative, measurements, provider behavior, and related artifacts from the current tree and every affected PR commit; use wholly synthetic evidence.
- Deeply sanitize every live diagnostic output and halt/reconcile ambiguous create outcomes.
- Newly discovered P1: prevent the API key from being sent through a configurable arbitrary or insecure origin.

Ray-56:
Thanks — the credential-origin blocker is fixed. Remaining Must Fix:\n- Remove real-call narratives, measurements, transcript/provider artifacts, and every affected public-history object.\n- Deeply sanitize live provider identifiers and diagnostics.\n- Persist one stable intent and stop after an ambiguous create instead of inviting a fresh resubmission.

Ray-56:
Superseding review under the community policy adopted on 2026-09-11: The former blanket requirements for synthetic-data/history-cache purging and production-level external certification are withdrawn where applicable. Previously closed basic-safety items remain closed; no actual sensitive-data disclosure was established by those residual items.

PR #361, opened 2026-09-07
feat(apps): add sundials, a warm lead agentic system

WeeeHung:
## Summary

Sundials is a CALL-E demo under `apps/typescript/sundials/`: an embeddable widget, a consented discovery call, and a ranked lead console. A merchant can observe inbound intent on a site, take explicit consent, ring while the visitor is still warm, and queue who sales should talk to next.

It belongs here because it packages a portable phone-call workflow: explicit consent, E.164, masked numbers in summaries, at most one host-side retry (no CALL-E recurrence), and cancellation if the lead later completes, is disqualified, requests a new call, or Brain is set to don’t retry.

The app ships **mock leads, calls, and Brain proposals** so reviewers can see the console as it would look after the loop has run. Turn on the mock overlay at `/app/settings`, then walk `/app/home` and `/app/leads`. Harbor `/demo` is the inbound-site walkthrough (Allow tracking → Talk to sales). A real outbound call happens only after widget consent and only when live calling is enabled; the mock queue does not place calls.

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist


Ray-56:
Must Fix before merge:

- Replace the public hardcoded SDK key with real authentication for live dispatch, status, transcript, console, and webhook surfaces.
- Require explicit per-run AI-call consent bound to the exact destination; generic contact consent is insufficient.
- Enforce strict E.164 and reject unknown/unbound country routing instead of defaulting to US.
- Stop on failed, no-speech, or ambiguous outcomes; do not automatically redial without fresh intent.
- Authenticate and re-fetch webhook results instead of persisting attacker-supplied transcript, recording, or dossier data.
- Replace plausible personal names, emails, and full phones with reserved fiction and purge every affected PR commit.

WeeeHung:
@Ray-56 
Thanks for the quick PR Review and feedbacks :D

On the hardcoded SDK key (1): as this is a hackathon demo, so there isn’t a real key issuer behind it. The hardcoded key 'hardcoded-sdk-key' is also clearly a mock key. We can still show the full flow without a proper auth service. I wanted to check whether a real SDK key issuer is needed for this submission to be complete, or whether the demo key is acceptable here. 

On consent and retry (2, 4): I have been thinking of Sundials as close to a human sales call. The lead gave a number and asked to talk, so one follow-up if nobody answers feels like what a rep would do. The scheduled recall is also intentional: for high-ticket leads we don’t want to drop them after a single miss.

I may be missing context behind those two notes — if so, please tell me. I’m happy to implement them if you see them as required for the hackathon project.
Other than that, I'll be pushing a fix for (3, 5, 6).

Ray-56:
Must Fix before merge:

- Protect the remote dispatch endpoint with server-verifiable authentication; the hard-coded public SDK key is not authentication.
- Treat failed/no-speech provider outcomes as ambiguous and stop for reconciliation instead of automatically scheduling another call.

Ray-56:
Thanks — remote authentication and the failed/no-speech stop behavior are fixed. Remaining Must Fix: the original unique commit still contains plausible contact fixtures; replace them and remove every affected object from public history before merge.

Ray-56:

PR #363, opened 2026-09-07
feat(apps): add calle-conformance, a corpus of real API responses and a checker

kasbsquall:
Walkthrough, three minutes: https://youtu.be/tAwQ147nacg

## Summary

On the free tier, **every request that reaches the planner spends one unit of the 20-per-24-hour allowance, including the ones the planner refuses.** No number is dialled, no phone rings, no call object is created, and the allowance drops by one anyway. No conflict on retry was ever captured, so nothing is claimed here about idempotency.

Nothing in the API exposes this. There is no usage endpoint: a merged app in this repository records that `/v1/account`, `/v1/balance`, `/v1/credits`, `/v1/usage` and `/v1/me` all return 404. The remaining allowance appears in exactly one place, the body of the 429 that tells you it is gone.

The consequence is ordinary. A developer in an unsupported region, or one iterating on a task the planner declines, can exhaust a day of allowance without ever reaching a person. More than twenty apps here keep a local call budget that counts one unit per created call. Under the measured rule, all of them undercount.

**How it was measured, and how to falsify it.** Two instruments, neither of which places a call. `src/watch.ts` samples the counter while at the cap: eleven consecutive readings held at 20 of 20 across two and a half hours, and capacity returned per unit at +24h, ruling out a fixed daily reset. `src/headroom.ts` then issued requests that reach the planner and are refused there; two were accepted and the third met the limiter. Both rest on payload validation running *before* the rate limiter while destination screening runs *after* it. The README states the three conditions that would falsify the reading, and `npm run headroom` re-runs the second instrument in about a minute.

This is one account, one region, one free tier, over one window. It is a measurement, not a documented guarantee.

**The corpus and the checker.** `fixtures/` holds fifteen real responses, rewritten for publication. Eight behaviours a caller would not predict from the documented shape are declared in `src/quirks.ts` as executable predicates rather than prose, so one definition labels the corpus, verifies the rewriting, and scores third-party code. `src/mask.ts` refuses to emit a fixture whose quirk set changed under rewriting, and `fixtures/README.md` is generated from the manifest so a behaviour cannot be documented without a predicate that decides it.

```
$ node src/replay.ts ../../..

corpus: 15 real responses, 8 behaviours declared as predicates.
6 projects carry call-shaped payloads to score against them.
A dot means the behaviour never appears in that project's payloads.

project                            n    1  2  3  4  5  6  7  8
---------------------------------  --  -- -- -- -- -- -- -- --
apps/python/casechaser              7   .  .  .  .  .  .  x  .
apps/python/redline                 1   .  .  .  .  .  .  x  .
apps/python/ringdown                2   .  .  .  .  x  .  x  .
apps/typescript/calle-conformance  15   x  x  x  x  x  x  x  x
plugins/zapier-calle                3   .  .  .  .  .  .  x  .
skills/verify-by-phone              1   .  .  .  .  .  .  .  .

53 projects call this API and ship no JSON payload for this checker to read.
```

No API key, no network, nothing installed: Node 22 runs the TypeScript directly. A dot is not a defect: it says a behaviour is absent from the payloads that project tests against, and nothing more.

Exit status is a verdict only when you ask for one. With no `--require`, the checker reports and exits 0 whatever it found. With `--require` it is a gate: 20 when a required behaviour is missing from a project, 45 when there was nothing to judge at all, and the README says what each means.

**Serving the corpus, and scoring a fake against it.** `src/quirks.ts` has said since it was written that a predicate scores a fake server against reality. That is now a command.

`src/fake.ts` serves the fifteen captured responses over HTTP on the routes the SDK calls, so `CALLE_BASE_URL=http://127.0.0.1:4010` is the whole integration and a project with no fixtures yet can develop against the real behaviours. The SDK renames seven fields on the way in, so payloads are mapped back to the spelling the platform sends; a fake answering in the SDK's own vocabulary leaves every mapped field `undefined`. Two things are modelled rather than served: a create the planner rejects still spends a unit, which makes the measurement above reproducible locally with no key, and the five balance endpoints return 404 as they do on the platform.

`src/score.ts` drives any fake through the SDK the way an application would and reports which of the eight behaviours it reproduces, exiting 1 when any are missing. Most mocks are written from the documentation, because the documentation is the only description of the API that ships with the API. The number it prints is the list of payloads that will reach production untested. It refuses a real CALL-E origin, because scoring means creating calls.

**What it found, and what it changed.** An absence is a shortlist. `docs/found-by-the-corpus.md` records the one that turned out to matter, with the file and line so it can be checked rather than believed: `apps/typescript/call-on-behalf` mapped a failed call to an outcome by looking for the words `voicemail`, `machine`, `answer`, `busy` or `unreachable` inside `attempt.failureCode`. The three codes in this corpus are `404`, `486` and `603`, so every one fell through to the default branch. Its own test exercised that branch with `failureCode: "busy"`, which is the documented vocabulary and is not a value an attempt carries. The test proved the branch worked. It could not show the branch was reachable.

That entry is now closed, and the trail is public rather than asserted. A maintainer opened #375 the same day, citing this report by line, and replayed the app through its own localhost fake against each code. #446 removed `failureOutcome()` and added twenty full-workflow cases covering the numeric codes, an unknown value and `null`. It merged on 11 September as `5741719`, and the comment left in its place says `failureCode` is diagnostic context and cannot distinguish no answer, a busy line or voicemail.

The same replay contradicted part of what this report had drawn from the finding, and that part is withdrawn in the README and in `docs/found-by-the-corpus.md`, each with its date. The misclassification reached the outcome; the voicemail consequence did not reproduce, because the full workflow reads the transcript before it consults the failure code. Defect 2, in `hirecall`, stands unchanged and is stated narrowly.

The README also lists six dated ceilings, including a hypothesis held for several hours and then discarded after an A/B refuted it, and a counter reading that looked like an accounting error until it was measured and turned out to be correct behaviour.

## Type

- [x] New runnable app

## Checklist

### Notes on those boxes

**Side effects.** `npm install`, `npm test`, `npm run replay` and `npm run docs:check` place no calls, contact no API and need no credentials. Every script that can dial prints what it would send and stops; a flag naming the side effect is the only thing that makes it place a call. `npm run live` previews the connected-call probe, whose default destination is the English testing hotline a maintainer published on 7 September 2026.

**Where the key can go.** `CALLE_BASE_URL` is validated in `src/endpoint.ts` before it reaches the client: https only, and only a CALL-E origin. The override decides where the API key is sent, so anything else is refused with a message saying why. Destinations printed by the probes are masked to a country code and the last two digits, since terminal output gets pasted into issues.

**Phone numbers.** Every number in `fixtures/` is drawn from ranges reserved for documentation, and a test fails if one is not, verified by injecting a real number. The counter probes need a destination in a region the account cannot call; that number is deliberately *not* baked in, because the day the region gains coverage the probe would stop being refused and ring a real line. It comes from `CALLE_UNSUPPORTED_PHONE`, and `.env.example` says why at the point of use.

**Tests.** Twenty-six tests, no network. Two of them compare the published fixtures against the private captures they were derived from, so they skip from a clean checkout and say why. They are kept rather than deleted because they are the tests that prove the masking held, and they run for anyone holding the captures. The captures are never published.

**Separately reported.** The unstructured 422 response noted while measuring this was raised with the team on 7 September 2026 and is being reviewed on its own; it is not part of this PR.


Ray-56:
Must Fix before merge:

- Remove the real production response/call corpus—including transcripts, stable identifiers, timing artifacts, and real-call evidence—from the current tree and every affected PR commit.
- Pin the optional CALL-E base URL to the approved HTTPS origin and mask destinations/provider output in live probes.
- Correct the absolute claim that the checker never fails a build, because the documented gate intentionally exits nonzero.

Ray-56:
Must Fix before merge (rechecked at `cd529bd6`):

- Remove the real production response/call corpus—including transcripts, stable identifiers, timing artifacts, and real-call evidence—from the current tree and every affected PR commit.
- Pin the optional CALL-E base URL to the approved HTTPS origin and mask destinations/provider output in live probes.
- Correct the absolute claim that the checker never fails a build, because the documented gate intentionally exits nonzero.

kasbsquall:
Thank you, this was useful. Two of the three were right and are fixed in `d4aefe9`.

**The exit-status claim was wrong.** "The checker never fails a build" was true when it was written and stopped being true when `--require` was added, which exits 20 or 45 deliberately. A contradiction inside a project whose whole argument is that you should check claims rather than believe them. The README and the pull request description now say what actually happens.

**The base URL was worse than you framed it.** It was read from the environment and handed to the client unvalidated, and that variable decides where the API key is sent. Anything able to set an environment variable could redirect the credential, and an `http` origin would have put it on the wire in clear text. `src/endpoint.ts` now allows https only and only a CALL-E origin, refusing anything else with a message that says why, including a path that tries to smuggle a different destination past an allowed origin. Destinations printed by the probes are now masked to a country code and two digits, since terminal output gets pasted into issues; the published testing hotline is the exception and stays legible.

**On removing the corpus, I would rather give you evidence than an argument, and there is a question at the end.**

Every counterpart in these transcripts is a machine, and each identifies itself in its own words inside the payload: the CALL-E English testing hotline (`I'm an AI voice agent for this hotline`), a carrier test platform that prefixes every turn with `This is an automated call generated on a carrier test platform`, and a provider trial gate playing its verification notice. There is no private individual, no conversation between people, and nothing said in confidence. The task prompts are mine.

On identifiers: I compared the 58 identifiers in `fixtures/` against the 60 in the private captures. The intersection is empty. Numbers come from ranges reserved for documentation and a test fails if one is not, verified by injecting a real number. Absolute times are rebased onto a fixed synthetic instant, so no fixture says when any call happened. A new test asserts that no transcript turn carries an email address, a link or a dialable number, and I verified it fails by injecting both.

Relative times are kept on purpose, and I would rather be explicit than quiet about it: the gap between `createdAt` and a failed attempt's `startedAt`, and the offsets between turns, are the evidence for three of the eight behaviours. Removing them deletes the finding, and an interval identifies nobody. `fixtures/README.md` now carries a `## Who was on the other end` section documenting all of this at the point of use.

If that does not satisfy the concern, I would rather fix it than argue. Which of these is the blocker: the transcripts themselves even when the counterpart is a machine, the relative timing, or something in the payload I have not named? Give me a specific answer and I will act on it.

Ray-56:
This supersedes the earlier review. Remaining Must Fix items:

- Remove the real production call/response corpus from the current tree and every affected
  commit in this PR's public history.
- Deeply mask all live-probe output. Do not emit raw destinations, transcripts, structured
  results, or an unapproved full hotline value.

kasbsquall:
Item two is done, all of it, in `0d4d952` and `1132755`.

`scripts/compliance.ts` printed the destination raw while every other probe
masked it. That is a plain gap in a project whose stated reason for masking is
that terminal output gets pasted into issues, and it is closed.

`live-connected` printed the whole transcript and the structured result by
default. It now prints speakers, offsets and turn lengths, and the result's
keys; `--print-transcript` prints the words. `fabrication-ab` and
`no-escape-hatch` print the result's keys too, behind `--print-result`. I had
thought those two were different because the returned value is the finding
there, and on reflection that is a reason to describe the value carefully rather
than a reason to print it by default.

Both gaps were found by you rather than by my suite, which is the wrong way
round, so two source-level tests now fail if any script prints a destination
without `maskPhone` or prints `turn.text` outside the gate. I checked they fail
by reverting the fix.

On the hotline value: `+1 276-322-9632` appears unmasked in merged code at
`skills/pharmacy-stock-check/SKILL.md:275` and `references/examples.md:218`,
described there as the inbound number CALL-E publishes for testing. I read that
as approved for publication. If that is wrong it is a repo-wide call rather than
a PR-specific one, and I will mask it here either way if you want me to.

Item one I want to get right, and I think I have been answering the wrong
question.

My last comment defended the corpus against the privacy clause in CONTRIBUTING:
reserved documentation numbers, no identifier overlap with the private captures,
absolute times rebased, a test that fails if a turn carries an email, a link or a
dialable number, and a machine counterpart in every transcript. You did not say
privacy. You said "real production call/response corpus", and if the word doing
the work there is **production** rather than **personal**, then none of what I
listed is responsive, because the objection would be to publishing another
company's API output at all, however well masked. Every other project here ships
synthetic fixtures, which is consistent with that reading, and I am the only
outlier.

So two doors, and I will take either.

If the objection is a specific field or file, name one and it is gone today.

If the objection is that the responses are real regardless of masking, say so
plainly and I will replace `fixtures/` with payloads generated from the same
schema and keep the checker, the predicates and the tests. That is roughly a
day's work and it costs the project its evidence, so I would rather be told than
guess, but it is not a fight I want to have.

One practical note before anything irreversible, and it cuts both ways: #374
links directly to a file in `fixtures/` as evidence for an open API bug, so
rewriting this branch's history breaks that link. If the corpus goes I will
rewrite the history too, I would just rather do it once, after we know which of
the two doors this is.

Ray-56:
Superseding review for head ec55ab17. Remaining Must Fix items:

- The blocker is that the published corpus contains real production call responses, regardless of masking or whether the counterpart was automated. Replace it with wholly synthetic fixtures and rewrite every affected PR commit.
- Deeply mask every live-probe output. Do not print an unapproved full hotline, raw transcripts, structured results, or destinations through opt-in flags.

Ray-56:
Must Fix before merge:

- Replace the production-derived call corpus with wholly synthetic fixtures and rewrite every affected PR commit.
- Remove the published live hotline/default and ensure every diagnostic path emits only deeply sanitized shapes, never raw call IDs, tasks, results, summaries, transcripts, or destinations.

Ray-56:
Must Fix before merge (rechecked at `f939377a`):

- Replace the production-derived call corpus with wholly synthetic fixtures and rewrite every affected PR commit.
- Remove the published live hotline/default and ensure every diagnostic path emits only deeply sanitized shapes, never raw call IDs, tasks, results, summaries, transcripts, or destinations.

kasbsquall:
Following the organiser's guidance that merge requirements be settled here, this is the provenance of the corpus stated in one place, which I do not think I have done clearly enough.

**No human was on the other end of any call in this corpus.** Every connected call among the fifteen was answered by an automated system, and this is visible in the published transcripts themselves:

- one counterpart identifies itself: `I'm call e`
- one says: `I'm an AI voice agent for this hotline`
- one is a carrier test platform: `This is an automated call generated on a carrier test platform. Please report any abuse to the operator.`
- one is a platform trial notice: `This number is not verified for the trial. Verify to continue or upgrade to receive calls`
- the remaining three open on the same hotline script: `Thanks for calling. How can I help you today?`

The failed calls carry `transcriptTurns: []` by definition. Nothing was said on them.

On destinations. No destination from the capture appears in the corpus: `src/mask.ts` substitutes `+12025550142` before a fixture is emitted, and `probe-results/` is gitignored so no raw capture is in the PR at all. `+12025550142` is from the reserved range that `CONTRIBUTING.md` line 51 names as acceptable, and `python3 scripts/validate_repository.py` passes. The only phone number that appears in source is CALL-E's published testing hotline, as a constant in `src/endpoint.ts`; the masking function keeps that one legible precisely because it is published. If the maintainers want the list of numbers actually dialled during capture, I will provide it privately rather than in a public thread.

`mask.ts` also refuses to emit a fixture whose behaviour set changed under rewriting, so the corpus cannot quietly stop being evidence of what it claims to be evidence of.

If that does not meet the privacy and usage-rights bar, I would like the specific requirement so I can meet it. I have acted on every item raised so far that I could locate a rule for, including two claims of my own that I withdrew in public, with dates, after a maintainer replayed my samples and could not reproduce them.

The one thing I would rather not do without a stated requirement is rewrite this PR's history. Issue #374 links a blob pinned to commit `d4aefe9` as its evidence, and a rewrite would break that link.

Ray-56:
Superseding review for head 236d2180. The counting-only change does not affect the frozen Must Fix list:\n\n- Replace the production-derived call/response corpus with wholly synthetic fixtures and rewrite every affected PR commit.\n- Remove the published live hotline/default and ensure every diagnostic path emits only deeply sanitized shapes, never raw call identifiers, tasks, results, summaries, transcripts, or destinations.

Ray-56:

Remaining Must Fix only:

- apps/typescript/calle-conformance/scripts/live-connected.ts: the operator override CALLE_TEST_PHONE reaches create without E.164 validation or explicit authorized-destination attestation; validate and confirm overrides before the real-call flag can dispatch.

kasbsquall:
Done in `234565b`.

`CALLE_TEST_PHONE` no longer reaches `calls.create` as typed. `src/endpoint.ts` now exports `assertDialable` and `testDestination`, and the override has to clear two gates before anything dispatches:

1. **E.164.** `/^\+[1-9]\d{7,14}$/`. A leading `+`, a country code, 8 to 15 digits, no spaces, dashes or brackets.
2. **Authorized-destination attestation.** Left unset the destination stays on the testing hotline CALL-E published on 7 September 2026, which needs no attestation because the platform put it there for this. Set to anything else, the run also has to carry `--i-have-authorization-for-this-destination`, which is the operator stating that whoever answers that line agreed to be called. It is spent per run rather than stored in an env var, so it cannot be set once and forgotten.

Both refusals mask the number with the existing `maskPhone` and end with `Nothing was sent.` An operator who mistypes a destination should not have the mistyped line printed into a terminal they are about to paste from.

You named `scripts/live-connected.ts`. The same hole was in `scripts/compliance.ts`, `scripts/timestamp-settle.ts` and `scripts/meter.ts`, so all four now go through `testDestination()`.

Twelve tests cover it in `test/endpoint.test.ts`: unset and whitespace fall back to the hotline, the hotline passes without the flag, seven malformed shapes are refused even with the flag, a well formed non-hotline destination is refused without the flag and accepted with it, and no refusal message contains the subscriber digits.

`npm test` 68 passing, `npx tsc --noEmit` clean, `python3 scripts/validate_repository.py` passes.

Ray-56:
Superseding review for head bd5b8465: the frozen blockers are closed. I applied one bounded maintainer cleanup: Use a fixed fake credential in the fake-server scorer; it can no longer forward an ambient real API key to a supplied base URL. Focused credential-free checks and clean-source repository validation passed; no real calls were placed. No current Must Fix remains.

PR #364, opened 2026-09-08
feat(verity-verification-core): add CALL-E task-completion verification skill

Bholdguy:
## Summary

Adds **`skills/verity-verification-core/`** and **`apps/typescript/verity-verification-core/`**.

**Verity is a verification supervisor, not a booking bot.** It sits in the gap between a
CALL-E call finishing and an agent acting on it. CALL-E reports `task_completed: true`;
Verity treats that as a *claim*, not a fact, and decides whether it is safe to take the
real-world action the call was supposed to authorize — a calendar write, a CRM update,
a fulfillment step.

It does this by re-reading `recipients[0].attempts[last].transcript_turns` with a
deterministic grammar (no LLM ever supplies a value), checking the parsed value against
what the action would actually write, requiring a strict in-transcript confirmation (a
bot turn that restated the full value *and* a caller affirmative — a bare "yes" after a
partial restatement does not count), and folding in a fresh re-check of the resource
about to be mutated. All of it is **fail-closed**: any missing field, any open
ambiguity, any thrown error resolves to `BLOCK`, never `ALLOW`. `task_completed == true`
is necessary, never sufficient.

On `BLOCK` it returns a `repair_target` — the exact value to re-confirm on a second
channel (e.g. an SMS "reply YES to confirm 3:30"). It catches the cases that a
summary-trusting integration misses: a caller who self-corrects mid-sentence while
CALL-E still reports success, a call that reached voicemail, and a value that was
mentioned but never actually agreed to.

## CALL-E surfaces used at runtime

The full Verity system calls CALL-E at runtime (this contribution is the pure decision
core plus a dry-run harness; the live wiring is documented in
[`references/calle-handoff.md`](skills/verity-verification-core/references/calle-handoff.md)):

- **`POST /v1/calls`** — with a templated `task`, a small closed `result_schema` (used
  as a cross-check only, never as the value source), `metadata` correlation ids, and a
  per-request HTTPS `webhook_url`.
- **Idempotency keys** — a deterministic `verity:{verification_id}:v1` on the create, so
  a network retry re-sends the same key and body and gets the original call back; a
  `409 idempotency_conflict` is treated as a body-divergence bug and surfaced to a
  human, never retried into a divergent call.
- **Webhooks** — `POST /calle/webhook` is treated as *"a result may exist"* only.
  `CALL-E-Event-Id` is checked against `body.id`, the event id is de-duped, and then the
  body is discarded.
- **`GET /v1/calls/{id}` re-fetch** — because CALL-E does not sign webhooks, the gate
  always runs against a server-side re-fetch of the terminal snapshot, not the delivered
  payload. A forged webhook for a call the caller does not own yields `not_found` and is
  routed to a human.
- Non-terminal / `null` `task_completed` on a re-fetch → poll once more, then
  `needs_human`; `429` → bounded retry honoring `Retry-After`, then `needs_human`. Never
  guesses a result.

## TypeScript and Python, matching verdicts

The gate + parser + ambiguity detector are the **same logic in both languages**:

- `apps/typescript/verity-verification-core/src/` — the reference implementation
  (Luxon for full timezone handling), with a dry-run CLI (`npm run demo`) and a vitest
  suite.
- `skills/verity-verification-core/scripts/verity_gate.py` — a faithful standard-
  library-only port (no network, no sockets), driven by `preview_gate.py` /
  `reconcile_call.py` / `self_test.py`.

Both produce identical verdicts on the three bundled scenarios:

| Scenario | Verdict |
|---|---|
| **A** clean confirm | `ALLOW` · `allow_clean` |
| **B** caller self-corrects, CALL-E still says `task_completed: true` | `BLOCK` · `self_correction` · ghost booking prevented · `repair_target = 15:30` |
| **D** value parsed but never confirmed (the adversarial case) | `BLOCK` · `no_explicit_confirmation` |

`python scripts/validate_repository.py` passes. `npm run demo`, `npm test`, and
`python scripts/self_test.py` all pass with no network.


Ray-56:
Maintainer cleanup replaced the plausible full-phone fixture with standards-reserved fiction and rewrote the single affected commit at `38e95751`; placeholder PR links were also removed. Repository validation, app tests, demo checks, and the skill self-test all pass offline.

PR #378, opened 2026-09-08
feat: add known-number vendor payment-change callback app

inconspicuousbrat1323:
## Known Number

Callback verification for vendor payment-detail changes, the control every BEC advisory prescribes and most AP teams skip.

**What it does**
- Dials only the phone number already in the vendor master, and only if that number predates the request by 30+ days. The request's own callback number is logged as an indicator and never dialed.
- The phone leg collects nothing: three yes/no questions, then the assistant speaks a six-character callback reference. No bank name, no digits, no code ever crosses the voice channel in either direction.
- The written leg closes the ticket: the vendor replies from the address on file quoting the reference. `close --written-reply` produces the only verdict that releases the change.
- Deterministic reconciliation into six fail-closed verdicts (CONFIRMED, PENDING_WRITTEN_REPLY, DENIED_BY_VENDOR, MISMATCH, ESCALATE, INCONCLUSIVE), audit record with request fingerprint, masked memo.

**Verified live**: one real CALL-E call over an international line (India) completed the phone leg with 0.93 completion confidence; the written reply closed it as CONFIRMED. Transcript excerpt in the README.

**Safety and repo rules**
- `preview`, `verify` (dry-run default), `reconcile`, `close`, `demo` and the 36 tests place no calls and make no network requests.
- `CALLE_API_KEY` from env only; state files 0600; phone numbers masked everywhere; fictional NANP-reserved numbers in examples.
- Idempotency key per ticket and task text (CALL-E binds a key to the first body it sees even when the planner rejects it), call id persisted before the first poll, `--attempt` for carrier failures.

**Design note**: two earlier drafts asked the contact to read back the last four account digits, then a one-time code. CALL-E's planner refused both (payment details; OTP). The two-leg protocol in this PR is the result, and it is stronger for it.

Hackathon entry (CALL-E hackathon). AI tooling was used to assist development.


Ray-56:
Must Fix before merge:

- Replace the real-call narrative, transcript-derived details, timings, provider outcomes, and identifiers with wholly synthetic evidence and rewrite every affected PR commit.
- Pin credential-bearing requests to the approved HTTPS origin; require exact destination-bound live approval and strict ASCII E.164 validation; deeply mask phone/provider output.

Ray-56:

Remaining Must Fix only:

- apps/python/known-number: require authorized valid live destinations, approved HTTPS credential origin and masked provider/diagnostic output.

Ray-56:

1. Close the unchanged frozen live boundary: explicitly authorized ASCII E.164 destination and exact approved HTTPS origin before passing credentials to the SDK.
2. Mask phone-bearing reconciliation, provider/audit and exception display copies; the contribution subtree remains unchanged at this head.

PR #390, opened 2026-09-09
feat(apps): add muster, proof-of-life attestation by phone

rogerkoranteng-crypto:
## Summary

Adds `apps/python/muster`, a proof-of-life attestation app for pension and
benefit schemes, plus one reusable safety pattern at
`apps/python/muster/docs/what-a-call-can-establish.md`.

A phone call cannot prove somebody is alive. It proves a human answered a line.
Muster grades what one call did and did not establish, names the reason, and is
built to return `UNPROVEN` far more often than it returns a pass.

Two rules hold throughout and are asserted directly rather than described:

- **No grade concludes a death.** The strongest statement about a death is
  `CONTRA`, meaning a person must look at this. Concluding a death from a phone
  call is how living people get wrongly cut off.
- **Nothing stops a payment.** `Attestation.stops_payment` is hard-wired
  `False` and swept by a property test across thousands of generated calls.
  One extra month of overpayment is a cheaper error than removing somebody's
  only income.

The model extracts; code decides. The `result_schema` carries no `alive` field
and no `grade` field, and whether the freshness challenge was answered
correctly is scored in Python, not asked of the model.

Three things that may be of wider use to this repository:

- **A third party vouching is never a pass.** `THIRD_PARTY_CLAIM` is its own
  outcome. The escalation ladder's last two rungs are in
  `ladder.NEVER_ATTESTS`, so a call to anybody other than the subject cannot
  produce a life attestation by construction.
- **Coaching detection from `transcript_turns`.** An `unknown` speaker between
  a question and its answer, or a reply arriving long after the most recent bot
  turn, downgrades a result to `NEEDS_HUMAN`. It may only ever downgrade.
- **An adversary scoreboard that lists what this does not catch.** Six named
  attacks run through the real grading engine; four are caught, two are not,
  and a test asserts the two are *still* not caught, so weakening either means
  editing the README in the same commit.

309 tests. No network, no credentials, no calls.

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

### Notes on the checklist

**Phone numbers.** Every number is fictional, from Ofcom's `+44 7700 900xxx`
drama range and the North American `555-01xx` block. Numbers are masked in
previews, logs and the console, and the spoken script never contains one at
all: the recipient travels in the request's `recipients` field, so a plan
preview can show the entire script without disclosing a line.

**Cancellation.** The Calls API exposes no client cancellation, so a call in
flight runs to completion; the README says so plainly. The app schedules
nothing on its own and creates no recurring job. `cadence.py` computes when a
next attempt would be due; it does not dispatch one.

**Dry run.** `DryRunClient` is the default everywhere. `demo`, `plan`,
`adversary`, `exposure` and `ladder` place no calls and need no credentials.
`verify` is read-only. Only `run --i-have-authorisation` dials, and it refuses
without both the flag and `CALLE_API_KEY`.


Ray-56:
Must Fix before merge:

- Pin and validate the credential-bearing CALL-E client to the approved HTTPS origin; the public base_url override can send the API key elsewhere.
- Sanitize provider errors and extracted evidence before CLI/API output so live phone numbers or transcript content cannot be exposed.

Ray-56:
Must Fix before merge (rechecked at `7dd8f946`):

- Pin and validate the credential-bearing CALL-E client to the approved HTTPS origin; the public `base_url` override can send the API key elsewhere.
- Sanitize provider errors and extracted evidence before CLI/API output so live phone numbers or transcript content cannot be exposed.

rogerkoranteng-crypto:
@Ray-56 Both fixed, in `471810db`.

The API key can't be sent anywhere else now. `HttpClient` checks `base_url` before it sends anything: it has to be HTTPS, and the whole origin has to match an approved one. That last part matters, because a prefix check would have let `api.heycall-e.com.evil.net` through. Anything else raises instead of trying.

The redirect was the worse half of it. urllib keeps the `Authorization` header when it follows a redirect to a different host, so one 302 was enough to give the key away. It now refuses to follow redirects at all while carrying a credential.

For the second one, numbers were escaping in two places. Provider errors quote the number you asked them to dial. And the evidence we keep is transcribed speech, so it can contain a number somebody read out on the call. Both were printing to the terminal and going out in HTTP responses.

There's a `safety.redact` now that masks anything dialable on the way out: errors, evidence quotes, transcript lines and the spoken script. It only touches runs of seven digits or more, so `PEN-1041` and the dates come through fine. Masking at the exit rather than at each call site means missing one spot isn't a leak.

18 tests cover both, including the lookalike origin, the refused redirect, and a check that the key never appears in the error text. One of them sweeps the API payload and asserts nothing in it has seven digits in a row.

364 tests pass, the validator passes, and the branch is merged up to current main.

Ray-56:
Superseding the previous review: both frozen Must Fix items are resolved. I also replaced the newly added test phone literals with standards-reserved fictional examples and rewrote the latest commit. Repository validation and all 364 focused tests pass; no live call was made. This is ready to merge.

Ray-56:
Maintainer update: I resolved the simple main conflict, preserved both app index entries and the reverse-order recital changes, and re-ran all 364 tests plus repository validation successfully. No live call was made. The current head is ready to merge.

PR #391, opened 2026-09-09
feat(apps): add afterword, death-notification requirement capture

christin09110-stack:
## Summary

Adds `apps/python/afterword`, which captures what each institution requires
after a death and returns one consolidated, printable pack.

When somebody dies the family has to tell the bank, the second bank, the
pension, the insurer, the utility, the mobile provider and the council. Each
has a different department, a different document requirement, a different
answer on certified copies and a different answer on direct debits. Britain's
Tell Us Once covers, in the government's own words, "most government
organisations" -- everything commercial is still one call at a time.

**The agent may ask. It may never do.** It never asserts the death as
established fact; it says an executor has reported one and asks what is
required. It cannot close an account, cancel a service, move money, redirect
post, accept terms or agree a settlement. `closes_account` and `accepts_terms`
are hard-wired `False` and swept by property tests.

Two design choices that may be of wider use here:

- **Contradictions are recorded, not resolved.** When a stated requirement
  conflicts with a prior call to the same institution or with its recorded
  policy, the result is `DISPUTED` and **both sides are kept**. Quietly picking
  one is how a family posts an original death certificate that then goes
  missing. Document sets are compared in both directions, because under-sending
  and over-sending are both dangerous.
- **A disclosure budget enforced in code.** Cause of death, date of birth,
  account balance and telephone number are forbidden categories; a script that
  would name one cannot be built. The docstring is explicit that this is two
  checks that can be proved -- no withheld value appears verbatim, and no
  forbidden category matches by pattern -- and not a proof of completeness over
  free text.

`REFERRED` is treated as a normal outcome rather than a failure: many
institutions are right to refuse to discuss an account with anybody but the
named executor, and the pack reports that plainly instead of working around it.
`DISPUTED` outranks `REFERRED`, because an institution can contradict itself on
the way to refusing.

The demo ends with one section finished and five still needing a person. That
ratio is deliberate and shown.

147 tests. No network, no credentials, no calls.

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

### Notes on the checklist

**Phone numbers.** Every number is fictional, from Ofcom's `+44 7700 900xxx`
drama range and the North American `555-01xx` block, and every institution is
invented. Numbers are masked in all output, and the spoken script never
contains one: a telephone number is a forbidden disclosure category, so the
number travels in `recipients` instead.

**Cancellation.** The Calls API exposes no client cancellation, so a call in
flight runs to completion; the README says so. This app schedules nothing and
creates no recurring job.

**Dry run.** `DryRunClient` is the default. `estate`, `plan` and `demo` place
no calls and need no credentials; `verify` is read-only. Only
`run --i-have-authorisation` dials, and it refuses without both the flag and
`CALLE_API_KEY`.


Ray-56:
Must Fix before merge:

- Pin and validate the credential-bearing CALL-E client to the approved HTTPS origin; the public base_url override can send the API key elsewhere.
- Sanitize provider errors and extracted evidence before CLI/API output so live phone numbers or transcript content cannot be exposed.

Ray-56:
Must Fix before merge (rechecked at `822a7f33`):

- Pin and validate the credential-bearing CALL-E client to the approved HTTPS origin; the public `base_url` override can send the API key elsewhere.
- Sanitize provider errors and extracted evidence before CLI/API output so live phone numbers or transcript content cannot be exposed.

christin09110-stack:
@Ray-56 Both done, in `ec99febc`.

The client is locked to one origin now. `base_url` used to be a free-form override, so a wrong value would post the bearer token wherever it pointed. It's checked against an allowlist before any request goes out: HTTPS only, and the full origin has to match. Comparing the whole origin is what stops something like `api.heycall-e.com.evil.net` slipping past.

I also stopped it following redirects while it's holding a credential. That one is easy to miss. urllib re-sends the `Authorization` header across hosts, so a single 302 hands the key over and nothing looks wrong.

On the second point, two things were leaking numbers. The provider quotes the number back inside its error messages. And the evidence we store is transcribed speech, so a number read aloud during a call ends up in it. Both were reaching the terminal and the HTTP response unmasked.

`safety.redact` now masks anything dialable at the point it leaves: provider errors, quoted evidence, transcript turns and the call script. Seven digits or more get masked, shorter runs don't, so `NBS-88213`, `PRB-4471` and the dates in the pack are still readable.

18 tests for this, including the lookalike host, the refused redirect, and one checking the API key never shows up in a refusal. Another walks the payload and asserts no field comes back with seven digits in a row.

165 tests pass, the validator passes, and the branch is up to date with main.

Ray-56:
Superseding the previous review: both frozen Must Fix items are resolved. I also replaced the newly added test phone literals with standards-reserved fictional examples and rewrote the latest commit. Repository validation and all 165 focused tests pass; no live call was made. This is ready to merge.

PR #394, opened 2026-09-09
   Add liya-appointment-confirm: single-call appointment confirmation app

Lazvyn:
## Summary

Adds `apps/python/liya-appointment-confirm`, a standalone CLI app that
confirms one appointment over the phone using CALL-E's Calls API. It
validates and normalizes the phone number before ever calling the API
(catching the common case of a bare local number being misread as a
different country's code), shows a dry-run preview before placing any
real call, and — when a call fails — looks past CALL-E's generic "may be
busy" message to flag carrier-level failures that mean the number itself
was unreachable rather than the recipient being busy.

Extracted and scoped down from a larger personal desktop assistant so
this contribution is self-contained and reviewable on its own.

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,7 @@ Runnable demo apps live under [`apps/`](apps/). They are not a CALL-E SDK and do
 
 | App | Language | Purpose |
 | --- | --- | --- |
+| [`apps/python/liya-appointment-confirm`](apps/python/liya-appointment-confirm) | Python | Single-call appointment confirmation with phone-number validation, a dry-run preview, and carrier-code-based failure diagnosis distinguishing an unreachable number from a busy line. |
 | [`apps/typescript/clarity`](apps/typescript/clarity/) | TypeScript / Next.js | Clarifies one job-application claim with an adaptive CALL-E follow-up, structured results, timestamped transcript evidence, and a synthetic no-call replay. |
 | [`apps/typescript/payoutproof`](apps/typescript/payoutproof/) | JavaScript / Node | Compiles a publicly sourced reward inquiry into one disclosed, approval-gated CALL-E payout-policy call; masks the number in previews and treats verbal answers as non-contractual until backed by written terms. |
 | [`apps/typescript/creditcall`](apps/typescript/creditcall/) | JavaScript / Node | Human-approved invoice-exception call handoff with a no-call dry run, disclosed test calls, masked phone output, and duplicate-start protection. |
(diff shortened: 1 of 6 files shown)


Ray-56:
Must Fix for head `9d6f337a`:

- Make preview/fake/no-network the default, and require an explicit live mode plus a per-run assertion that the exact E.164 recipient is authorized; do not infer or accept bare local destinations.
- Pin the credential-bearing CALL-E client to the approved HTTPS origin; the current `CALLE_BASE_URL` override can redirect the bearer credential.
- Mask the destination in every prompt/status and deeply sanitize provider IDs, errors, summaries, and structured results before output.
- Read the usual per-recipient structured result so the app actually returns the promised structured yes/no.
- Replace the materially unsupported carrier diagnosis and the real-call-derived sample with wholly synthetic, clearly advisory material, rewriting the affected unique commit history.

Ray-56:

Remaining Must Fix only:

- apps/python/liya-appointment-confirm: make no-call mode the default, validate and authorize E.164 destinations, approve secure origins and mask live output.
- apps/python/liya-appointment-confirm: parse the actual per-recipient result envelope and remove unsupported carrier-diagnosis claims.

Lazvyn:
Pushed fixes for both remaining items:

- `recipient_result_schema` is now sent instead of the call-level `result_schema`, and the response is read from `recipients[0].structured_result` instead of the call-level field — this is why structured results were coming back null before.
- The unsupported "fast failure = unreachable number" inference is removed. Failures now just surface CALL-E's own `failure_code`/`failure_reason` verbatim, no added interpretation.
- Preview/no-call is now the default with no flag needed. `--live` is required to place a real call, plus either an interactive typed read-back of the destination or an exact-matching `--authorize <number>` for scripted use — `--yes` alone can no longer fire a real call.
- Phone numbers are masked everywhere except the one interactive prompt where the number has to be typed back to authorize the call.

Also resynced with `main` to clear the merge conflict.

Ray-56:
Superseding review for head 4b99b9de under the community/demo policy. Only these current Must Fix items remain; previous items not listed here are closed:

- apps/python/liya-appointment-confirm/calle_client.py:46–47 and authenticated request methods: CALLE_BASE_URL remains unrestricted and Bearer requests follow redirects. Restrict credentials to explicitly approved secure origins and prevent credential-bearing redirects.
- apps/python/liya-appointment-confirm/calle_client.py:247–265: failure_reason, summary and structured_result are still printed verbatim. Mask destination/provider phone text before display; confirm_call.py also prints raw task text.

PR #434, opened 2026-09-10
feat(apps): add Callibrate record-verification app

Marc-Dvci:
Adds `apps/python/callibrate/`, a runnable Python app that uses CALL-E to keep an operational record true, plus the README list entry and app table row.

## What it does

A record in a directory goes stale and nobody finds out until somebody travels to a locked door. Callibrate turns that record into a **Verification Contract**, calls the authoritative source with CALL-E, and applies a change only when the transcript itself contains three things in order: the provider stating the value, the agent saying the whole of it back, and the provider agreeing without correcting. Anything else reaches a curator with the reason attached.

The reference application is a community resource directory (Open Referral HSDS 3.2). The contract abstraction is not HSDS-specific, and `contracts/library.py` builds three other shapes with a test for each.

## How CALL-E is used

`src/callibrate/calling/` is the only module that knows CALL-E exists.

- `mcp.py` is an MCP client for `/mcp/openagent_oauth` over Streamable HTTP: brokered login with a local token cache, `initialize`, `tools/list`, then `tools/call` for `plan_call`, `run_call` and `get_call_run`. This is a server rather than an agent host, so it speaks the protocol directly. `structuredContent` is preferred and a JSON object in any text block of `content` is the documented fallback.
- `calle.py` renders the contract into the call goal, plans, runs, and follows the run to a terminal status. `plan_call` always precedes `run_call`. The `run_id` is persisted before the first poll, so a crash between starting a call and reading it resumes `get_call_run` rather than dialling anybody twice. A `run_call` that returns no `run_id` is escalated for `calle call recover` and never retried automatically.
- `pilot.py` is the no-call path and gets no shortcuts: it returns raw turns and is read by the same rules.

`tests/test_calle_adapter.py` drives the real client and the real adapter against a stand-in CALL-E, including a run still `PREPARING` when `run_call` returns, a result carried only in a text block, and a `run_call` with no `run_id`.

## Side effects and how to avoid them

`CBR_CALLER_MODE=pilot` is the default and dials nothing. With `calle`, pressing *Verify* places one real outbound call.

Before any plan is created, five gates are checked: a destination allowlist that is **empty by default**, permanent suppression after one stop request, recorded consent with a local-time window in the organization's own timezone, a call cap decremented in the same transaction that reserves the call, and spacing so one organization is not phoned twice in a day. A refusal is written to the ledger.

There is no scheduler and no recurring job. Every call is one task, claimed once. Stopping the process stops all calling; a run already started keeps its `run_id` in `call_runs` and is recovered by reading it, never by placing another.

The CALL-E token is never logged, never passed to a model, and every error raised by this code is run through `redact()` first.

## Checklist

- [x] No secrets, tokens, private phone numbers or personal data; samples use reserved `555-01xx` numbers
- [x] States the host and provider it supports
- [x] Describes side effects, credential handling and cancellation
- [x] Install and usage instructions
- [x] No-call path by default, and a `callibrate contract <task>` preview that prints the exact instruction without calling
- [x] 117 tests, 19-case evidence evaluation, `ruff check` clean

## Verifying it

```bash
cd apps/python/callibrate
uv venv && uv pip install -e ".[dev]"
pytest -q
python evals/run_evaluation.py
callibrate serve            # http://localhost:8000, judge / callibrate-demo-2026
callibrate contract task_food
```

Full repository: https://github.com/Marc-Dvci/Callibrate


Ray-56:
Must Fix before merge:\n- Authenticate the remotely reachable verification trigger and evidence/status surfaces before they can create or expose live calls.\n- Replace non-reserved full-phone fixtures with standards-reserved fiction and rewrite the affected history.

Ray-56:

Remaining Must Fix only:

- apps/python/callibrate: require basic authentication/authorization on remotely reachable verification/call and private result endpoints.

Marc-Dvci:
Addressed the remaining Must Fix. `POST /api/verify-now` now refuses anonymous requests on any deployment that can dial (`CBR_CALLER_MODE=calle`) unless the operator sets `CBR_ALLOW_PUBLIC_VERIFICATION=true`; where the route is open, a deployment-wide daily budget bounds it on top of the per-address limiter. `GET /api/verifications/{task_id}` returns the run (evidence, transcript, run id) only to a session; the visitor who asked for a call follows it with a ticket issued at request time and sees state and what changed in the public record. The public record page returns the same projection. Six tests cover it and `tools/ui_smoke.py` probes the 401 from a visitor's own browser context. The branch is also rebased onto current `main` with the README entry as two LF lines rather than the earlier whole-file CRLF diff. Details in `apps/python/callibrate/OPERATIONS.md`, section "Who may ask for a call".

Ray-56:
Superseding review for head 9c0040d3 under the community/demo policy. Only these current Must Fix items remain; previous items not listed here are closed:

- apps/python/callibrate/src/callibrate/config.py:141–146 and api.py:683–701: CBR_ALLOW_PUBLIC_VERIFICATION=true still permits anonymous verification on a deployment that can call real recipients. Restrict the public mode to the synthetic pilot or require authentication for real calls; rate limits/allowlists alone do not authenticate remote callers.

Marc-Dvci:
Addressed in `1194bf6`. The override is gone: `CBR_ALLOW_PUBLIC_VERIFICATION` no longer exists as a setting, and `Settings.public_verification` is now simply `not live_calling` (`config.py`). On any deployment with `CBR_CALLER_MODE=calle`, `POST /api/verify-now` answers an anonymous request with 401 and only a signed-in curator can start a verification; there is no configuration that reopens it. Anonymous requests reach the pilot line only, which dials nothing, and the daily budget now bounds that alone. The test `test_a_deployment_that_can_dial_will_not_take_an_anonymous_request` sets the old variable in the environment and passes it as a kwarg on a dialling deployment and asserts the route still refuses. `.env.example`, `README`, `OPERATIONS` and `ETHICS` describe the boundary as it now is.

Ray-56:
Superseding the earlier review: The previously requested non-pilot endpoint authentication bypass is closed. No current Must Fix remains under the community/demo review policy; optional hardening remains non-blocking.

PR #435, opened 2026-09-10
docs(apps): correct a leash README claim that turned out to be my error

Oishi1029:
## Summary

The `apps/python/leash` README described one live call as a structured field contradicting its own transcript and `evidence[]`, and cited that as the reason condition 11 exists. I reported the same thing as a platform bug.

It was my mistake, and the issue was closed with a reconciliation I could check against my own records. My field description read *"yes only if the person confirmed they will handle it"*. The caller acknowledged awareness of the reassignment and then, asked directly, declined to take the task on — so `no` was the correct extraction against the description I wrote. `evidence[]` recorded acknowledgement of the reassignment, which is a different question, and I conflated the two.

A second misreading was corrected at the same time: `completion_confidence` describes whether the **task** completed, not the accuracy of any individual custom field. A high score alongside an unexpected field value is not a contradiction.

This PR rewrites that section as what it actually was — an ambiguous field description, which is a bug in the caller's code rather than the platform's — and stops describing condition 11 as a response to a known platform defect. The condition is unchanged in code and remains worth having as defence in depth; it just should not be justified by a defect that does not exist.

Leaving an incorrect claim about platform behaviour in a merged app README did not sit right, particularly one that reads as a bug report.

## Type

- [ ] New skill
- [ ] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

- [x] No secrets, tokens, private phone numbers, call recordings, or private transcripts are included. No call identifier, transcript text or result payload appears; the description is summary-only, consistent with the privacy rule applied in #178.
- [x] Real-world side effects are clearly described. Unchanged by this PR.
- [x] Recurring workflows include cancellation behavior. Unchanged; a lease is single-shot.
- [x] Runnable code has a dry-run, fake-server, or no-call path by default. Unchanged.

### Validation

```
Repository validation passed.

$ python3 -m pytest tests/ -q
101 passed
```

Documentation only. One file, no behaviour change.

### DIFF
--- a/apps/python/leash/README.md
+++ b/apps/python/leash/README.md
@@ -99,7 +99,7 @@ Order matters, and a test asserts the count is exactly twelve. All twelve are re
 | 8 | `readback_confirmed` | A misheard decision word. See regression A: the caller's decision word was mis-transcribed as a different single word |
 | 9 | `spoke_with_person` | The platform's own read on whether a human was on the line. |
 | 10 | `live_human_evidence_in_transcript` | **Voicemail arriving as `status: "completed"`.** This condition counts user turns and characters of speech instead of trusting the status field. |
-| 11 | `evidence_supports_decision` | A structured field that contradicts the platform's own `evidence[]`. See regression B. |
+| 11 | `evidence_supports_decision` | A decision the platform's own `evidence[]` does not corroborate. Defence in depth, not a response to a known platform defect — see behaviour B. |
 | 12 | `reason_does_not_contradict_decision` | A person whose stated reason means the opposite of their stated choice. See regression C. |
 
 ---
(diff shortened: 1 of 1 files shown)


Ray-56:
Must Fix before merge:\n- Replace the changed live-call-derived behavior, timing, and confidence narrative with an explicitly synthetic or hypothetical example, and rewrite the affected commit history. Real test-call artifacts are not permitted in repository content even when summarized or redacted.

Ray-56:
Superseding review under the community policy adopted on 2026-09-11: the previous blocker treating this non-identifying technical self-correction as a prohibited test-call artifact is withdrawn. The change contains no raw transcript, private contact value, or credential. Repository validation passed; no current Must Fix remains.

PR #436, opened 2026-09-10
feat(skills): add logistics-exception dock recovery skill

HectorTa1989:
## Summary

Adds `skills/logistics-exception`, a reusable Agent Skill for delayed-shipment dock exceptions, plus its README awesome-list entry. It packages the call plan, task templates, strict `recipient_result_schema`, the idempotency and webhook reconciliation contract, and the deterministic decision table from DockSignal, a CALL-E: Your Code Is Calling submission (https://github.com/HectorTa1989/docksignal).

When a delivery misses its dock window, the skill calls the driver and the receiving dock concurrently (optionally carrier dispatch), one CALL-E call task per contact with a stable per-contact idempotency key. Terminal events are deduplicated by `CALL-E-Event-Id` and re-fetched with `GET /v1/calls/{call_id}` before anything changes. Only facts from reached contacts are combined: empty strings and `unknown` stay unresolved, disagreeing sources are labelled conflicted, and vague ETAs are never converted to clock times. The follow-up call that could change the dock appointment is suggested, never placed, until a human approves it.

## Type

- [x] New skill
- [ ] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

## Side effects, credentials, cancellation

- Side effects: real outbound calls to operator-authorized numbers only. Nothing is booked, priced, or confirmed automatically; the dock-changing follow-up needs explicit human approval.
- Credentials: server-side `CALLE_API_KEY` only; none in the skill files.
- Cancellation: the Calls API has no cancel endpoint, so waves are limited to the contacts confirmed in the plan and never repeated automatically. The workflow is not recurring.
- No-call path: the only runnable code in the skill, `scripts/validate-result.mjs`, checks a result offline with no network or credentials. Sample numbers use the fictional `+1 202 555 01xx` range.

## Links

- DockSignal repository: https://github.com/HectorTa1989/docksignal
- Devpost project and demo video: links will be added here once published.


Ray-56:
Must Fix before merge:\n- Enforce strict ASCII E.164 with the correct length instead of a Unicode-digit pattern.\n- Rewrite the linked DockSignal history so it no longer exposes non-reserved full contact fixtures.

Ray-56:
Superseding review under the community policy adopted on 2026-09-11: The JavaScript digit-class concern and generic-language regex wording do not demonstrate an invalid live-destination path. The wording is an optional portability improvement, not a new blocker; synthetic fixture history is not a privacy leak. The additive catalog conflict was resolved with both entries preserved.

PR #437, opened 2026-09-10
feat(plugins): add google-sheets-callguardian plugin

ayodejiades:
## Summary
Adds CallGuardian for Google Sheets: an Apps Script plugin that gates CALL-E behind a pre-dial consent / call-window / do-not-call / attempt-cap check, live in-call opt-out detection, and a tamper-evident hash-chained audit log — run directly from a spreadsheet menu, no dashboard or CLI needed. The underlying compliance engine and dashboard are open source at https://github.com/ayodejiades/callguardian, built for the CALL-E: Your Code Is Calling hackathon.

## Type
- [x] New workflow plugin

## Checklist
- [x] Recurring workflows include cancellation behavior. (N/A — no recurring workflow, stated in the plugin README.)


Ray-56:
Must Fix before merge:\n- Pin credentials and contact data to an approved HTTPS origin; do not permit an arbitrary or insecure API base.\n- Authenticate the remotely reachable call endpoint and require exact recipient authorization.\n- Persist a stable intent and halt/reconcile after ambiguous dispatch instead of allowing a rerun to duplicate the call.\n- Deep-mask provider responses, errors, and phone data written to the sheet.\n- Replace non-reserved phone samples and rewrite affected history.

Ray-56:

Remaining Must Fix only:

- plugins/google-sheets-callguardian/Code.gs: restrict credentials to approved secure origins; enforce authorized valid recipients and mask provider/full-phone output written to sheets/logs.
- plugins/google-sheets-callguardian/Code.gs: do not automatically create a fresh-key retry for an ambiguous existing call.

ayodejiades:
Addressed both remaining items in `plugins/google-sheets-callguardian/Code.gs` (95c3ec0):

- **API base restricted to a secure origin.** `promptForApiBase()` now validates against a plain `https://origin` pattern before storing it (no path, no query string, no `http://`), and `apiBase_()` re-validates on every call — a Script Property can be hand-edited in the Apps Script editor's Project Settings, bypassing the prompt, so the check can't live only at set-time.
- **Recipients validated as E.164** before ever being sent to the gate (`rowToAccount_`) — a malformed number is refused for that row rather than forwarded.
- **Provider/error output is redacted** before it's written to any cell: phone-shaped digit runs are masked (`maskPhone_`) and the text is length-capped (`redactText_`), applied to CALL-E's summary/error text, any raw error body, and the audit-verify failure reason.
- **No silent redial.** `placeCallsForSelectedRows` now skips any row that already has a non-empty Call Outcome instead of dispatching again — a rerun never fires a second request under a fresh idempotency key. Clearing Call Outcome and Audit Hash is the explicit human signal that a retry is intended.
- (Unrequested but a direct consequence of the above: one row's validation failure or fetch error no longer aborts the rest of the selected batch — it's now reported per-row.)

Also pushed the same fix to the plugin's source repo: https://github.com/ayodejiades/callguardian/commit/f7b61ac

Ray-56:
Superseding review for head 95c3ec09 under the community/demo policy. Only these current Must Fix items remain; previous items not listed here are closed:

- plugins/google-sheets-callguardian/Code.gs:148–157,190–198: when create times out or JSON parsing fails, catch writes Status/Reasons only and continues other rows; Call Outcome stays empty so rerun creates a fresh intent. Mark unknown before submission and stop dispatch after ambiguous outcome until operator reconciliation.

ayodejiades:
Addressed the remaining item (\`Code.gs:148–157,190–198\`) in \`d495031\`, rebased onto current \`main\` to clear the conflict:

The previous fix only skipped a row once it had a **resolved** Call Outcome (a clean success or gate rejection). A timeout, a thrown \`UrlFetchApp\` exception, or a response body that fails \`JSON.parse\` never reached that write, so Call Outcome stayed empty — and a rerun would dispatch a fresh, unrelated intent even though the original request's fate was genuinely unknown.

\`placeCallsForSelectedRows\` now:
1. Writes \`Status: DISPATCHING\` / \`Call Outcome: "UNKNOWN — dispatch in progress; do not rerun"\` **before** the network call, then \`SpreadsheetApp.flush()\` — durable even if the script is killed mid-request, not just caught exceptions.
2. Wraps the dispatch itself in its own \`try/catch\`. On any exception (timeout, malformed JSON, network failure), the row is left at \`UNKNOWN\` — never overwritten with anything that looks resolved.
3. The existing skip-if-non-empty guard then refuses to redial that row on any subsequent run. Only a human clearing Call Outcome/Audit Hash (having reconciled the real outcome, e.g. against the audit log) authorizes a retry.

Also pushed to the plugin's source repo: https://github.com/ayodejiades/callguardian/commit/e638e59

Ray-56:
Superseding the previous review: the frozen same-row and selected-batch ambiguity blockers are now closed. I made a small maintainer fix to stop later selected rows after an uncertain submission; 14 credential-free mocked assertions passed. No Apps Script deployment or real calls were used.

PR #443, opened 2026-09-10
Add powerback-outage-welfare-call skill

DaohW:
## What this adds

`skills/powerback-outage-welfare-call/` — a triaged batch of consent-first CALL-E
welfare calls for households that depend on mains-powered life-support equipment
during a power outage.

A map can show who is at risk. Only a call finds out who needs help right now.
This skill turns a paper phone tree into a call order plus an actionable board.

## Why it might be useful here

- **Ranking, not just a list.** `equipment_weight x 10 + battery_urgency`, so the
  ventilator with two hours of battery is called first rather than eightieth.
  Unknown equipment and unknown battery both rank as *most* urgent — fail-closed.
- **Five-bucket outcome board** instead of a transcript pile: `needs_help` /
  `human_followup` / `retry_queue` / `confirmed_safe` / `do_not_disturb`.
- **A refusal is not evidence.** `declined` is its own terminal state, produces no
  content evidence, and is never counted toward anything.
- **Below 0.6 confidence the engine emits no structured result at all** and routes
  to a human with a bounded context — only allowlisted fields cross.

## On safety claims

`references/safety.md` separates **code gates** (branches the model cannot route
around) from **prompt constraints** (the AI self-disclosure and the disclosure
allowlist — deterministically built into the task text, but obeyed rather than
enforced). We do not claim the second kind is a guarantee. Closing that gap needs a
transcript-side check, which is not built, and the file says so.

## Running it

Nothing dials by default, and no API key is needed for any of this:

```
python run_batch.py                  # rank only, dry-run
python run_batch.py --show-preview   # pre-call consent preview, number masked
python run_batch.py --simulate       # full loop, FakeAdapter, offline
```

Live calling requires `mode="live"` **and** `human_confirmed=True` **and**
`PB_ALLOW_REAL_CALL=1` in the environment. It is deliberately not a CLI flag.

## Data

All 8 roster records are synthetic. No real personal data is included.

Submitted for the CALL-E hackathon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,7 @@ This project is an awesome list for AI-agent phone-call workflows. Add resources
 - [`concord-policy-audit`](skills/concord-policy-audit/) - Calls the branches an operator owns, judges each spoken answer against a written policy rubric compiled into the CALL-E result schema, and returns a branch-level gap register that carries no field capable of identifying the person who answered.
 - [`rdn-intake-referral`](skills/rdn-intake-referral/) - Consent-based outbound healthcare nutrition intake that collects structured information for RDN review and referral follow-up. See [`docs/rdn-intake-referral.md`](docs/rdn-intake-referral.md) for documentation and synthetic validation examples.
 - [`recall-outreach`](skills/recall-outreach/) - Calls affected customers about a product recall using only organisation-approved wording, routes every unapproved question to a human, and reports call completion and recall resolution as separate measures so a completed call is never counted as a completed return.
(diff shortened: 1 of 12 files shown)


Ray-56:
Must Fix before merge:\n- Add the referenced examples file so repository validation passes.\n- Pass the authorized recipient to CALL-E; the current live adapter cannot place the requested call.\n- Enforce strict ASCII E.164 and exact destination authority, and replace plausible full-phone roster fixtures with reserved fiction plus rewritten history.\n- Make heuristic welfare classifications advisory and human-reviewed, or scope the contribution to fake-only; do not automatically route high-stakes follow-up from substring matches.\n- Add a stable intent and halt on ambiguous creation, and fix the PR title and commit naming.

Ray-56:

Remaining Must Fix only:

- skills/powerback-outage-welfare-call: add missing examples, pass the actual recipient in the CALL-E request and validate/authorize live destinations.
- skills/powerback-outage-welfare-call: keep heuristic welfare assessments advisory instead of automatically triggering consequential safety routing from unbound results.

PR #444, opened 2026-09-10
feat(ringfence): fraud-verification calls for flagged bank transactions

Sarcastic-Soul:
## Summary

RingFence is an independent fraud-verification-call service: a bank/company's own fraud system flags a transaction, RingFence places (or, by default, previews) a call to the account holder's **on-file** number — never a number supplied by the flagged request itself — asks a disclosure-first set of documented scam red-flag questions, and returns a fail-closed **advisory recommendation** (`ADVISE_ALLOW` / `ADVISE_BLOCK` / `ESCALATE_TO_HUMAN`) for a human reviewer inside the institution.

**Security invariant** (the one thing the whole project depends on): the dial target is always `case.on_file_phone`; a request-supplied "callback number" is never dialed, only logged as a security event if present. Enforced in one place (`verify_call.resolve_dial_target`) that every surface calls through, and covered by its own dedicated test.

**Advisory only, human review always required.** Nothing this project returns is an action on money. The recommendation comes from a heuristic chain — speech recognition, a language model's reading of what was said, then a documented red-flag table — applied to a single phone conversation, which is not a basis for releasing or holding someone's funds automatically. So every `Disposition` carries `advisory: true` and `requires_human_review: true`, serialized into every API response, stored record, audit report, and both demo pages; `ADVISE_ALLOW` means "this call surfaced no red flags", never "release the funds". The README states plainly that integrators must not wire these values into an automated approve/hold action. Covered by `tests/test_decide.py`, `tests/test_webhook.py` and `tests/test_demo_server.py`.

**Surfaces**: a CLI (`case submit` dry-run by default, `--live --confirm-live` gated), a webhook (`POST /cases` / `GET /cases/{id}`, durable crash-safe case ledger), an MCP server, and a two-sided demo (`ringfence/demo_server.py`) — a customer "send money" page and a bank fraud-ops dashboard, where the bank decides automatically, on an amount threshold, whether a transaction needs verification at all; the customer never chooses.

**The demo server — the only deployable surface — is simulation-only and has no live-call path at all.** It never imports a CALL-E client, never reads `CALLE_API_KEY`, contains no code that could place an outbound call (asserted structurally by `test_this_server_has_no_live_call_path_at_all`, not just behaviourally), and collects no phone number of any kind. It also refuses to start on any non-loopback bind unless `RINGFENCE_DEMO_TOKEN` is set, and with it every page and API route requires that token (`Authorization: Bearer`, `?token=` once per browser, or the `HttpOnly` cookie set from it); the sole unauthenticated route is `GET /healthz`, which returns `{"status": "ok"}` so a platform health check can run. What is real in the demo is the decision: recorded fixtures run through the actual `resolve.classify()` and `decide()`.

**Measured evaluation**: 21 curated fixtures (15 of them categorized and scored) plus a 98-case exhaustive synthetic corpus — every combination of the 5 verification-call signals, every documented CALL-E ambiguity marker — checked against an independently written oracle, not `decide()` grading itself. Naively trusting CALL-E's raw `status`/`failure_code` (what most integrations do, per CALL-E's own documented warning) disagrees with the cross-referenced classification in 28/98 cases, 2 of which are unsafe. 113 cases scored in total, reproducible with `python -m ringfence.evaluation`.

**Validation against the live API**: the live path was exercised end to end against the real CALL-E API before submission, on numbers the author controls and with the recipient's consent. **No call outputs, transcripts, identifiers, or timings from those runs are included in this repository** — a verification call's contents do not belong in a public repo. What remains here is what they produced: four real bugs found by exercising the live API rather than by code review, each fixed with a regression test — a connection failure being read as a confirmed no-answer, `region`/`locale` never being sent at all, `audit.py` keying off guessed transcript speaker labels (which would have produced an empty audit trail for every real call), and a vague caller identity that CALL-E's own content-policy layer refuses outright. Two of the four were invisible to the fixture corpus, because the fixtures encoded the same wrong assumption the code did. Details in the README's "Validation against the live API" section.

Full details, safety rigor, and the security invariant are in `apps/python/ringfence/README.md`.

## Changes since first review

All four must-fix items are addressed:

1. **Real-call material removed from the tree, this description, and the branch history.** The `artifacts/` directory of real-call results is gone, the one fixture modeled on a real call is replaced by a synthetic fixture of the same provider-failure shape (`fixtures/21_connection_failed_during_attempt.json`, so the regression coverage survives), and every call identifier, timestamp, transcript excerpt, and artifact link is gone from the README, source comments, and tests. The branch's two commits were squashed into a single commit whose tree never contained that material, so it is absent from history, not merely deleted in a later commit. (This required a force-push; earlier review comments anchored to removed lines may now show as outdated.)
2. **The deployed demo service is authenticated, and its live path is hard-disabled** — both, rather than either. See the demo paragraph above: no live-call code remains in that server, and a public bind without a token refuses to start instead of serving openly.
3. **Results are advisory with human review**, per the "Advisory only" paragraph above: `ALLOW`/`BLOCK` no longer exist as values, and nothing in the project produces a consequential financial action from a call interpretation.
4. **The nonconforming commit title is gone.** The branch is now a single commit, `feat(ringfence): add fraud-verification calls for flagged transactions`, matching `docs/git-naming-conventions.md`.

## Type

- [x] New runnable app

## Checklist

- [ ] Recurring workflows include cancellation behavior. (N/A — no recurring workflow; single-attempt per case by design, no scheduling.)

### DIFF
--- a/apps/python/ringfence/.env.example
+++ b/apps/python/ringfence/.env.example
@@ -1,0 +1,11 @@
+# Copy to .env and fill in, or export directly. Never commit a real key.
+
+# Only needed for the operator-run live-call surfaces (CLI --live
+# --confirm-live, webhook, MCP). The demo server never reads it.
+CALLE_API_KEY=
+
+# Required to expose ringfence.demo_server on anything but loopback: the
+# server refuses to start on a public bind without it, and gates every page
+# and API route behind it. Generate with:
+#   python3 -c 'import secrets; print(secrets.token_urlsafe(32))'
+RINGFENCE_DEMO_TOKEN=
--- a/apps/python/ringfence/.gitignore
+++ b/apps/python/ringfence/.gitignore
@@ -1,0 +1,7 @@
+.venv/
+__pycache__/
+*.pyc
+.pytest_cache/
+results/
+.env
+*.egg-info/
--- a/apps/python/ringfence/README.md
+++ b/apps/python/ringfence/README.md
@@ -442,7 +442,7 @@ python -m ringfence.demo_server --host 0.0.0.0 --port "$PORT"
 |---|---|---|
 | `ringfence case submit --file case.json` | Dry run. Renders the disclosure-first call script, resolves the dial target, validates the case. **No call placed, no credits spent.** | `--live --confirm-live` together — **both required, no override.** `CALLE_API_KEY` must also be set. |
 | `ringfence case resolve --from-fixtures DIR` | Reads local JSON only. No network call, no API key needed. | N/A — read-only. |
-| `POST /cases` (webhook) | Dry run, unless the server itself was started with `--live --confirm-live`. | Start `ringfence.webhook` with `--live --confirm-live`. |
+| `POST /cases` (webhook) | Dry run, unless the server itself was started with `--live --confirm-live`. | Start `ringfence.webhook` with `--live --confirm-live`; live mode is loopback-only. Do not expose it through a public proxy. |
 | `ringfence_submit_case` (MCP) | Dry run. | Pass `live=True` **and** `confirm_live=True` together in the same tool call — omitting either keeps it a preview. |
 | `GET /cases/{id}` / `ringfence_get_case` | Read-only lookup of a stored recommendation. | N/A. |
 | `ringfence case audit` / `case report-index` | Reads local JSON only, writes a report file. No network call. | N/A — read-only. |
--- a/apps/python/ringfence/README.md
+++ b/apps/python/ringfence/README.md
@@ -1,0 +1,687 @@
+# ringfence
+
+**Independent fraud-verification calls for flagged transactions — never a
+number the flagged request itself supplied.**
+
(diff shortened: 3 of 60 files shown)


Ray-56:
Must Fix before merge:\n- Remove all real-call outputs, transcripts, identifiers, timings, and the real-call fixture from the tree, PR description, and affected history.\n- Authenticate the remotely reachable deployed service or hard-disable its live path.\n- Do not automatically produce consequential financial ALLOW or BLOCK actions from heuristic call interpretation; make results advisory with human review.\n- Bring the nonconforming commit title into the documented naming convention.

Ray-56:
Superseding review under the community policy adopted on 2026-09-11: Re-reviewed under the bounded community policy. Obsolete production-hardening and blanket synthetic-history blockers are withdrawn; the current basic safety list is closed. Maintainer cleanup: Live webhook now rejects non-loopback before server construction; public simulation remains no-call, ADVISE_* outputs remain human-review-only.

PR #454, opened 2026-09-11
feat(ground-truth-gate): gate a physical-world claim behind one disclosed call

ishal1410:
## Summary

Adds `skills/ground-truth-gate`, for the moment an agent is about to state a physical-world fact its sources cannot support and has not yet decided whether a call is warranted at all.

It triages the claim from three inputs — scope mismatch, evidence age, and the cost of being wrong — answers provisionally, and releases the fact only on positive evidence from the party whose word is binding. **Absence of evidence is never evidence:** an inconclusive call leaves the provisional answer standing rather than overwriting it. Five of the six triage rows place no call at all.

It belongs here because the repository's existing skills all begin after the decision to verify has been made. This one owns the step before that, and hands off rather than duplicating:

- **Calibrated abstention stays in `verify-by-phone`.** Its `gate.py` already implements split conformal abstention over yes/no/unknown. `release()` takes that abstain decision as an input and fails closed when it is missing. Two copies of a rule are two rules.
- **Durable webhook delivery stays in `apps/python/webhook-result-receiver`.** That app already does at-least-once delivery, event-ID dedup, conflict on a repeated ID with a different payload, and an authenticated re-fetch before acceptance. This skill ships no receiver; `gate.py --reconcile` consumes the object that app's re-fetch returns.
- **The nearest neighbour is named explicitly.** `references/composition.md` points at `apps/web/local-atlas/docs/calls-not-placed.md`, whose opening line is this skill's thesis, written first and already merged. The difference is one step later in the pipeline: local-atlas decides *whether to dial*; this decides *whether what came back may be released as fact*. `local-atlas` has no inconclusive state; here `unknown` is first-class.

Disclosure follows 47 CFR 64.1200(b)(1)-(2). `CALLE_CALLER_IDENTITY` and `CALLE_CALLER_CALLBACK` name the responsible entity and a callback number, and both fail closed. The identity is spoken inside a quoted script, so a value carrying a straight or curly quote is refused rather than escaped — an apostrophe would close that script early and the remainder would reach the calling agent as a fresh instruction. The callback is validated as E.164 and never repaired by guessing.

`scripts/mock_calle.py` serves all six call outcomes locally, so a reviewer can run the entire loop in about a minute with no CALL-E credential and no phone call (the mock accepts any key):

```
cd skills/ground-truth-gate
python3 scripts/mock_calle.py --outcome unknown &
CALLE_API_KEY=test python3 scripts/place_call.py --send --poll   --input assets/sample-claim.json --base-url http://127.0.0.1:8787
python3 scripts/self_test.py
```

## Type

- [x] New skill

## Checklist

- [x] Branch name, commit messages, and PR title follow `docs/git-naming-conventions.md`. (`scripts/check_branch_name.py` passes.)
- [x] Real-world side effects are clearly described. Dry run is the default; `gate.py` opens no socket at all. Only `place_call.py --send` places a call, and only with an explicit credential.
- [x] Phone numbers are masked in documentation and test fixtures unless they are clearly fictional. Every committed number is in the reserved `555-01xx` range.
- [x] Recurring workflows include cancellation behavior. Not recurring: one claim, one call, and an unresolved claim is surfaced at T+6h rather than retried.
- [x] Runnable code has a dry-run, fake-server, or no-call path by default. Both: dry run is the default, and `mock_calle.py` is a complete local stand-in.

`scripts/self_test.py` passes under `python3` and under `python3 -O`. It raises explicitly rather than using bare `assert`, because `-O` strips asserts and a suite that cannot fail is a suite that reports success without testing anything.

https://claude.ai/code/session_018rzdv3kMB7tnpRu9WAmtiM


Ray-56:

Remaining Must Fix only:

- scripts/place_call.py accepts unrestricted provider base URL and sends Bearer credentials with urllib redirects; constrain HTTPS operator-approved origins and refuse credential-carrying redirects.
- scripts/place_call.py returns raw provider error bodies, which may echo live phone or credential values; report a sanitized status/error instead.

ishal1410:
Both Must Fix items are addressed in 4de3aa1, pushed to this branch.

**1. Unrestricted provider origin and credential-carrying redirects**

`resolve_base_url()` now requires https for any non-loopback host and checks the host against an allowlist: the default provider origin plus loopback. `http://` stays permitted for `127.0.0.1`, `::1` and `localhost` only, so the `mock_calle.py` rehearsal path still runs with no credential. Operators can add their own origins through `CALLE_ALLOWED_ORIGINS`.

Requests now go through an opener whose redirect handler refuses any redirect while the request carries an `Authorization` header, so a 3xx cannot move the Bearer token to another origin.

While fixing that I also found the DNS-rebinding check was not reaching the path that actually sends: `require_public_https()` supported a resolver, but `build_request()` did not take one. It does now, threaded from `place_call.py`.

**2. Raw provider error bodies**

Error bodies now go through `_sanitize_error_body()` rather than the old `[:400]` truncation, which was not content filtering: a provider that echoes the request back on error put the key and the recipient's number on the stderr line, just below the cutoff.

It does three things, in order: exact-replaces the operator's own API key with `[REDACTED]`, replaces `Bearer <token>` with `Bearer [REDACTED]`, then runs `gate.safe_print` and `gate.mask_text` so control characters are stripped and phone-shaped digit runs are masked.

Being precise about the limit, since "sanitized" can mean more than it does here: this redacts the operator's live credential and anything Bearer-prefixed. A different secret echoed by the provider under some other key name would not be caught by pattern alone. If you would rather this never print a provider body at all and report only the status code, say so and I will make that the default with the body behind an explicit flag.

**Also in the same commit**

Two documentation claims were false and are corrected. `SKILL.md` said "Nothing in this directory places that call", which is gone. `references/safety.md` said "This skill opens no sockets", which was true of `gate.py` and not of the directory; it now reads "`gate.py` opens no sockets ... `scripts/place_call.py` is what actually sends the request and should pass one". `place_call.py --send` does place the call. SKILL.md now documents `place_call.py`, `mock_calle.py` and `demo_card.py`, the mock rehearsal loop, `CALLE_API_KEY` / `CALLE_BASE_URL`, and a troubleshooting section.

Three correctness bugs found while reviewing the above:

- `require_api_key()` was evaluated as an argument to `create_call()`, so a missing key was reported as "the call MAY ALREADY HAVE BEEN PLACED" even though no socket had been opened. `TransportError` now carries `reached_server`, so DNS and connect failures say plainly that no call was placed, while timeouts and refused redirects keep the ambiguous warning.
- `--input` and `--call-id` are now mutually exclusive. `--call-id` previously won silently and discarded `--input`.
- `self_test.py` sets `__test__ = False`, and its `check()` collects failures for `main()` to report rather than asserting, because pytest collecting the module directly reported PASSED for checks that had failed.

**Verification**

- `python scripts/self_test.py` passes
- `python scripts/validate_repository.py` passes
- loopback rehearsal completes end to end against `mock_calle.py`
- `py_compile` clean on all six scripts
- mypy reports `Success: no issues found` on the four shipped modules (`gate.py`, `place_call.py`, `mock_calle.py`, `demo_card.py`). `self_test.py` reports 8 errors, all of them tests that pass a deliberately wrong type to prove it is rejected (`cost_if_wrong="maybe"`, `claim_from_dict(["nope"])`, a heterogeneous tuple of falsy abstention values). I left them rather than annotate around a check that is doing its job, but I will silence them if you would rather the file were clean.

Re-checked just now against the pushed commit, rather than quoting the commit message:

| case | result |
| --- | --- |
| `http://evil.example.com` | refused, scheme `http` |
| `https://evil.example.com` | refused, host not on allowlist |
| `http://127.0.0.1:8787`, `http://localhost:8787` | allowed, mock path |
| `https://api.heycall-e.com` | allowed |
| host added via `CALLE_ALLOWED_ORIGINS` | allowed; a host outside it still refused |
| 302 to another origin while `Authorization` is set | refused, `_CredentialRedirectRefused` |
| provider error echoing the live key and a number | `"sent":"[REDACTED]"`, `"to":"+14*******23"` |

One thing I cannot do from my side: no check run has appeared on the head commit, so the workflow looks like it is waiting on first-time-contributor approval. Happy to rebase or split the commit if that helps review.

Ray-56:
Superseding review for head ee7401ad: the frozen blockers are closed. I applied one bounded maintainer cleanup: Omit the redirect Location from errors and sanitize the complete provider error before truncation. Focused credential-free checks and clean-source repository validation passed; no real calls were placed. No current Must Fix remains.

PR #461, opened 2026-09-11
feat(apps): add call-e code commander for on-call escalation with spoken authorisation

ajay-7415:
## Summary

Adds **CALL-E Code Commander**, an on-call escalation commander for production incidents, at `apps/typescript/call-e-code-commander`.

When a service breaks, an automated council investigates, stages a hotfix, and proves it green in a sandbox. Then it stops. Nothing reaches production until CALL-E has phoned a human, explained the incident in conversation, answered their questions, and extracted an explicit spoken decision as typed JSON.

If the primary on-call does not answer, the ladder climbs to the backup and then the manager, telling each person why they are being woken instead of the previous rung. If the whole rota is exhausted, the deploy is blocked.

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

## CALL-E platform usage

| Capability | Use |
| --- | --- |
| `calls.create` | Places each rung of the escalation ladder |
| `recipient_result_schema` | Extracts the engineer's decision as typed JSON |
| `Idempotency-Key` | Derived from incident and contact IDs, so a retry never double-dials |
| `metadata` | Carries incident ID, service, severity, and rung |
| `calls.get` | Drives live rung state through the attempt lifecycle |
| `calls.listEvents` | Streams developer events into the incident timeline |
| `webhook_url` | Terminal call state delivered to a receiver, so an escalation survives a closed tab |
| `call.result_validation_failed` | An extraction that failed schema validation fails closed like any other unusable call |
| `callback_minutes` | An engineer who asks for ten minutes gets a callback, not an escalation past them |
| `completion_confidence` | Below 0.5, the decision escalates instead of shipping code |
| `additionalProperties: false` | Strict extraction, so an off-schema result returns null and fails closed |
| `transcript_turns` | Mirrored into the live transcript and the post-mortem |
| `evidence` | Attached to the post-incident review |
| `region` / `locale` | Resolved from the published regions table before dialling |
| Stable error codes | Branching on `unsupported_region`, `insufficient_balance`, `account_concurrency_exceeded`, `idempotency_conflict` and the rest. A concurrency limit and a rate limit both arrive as HTTP 429 and need different responses, which is why the code is what gets read |
| Server-side proxy | Key read from the server environment, never sent to the browser |

The decision schema follows CALL-E's own guidance to prefer string enums with an explicit `unknown` value over booleans for judgements that may be unclear. `answered_by` distinguishes `named_engineer`, `different_person`, `voicemail_or_ivr`, `no_answer`, and `unknown`, and only the first can authorise a deploy. Voicemail greetings and a colleague picking up the on-call phone both produce text that reads like approval, so this is the field that has to be precise.

## Safety and side effects

This app places real outbound phone calls when a server key is configured. With no key it runs a clearly labelled simulation, which is the default.

- Nothing deploys without the named engineer on the line and a sufficiently confident extraction.
- An unanswered rota fails closed. Silence is never read as consent.
- Coverage is resolved before a credit is committed. A valid E.164 number does not establish that its destination is supported, so an uncovered rung is refused in the dialog rather than failing as `unsupported_region` after acceptance.
- `completion_confidence` corroborates but never authorises. The platform documents it as confidence in CALL-E's task-completion judgment, not in the business answer, so it can withhold a deploy and never grant one.
- `failure_code` is preserved for support but never interpreted. It has no published enum, and the platform warns against inferring a decline or a no-answer from it. A failed call can still carry a schema-valid structured result, which is separate evidence and is what the audit trail records, but it can never authorise anything.
- A human "hold for review" stops the ladder rather than waking the next person.
- A ringing phone is never displayed as a person on the line. `in_progress` covers dialling, so the connected state waits for a transcript turn spoken by the recipient.
- The call task prompt forbids asking for passwords, tokens, or one-time codes.
- There is no API key field in the app. The key is read from the server environment only, per CALL-E's server-only guidance, and never reaches the browser.
- The on-call rota is server configuration too. Real names and mobile numbers are read from the environment and never written into source, so nothing in this contribution identifies anyone. Every number in the code, tests, and documentation is fictional.
- The custom code path runs a real static scan and reports findings only. It never generates a patch, because a confident wrong diff during an incident is worse than no diff.
- A webhook delivery is never believed. The SDK marks its signature helpers deprecated and states why: CALL-E deliveries are not signed. A delivery body contains a complete structured result saying who authorised what, so reading a decision out of it would let anyone who guessed the receiver URL ship code by sending JSON. Only the call id is taken from a delivery. Every recorded fact is re-read from the API with the server key and interpreted by the same `interpretDecision` the ladder uses.
- The hosted deployment cannot be used to dial a stranger. The proxy forwards three call routes and nothing else on the account, requires an incident correlation id, rate limits placement separately from polling, and rings only numbers the owner nominated. An unconfigured deployment refuses every call rather than allowing every call.
- An engineer who asks for time is not escalated past. The ladder waits, calls them back, and tells them it is the call they asked for. One deferral each, capped at thirty minutes, and the audit trail records both calls so the gap is explained rather than hidden.

## Verification

- 50 tests, no network or credits needed: `npm test`. The first 35 cover phone validation, region coverage, key sanitisation, idempotency, decision interpretation, escalation stopping rules, the static scanner, rota configuration, error-code branching, and a check that no live API key is committed. The remaining 15 cover the proxy route and recipient allowlists, rate limiting, the callback rules, the webhook trust model including a forged delivery that claims a deploy authorisation, duplicate delivery handling, and the audit export.
- `scripts/verifyLiveCall.js` places one real call and prints the structured decision end to end. It is opt-in and requires a key.

MIT licensed.


Ray-56:
Must Fix before merge:

- Authenticate hosted live-call and private incident/transcript/rota endpoints; allowlisting and rate limits alone do not authorize callers.
- Stop escalation after ambiguous submission, unreadable polling or nonterminal timeout; do not automatically call the next contact.
- Mask full phones and provider-derived phone content in UI, logs and exports.
- Make verifyLiveCall.js no-call by default with explicit per-run live intent and recipient permission.
- Describe deploy/rollback effects as simulated, and clarify that stopping the ladder does not cancel an accepted call.

PR #474, opened 2026-09-11
feat(apps): add routeready call-ahead route re-sequencing

Mumtio:
## Summary

Adds **RouteReady** (`apps/typescript/routeready`) and the **`route-readiness-call`** skill.

RouteReady is the delivery route that calls ahead. While a rider is still riding, it uses CALL-E to phone the next customers on the route, accepts an answer only when the customer's own words back it, re-orders the remaining stops around what they said, and shows the rider one instruction. Existing delivery entries such as `dispatch-pulse` and `cod-confirm` confirm an order before dispatch; RouteReady works during the route, and the call result changes the order of the stops that are left.

- **One line, one call at a time.** The next call goes to the soonest uncalled stop the rider will reach in 6 to 45 minutes. Each customer is called at most once a day and never redialled.
- **CALL-E at runtime.** `client.calls.create` with a strict `recipientResultSchema` and an idempotency key per run and stop; `calls.get` and `calls.listEvents` stream the conversation into the rider app while it happens.
- **Evidence gate.** An answer changes the route only if the call completed, reached the planned number, reached the customer, quotes their words, states a time, and CALL-E's confidence is medium or high. `task_completed` alone is never trusted.
- **Re-ordering.** Branch-and-bound over every order of up to nine stops, with each customer's ready time as the earliest delivery; tested against exhaustive search on random days.
- **Rider app.** The courier's phone app: a route map with the next-stop card and a live call strip, the stop list, the calls with their transcripts, and a Today screen comparing the day with and without calls. A laptop shows two phones side by side.
- **Made to be watched.** Scripted calls play line by line over about seventeen seconds, each answer holds before the next call starts, route changes appear as a message on the rider's screen, and the day can be paused or run slow, normal or fast.
- **Your own route, real calls.** At `/route` a visitor brings their own CALL-E key, pins real stops on a map, and moves the rider by dragging or by GPS; each customer gets a real CALL-E call before the rider arrives.
- **Skill.** Task text, result schema (kept identical to the app's by a test), evidence gate, answer-to-route rules, safety notes and examples.

On the simulated Dhaka day, the same customers with and without calls: failed attempts 2 → 0, wasted trips avoided 0 → 2, route finished 16 minutes earlier, for 7 calls ($0.35 at $0.05 each).

## Type

- [x] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Your route with real calls

Try it: https://awesome-phone-call-agents-xloz.onrender.com/route (free instance, the first visit can take a minute to wake up).

1. Paste your CALL-E API key. It is checked with a read-only request, lives only inside that route's CALL-E client in server memory, never reaches the browser or logs, and is dropped when the route ends.
2. Pin up to five stops on the map, each with a customer name and a phone number you are allowed to call, and place the rider.
3. Pick the call-ahead time and how the rider moves: **Drag on the map** to test from a desk, or **My phone's GPS** while riding.
4. When the next customer is within the call-ahead time, CALL-E calls them. The transcript streams in, the answer goes through the evidence gate, and a verified answer re-orders the remaining stops. Press **Delivered** or **Nobody home** at each door.

Each visitor gets their own route session; idle routes end after 20 minutes and every route after three hours. The re-ordering on this screen also weighs how long each customer waits, so a customer who needs half an hour never sends the rider past a ready customer next door. The whole flow was checked end to end in a browser against a local fake of the CALL-E API (`CALLE_BASE_URL`).

## Try it without an account

```bash
cd apps/typescript/routeready
npm install
npm test
npm run sim
npm start   # rider app at /app, two phones at /, your own route at /route
```

## Live calls

Opt-in only: `ROUTEREADY_LIVE=1`, a server-side `CALLE_API_KEY`, `LIVE_TARGETS` mapping stops to authorised numbers, the start token printed at startup, and a consent statement in the app. Stops without a live target keep scripted customers.

Author-reported live checks, all against CALL-E's official US test hotline, which answers with an AI receptionist:

- `npm run smoke:live`: accepted immediately, ringing after about 60 seconds, final result after 137 seconds. CALL-E returned `reached_recipient: "no"` together with `task_completed: true` and high confidence, which is why the gate does not trust `task_completed`.
- A live day in the app with one stop mapped to the hotline: the transcript streamed into the live call card while the call happened (4 to 23 lines over about two minutes, day clock in real time), the answer was classified unverified (customer not reached), and the route stayed unchanged.

As of 12 September 2026 CALL-E rejects Bangladesh destinations in both English and Bengali (`422 call_not_ready`, see CALLE-AI/call-e-integrations#98), so the demo day is set in Dhaka while live targets are US numbers.

## Checklist

- [x] Recurring workflows include cancellation behavior. (No recurring jobs; the README documents that CALL-E has no cancel endpoint and that Stop prevents new calls.)

## Validation

- `npm run check` (TypeScript)
- `npm test`: 69 tests (including `tests/safety.test.ts` for the review fixes)
- `npm run sim`
- `python3 scripts/validate_repository.py`: passed


Ray-56:
Must Fix before merge:

- Stop the live queue after ambiguous creation or nonterminal timeout, rather than freeing the line and calling the next stop.
- Restrict visitor/server provider credentials to approved HTTPS origins; arbitrary fake origins must be dummy-only.
- Authenticate shared live-day transcript reads and controls, or keep that surface loopback-only.
- Mask provider-derived phones in streamed transcripts, errors, logs and smoke-script output artifacts.
- Enforce the documented once-per-customer/day rule for repeated destinations, or explicitly document and require approval for distinct repeat-call intents.

Mumtio:
Thanks for the review. All five are addressed in dc0e293: the queue now fails closed on ambiguous creation and on calls still non-terminal after 15 minutes; real keys only go to approved HTTPS CALL-E origins (other CALLE_BASE_URLs accept dummy- keys only); live calls on the shared demo day require a loopback HOST, with cross-site and non-loopback-Host requests refused; phone numbers are masked in transcripts, provider errors, logs, error responses and smoke artifacts; and once-per-day is enforced by destination number, with duplicates rejected and same-day repeats requiring explicit approval. Covered by tests/safety.test.ts (69 tests total).

Ray-56:
Superseding the earlier review: the frozen acceptance list is now closed. The author fixed the call-ambiguity, credential-origin, destination and access boundaries; the bounded maintainer patch masks the two remaining display snapshots without changing private routing inputs. Focused offline checks and repository validation pass. Advisory routing and in-memory recovery remain documented demo limitations, not merge blockers.

PR #481, opened 2026-09-11
feat(apps): add procurepulse supplier quote race app

HectorTa1989:
## Summary

ProcurePulse turns an urgent stockout into a supplier quote race over CALL-E: it calls each
authorized supplier with the identical outcome-focused questions, validates every per-supplier
structured result against a closed schema, and ranks only quotes that can honestly be compared.
A vague answer becomes **needs review** with the specific reasons rather than a guessed number.
Approving a supplier is an internal record; asking that supplier to hold stock takes a second
explicit approval and a second call that says out loud it is not a purchase.

`apps/typescript/procurepulse/` is a dependency-free TypeScript library and CLI (Node 20+,
`tsx`/`typescript` as dev dependencies only).

Patterns this contributes to the repository:

- One `recipient_result_schema` per race: closed, every field a described string, business
  decisions as enums with `unknown`, and no CALL-E reserved recipient field names.
- Number-like answers kept as strings until a deterministic step decides whether they justify
  arithmetic: a comparable total only when quantity, unit, availability, currency, price and
  fees are all unambiguous. Hedges are matched as whole words, so "verified" is not "if".
- Completeness (how much the supplier answered) kept separate from CALL-E's own certainty.
- A task that discloses the agent is AI and the business it represents in its first sentence,
  forbids orders, holds, terms and payment details, and never contains the phone number.
- A stable `Idempotency-Key` per request, supplier and purpose; re-running dispatch skips
  suppliers that already have a call id, so a double click or retry never places a second call.
- Live lifecycle read from the snapshot itself (queued, ringing, on the call, finalizing,
  completed) with read-only polling as the default result path, so a timeout cannot redial.
- An unsigned-webhook receiver: path token, `CALL-E-Event-Id` matched against the body `id`,
  the call taken from `data.id`, the event claimed before any side effect, and an independent
  `GET /v1/calls/{id}` re-fetch that is the only thing ever persisted.
- Two human gates: internal approval of a validated quote, then a separately confirmed
  non-purchase hold call with its own result schema.

Default paths place no calls and need no API key: `preview` prints the preflight and the exact
`POST /v1/calls` body with masked recipients, `demo` runs the whole race (three quotes, ranking,
needs-review, approval, hold) against a loopback fake CALL-E, and `replay` ranks a saved
terminal snapshot. Live commands require `--execute --i-have-consent` plus an allowlisted E.164
destination. 27 offline tests; sample numbers use the reserved fictional 555-01xx range.

The payload shape was checked against the live CALL-E API by sending the real quote and hold
bodies with a deliberately invalid recipient, so nothing was dialed: CALL-E validates request
shape (422 on unknown fields), then `recipient_result_schema` (400 on reserved names), then the
phone number (400 `invalid_phone`).

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist


Ray-56:
Superseding the earlier review: the bounded maintainer fixes have closed the remaining Must Fix items under the 2026-09-11 community policy. Focused credential-free checks passed; no real call was placed.

PR #483, opened 2026-09-11
feat(skills): add no-show-shield day-batch confirmation skill

landbuild:
## What this adds

`skills/no-show-shield/` — a skill that calls **every booking on a given day**, captures each outcome as structured JSON, writes the results back to a calendar file, and returns a short list of only the slots that still need a human.

## Why it isn't a duplicate of `appointment-confirm`

They differ in scope, and the SKILL.md says so explicitly so a reader can pick correctly:

| | `appointment-confirm` | `no-show-shield` |
|---|---|---|
| Unit of work | one appointment | one day's book |
| Calendar | deliberately untouched | outcomes written back |
| Intended use | human reviews each result | unattended evening run, short morning list |

The use case is the solo operator — trades, mobile clinic, single-chair salon — who loses a whole slot to a no-show and has no hour in the evening to ring ten people.

## Safety

`references/safety.md` covers every item in CONTRIBUTING's safety list, and describes mechanisms rather than intentions:

- **Intent** — calls go only to customers already booked in the operator's own calendar file; no discovery or list building; runs are operator-initiated for one named date; `preview` places no calls and is the documented first step.
- **E.164** — validated before dialling; malformed numbers are skipped and reported, never guessed at.
- **Masking** — summaries print names, times and outcomes, never customer numbers.
- **Credentials** — `CALLE_API_KEY` from the environment only; never written to calendar, ledger, or summary.
- **No hidden recurrence** — no scheduler, daemon, or retry loop exists in the skill; a re-run only picks up what is still pending.
- **No duplicate jobs** — outcomes are written back immediately, so a booking is called at most once per run.
- **Call budget** — a file-backed cap with no override flag; exceeding it raises and stops the run. Raising it is a tracked edit.
- **Content boundaries** — service name and time only; no medical, legal, or financial discussion; cancellations accepted on first statement, never contested.

Demo data uses only ACMA-reserved fictional Australian numbers (`+61 491 570 xxx`). A live demonstration must pass `--override-phone`, which routes every call in the run to one number the operator types at the command line, so a demo cannot dial a third party by accident.

## Validation

```
python3 scripts/check_branch_name.py --branch feat/no-show-shield   # passes
python3 scripts/validate_repository.py                              # Repository validation passed.
```

## Notes from live use

Two things learned running this against real calls, both written into `references/examples.md` for the next person:

- The first call of a session can take several minutes between `create` and the phone ringing; later calls connect faster. Worth budgeting wall-clock time when demonstrating.
- Times need rendering into natural speech ("tomorrow at 9:30 in the morning") before they reach the model — passing ISO dates through makes the agent read them out digit by digit.

Reference implementation with tests and an honest limitations section: https://github.com/landbuild/no-show-shield

Submitted for the CALL-E "Your Code Is Calling" hackathon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,7 @@ This project is an awesome list for AI-agent phone-call workflows. Add resources
 - [`pharmacy-cash-price`](skills/pharmacy-cash-price/) - Asks one retail pharmacy for a cash price with no insurance and returns a structured quote, a refusal, or an explicit unknown, with the disclosure and the medical boundary written into the call.
 - [`logistics-exception`](skills/logistics-exception/) - Resolves a missed dock window by calling the driver and receiving dock concurrently with one strict CALL-E result schema, reconciling terminal results by event id and re-fetch, combining only reached-contact facts into a recovery card, and gating any dock-changing follow-up call behind explicit human approval.
 - [`otherend-task-test`](skills/otherend-task-test/) - Rehearses a CALL-E task text and result schema against a programmable receptionist line the operator owns, reads the deterministic grade (manifest, self-report, fabrication, disclosure, confidence calibration), and turns each failing check into a task-text edit before the task reaches real people.
+- [`no-show-shield`](skills/no-show-shield/) - Calls every booking on a given day to confirm it, writes yes/reschedule/cancel outcomes back to the operator's calendar, and returns a short list of only the slots that still need a human.
 
 ### Apps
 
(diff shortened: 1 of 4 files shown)


Ray-56:
The frozen Must Fix items are resolved by a bounded maintainer cleanup. Qualified supervised experimental/mock scope and interruption/ambiguous-outcome limitations without redesigning the external app. This supersedes the corresponding blockers; no production hardening is required.

PR #484, opened 2026-09-12
feat(apps): complete async live Reality Resolver execution

Angelo-The-S:
## Summary

- add asynchronous queued/running/completed/failed Reality Resolver execution after #389
- preserve the fake backend demo while enabling the guarded live CALL-E path
- enforce strict destination authorization and live capacity limits
- handle provider failures as technical failures with provider-safe diagnostics
- add offline and mocked-live coverage

## Validation

- Reality Resolver test suite — **409 passed**
- `python scripts/validate_repository.py` — **passed**
- `git diff --check` — **passed**

The repository-wide pytest command cannot currently complete collection because the external `calle-ai` dependency is unavailable in the local environment. The complete Reality Resolver suite passes independently.

No real CALL-E calls are made by the test suite.

## Notes

This PR contains only the incremental Reality Resolver work completed after merged PR #389.

The live HTTP path uses the real CALL-E pipeline, while the deterministic fake provider remains available for safe and reproducible demos.

### DIFF
--- a/apps/python/reality-resolver/api/custom_case.py
+++ b/apps/python/reality-resolver/api/custom_case.py
@@ -1,0 +1,171 @@
+"""Validation and normalization for one client-supplied Case.
+
+Custom cases stay in memory for one resolution. They are never written to
+``cases/`` and therefore cannot widen the shipped-case allowlist or expose the
+operator's untracked live fixture. The returned ``Case`` is the same model the
+deterministic engine already consumes.
+"""
+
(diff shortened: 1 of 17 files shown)


Ray-56:
The frozen Must Fix items are resolved by a bounded maintainer cleanup. Reused existing phone/obvious-secret masking for custom-case projections before truncation; synthetic regression passed. This supersedes the corresponding blockers; no production hardening is required.

PR #486, opened 2026-09-12
feat(apps): add dial phone-work agent app

FaresCH10:
## What this adds

**Dial**, under `apps/typescript/dial/` — a user-facing app that turns a plain-English request ("find the cheapest brake pad replacement near me this week, with a warranty") into goal-driven phone work over CALL-E: it finds local businesses, calls until it has enough *comparable* answers rather than a fixed count, compares only what was actually said on the phone, explains why a cheaper option lost, and books only through an explicit authorization step.

It also includes a business mode — the same engine pointed outward — for a business calling its own customers: spreadsheet import with a dry-run preview, three workflow templates, per-recipient structured outcomes.

Contribution area: **User-facing Apps**. Root `README.md` gains one bullet under *Apps* and one table row.

## Side effects, stated plainly

**This app makes real phone calls to real businesses, and they cost credit.**

- The no-call simulator (`TEST_PROVIDER=mock`) is the **default**. Every simulated call carries a visible banner in both UIs; a simulated result is never presented as real.
- `TEST_PROVIDER=real` without `CALLE_API_KEY` refuses to boot rather than running fake calls behind a UI that says LIVE.
- Every call identifies itself as an AI assistant at the start. Not configurable.
- Calling is bounded: `MAX_CALLS_PER_TASK` (default 5), extended one business at a time up to `MAX_CALLS_UNTIL_RESULT` (default 10) only while the goal is unmet, plus an atomically-reserved per-user daily ceiling.
- Every brief has explicit end conditions — ask at most twice, hang up on a website redirect, treat two identical replies as a recording — written after a real call ran fifteen exchanges against an answering machine.
- Pause / cancel stop everything not yet dialled. CALL-E has no in-flight cancellation, so Dial does not render a button claiming otherwise; it reports how many calls were already connecting.
- Business mode: calling hours enforced at dispatch, do-not-call / opt-out wins over a run already started, idempotent dispatch so a retried job cannot ring anyone twice.
- **Known limitation, stated in the README:** there is no endpoint to stop a business run once started. Listed first under *What is not done*.

## Things it refuses to do

Turn a non-answer into an answer; invent per-fact confidence (CALL-E gives one per call); convert currencies; let a model authorise a side effect; chase an impossible request.

## Running without calling anyone

```
cp .env.example .env     # mock provider already set; add a Gemini key
npm install && npm run build && npm run db:migrate && npm run seed
npm run dev:api & npm run dev:worker & npm run dev:web
```

Only a Gemini key is required. Database is embedded (PGlite), discovery falls back to OpenStreetMap.

## Verification

- 475 automated tests across seven packages, against real PGlite, the real HTTP layer and the real worker loop. Lint and typecheck clean.
- No `.env`, no build output, no secrets. All sample phone numbers use `555 01xx` local parts within real country codes.
- Real calls have been placed during development with a live key; what was learned from them is in the code and named in the README.


Ray-56:
Must Fix before merge:

- Stop and reconcile calls still nonterminal after the polling budget; do not requeue a fresh-key redial or advance conflicting calls.
- Restrict real credentials to approved secure CALL-E origins.
- Mask phone-bearing provider results, evidence and errors in API/display/log boundaries.
- Keep seeded fictional destinations fake-only; they are not inherently undiallable.
- Correct the no-bulk-calling/contact-upload claim to match the implemented import and batch workflows.

PR #494, opened 2026-09-12
Add CallFollowUp app

kusumaranikusumarani3232-hub:
## CallFollowUp

Adds CallFollowUp, a Streamlit application that uses CALL-E for AI-powered follow-up calls.

### Features

* Contact and phone number management
* Call goal configuration
* CALL-E call integration
* Call status and outcome tracking
* Notes and next-action tracking
* Follow-up data storage

The repository validation script passes successfully.

### DIFF
--- a/apps/python/callfollowup/README.md
+++ b/apps/python/callfollowup/README.md
@@ -1,0 +1,148 @@
+# CallFollowUp
+
+### AI-powered business follow-ups that turn conversations into actionable next steps.
+
+CallFollowUp is a Streamlit application that helps small businesses manage customer follow-up calls using the CALL-E phone-call API.
+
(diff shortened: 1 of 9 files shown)


Ray-56:
Must Fix before merge:

- Make credential-free preview independent of authenticated client initialization.
- Validate exact live E.164 and support only loopback/private use or basic authentication for remote credential-bearing deployment.
- Mask destination, history, transcripts, nested results and provider errors.
- Show actual live/dry-run and terminal status. Treat ambiguous creation as unknown, stop for reconciliation, and document that closing the UI does not cancel an accepted call.

Ray-56:

1. Require authentication for remotely reachable live/private records or enforce loopback binding; a private server IP is not a safe auth bypass (app/config.py:38).
2. Reject non-ASCII destinations with a full-string E.164 gate (app/call_service.py:32–43).
3. Mask free-text result/history/task previews and provider diagnostics, not only the dedicated phone field (call_service.py:139,192,221,239,275; dashboard.py:367,518–527).

PR #495, opened 2026-09-12
feat(apps): add DentalConfirm appointment confirmation app

Hary-08:
## DentalConfirm

DentalConfirm is a web application that uses CALL-E AI phone agents to automate dental appointment confirmations.

### What it does

- Manages upcoming dental appointments.
- Starts personalized appointment confirmation workflows.
- Uses CALL-E for phone-agent interactions.
- Tracks outcomes such as confirmed, reschedule, cancelled, and no_answer.
- Helps dental practices reduce no-shows and save staff time.

### Project

Source code:
https://github.com/Hary-08/dental-confirm

Live demo:
https://dental-confirm.onrender.com

### Technology

- Node.js
- Express
- HTML/CSS/JavaScript
- CALL-E SDK

### Safety

CALL-E credentials are stored in environment variables and are not exposed to the frontend.

Demo phone numbers should be fictional or masked.

Live outbound calling to Italian phone numbers is currently unsupported by CALL-E, so this project does not claim a successful live call to an Italian number.

### Testing

The project can be run locally with:

npm install
npm start

The application runs on port 3001 by default.

### DIFF
--- a/apps/web/dentalconfirm/README.md
+++ b/apps/web/dentalconfirm/README.md
@@ -1,0 +1,51 @@
+# DentalConfirm
+
+DentalConfirm is a web application that uses CALL-E AI phone agents to automate dental appointment confirmations.
+
+## Problem
+
+Dental practices spend significant staff time calling patients to confirm appointments. Missed appointments also create avoidable gaps in schedules.
+
+## Solution
+
+DentalConfirm provides a simple dashboard where staff can manage upcoming appointments and start an automated confirmation workflow using CALL-E.
+
+The workflow can handle outcomes such as:
+
+- confirmed
+- reschedule
+- cancelled
+- no_answer
+
+## Features
+
+- Dental patient appointment management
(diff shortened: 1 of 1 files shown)


Ray-56:
Must Fix for head 3403f80: the linked Hary-08/dental-confirm project retains an actual CALL-E credential in public default-branch history (public/index.html, introduced in 82fcca6 and removed from the current file in 1539819). Please revoke/rotate that credential and remediate the affected public history before merge, then confirm completion without posting the credential. Deleting it only from the latest file is insufficient. This is the frozen acceptance item for this documentation-only contribution; no external architecture redesign is required.

Ray-56:
The reachable-history cleanup is verified and closes that part of the previous review. Remaining Must Fix: confirm that the previously exposed CALL-E credential has been revoked or rotated, without posting its value. This supersedes the older combined history/rotation request; platform-purge confirmation is not required.

PR #506, opened 2026-09-12
feat: add triage-line — welfare-check calls with triage and escalation

splusb:
When disaster strikes, no one should wait by a silent phone.

When a February 2021 winter storm knocked out power across Texas, 246 people died, and about 60% of them were 60 or older ([Texas DSHS, via AARP](https://www.aarp.org/livable-communities/tool-kits-resources/info-2022/disaster-risks-to-older-adults.html)). The pattern repeats in nearly every disaster: after Hurricane Katrina, roughly half of the Louisiana victims were 75 or older ([Louisiana Dept. of Health](https://ldh.la.gov/assets/docs/katrina/deceasedreports/KatrinaDeaths_082008.pdf)), and the CDC notes that about 80% of older adults live with a chronic condition ([CDC](https://stacks.cdc.gov/view/cdc/20213/cdc_20213_DS3.txt)) that makes an outage or heat wave far more dangerous. Many die at home, isolated, because no one reached them in time.

The people most at risk are often known in advance, sitting on opt-in registries held by health departments and emergency managers. The bottleneck isn't knowing who to call. It's the hours it takes humans to actually dial them, one by one.

**TriageLine** removes that bottleneck. It's an autonomous welfare-check caller built on [CALL-E](https://www.heycall-e.com/): after a disaster or outage, it phones every affected resident at once, asks if they're safe and what they need, verifies and triages the answers by urgency, and escalates on its own — a call to a GP for anyone who needs help, and a call to an emergency contact for anyone it couldn't reach.

CALL-E makes the phone calls and returns structured results. TriageLine is the brain on top — it decides who to call, trusts what comes back only when the evidence supports it, respects consent boundaries, self-heals across language barriers, and turns a welfare check into action.

## Triage-Line — autonomous phone-task orchestrator on CALL-E

**Contribution area:** User-facing App (`apps/typescript/triage-line/`)

CALL-E makes the phone calls; Triage-Line is the orchestration brain on top. One engine drives two call topologies:

- **Fan-out** — call many residents with the same welfare-check goal in parallel, triage by urgency, and auto-escalate anyone who needs help to a GP follow-up call.
- **Chain** — each call's result reveals the next prerequisite and number to dial (clinic → GP → insurer → booked); the call graph grows itself.

### Value added on top of CALL-E
- **Verification** — gates on `completionConfidence`, checks required fields, and confirms evidence-bearing values appear in the transcript before trusting a result.
- **Consent policy** — encoded into the call task and asserted after (e.g. never agree to a fee); violations halt the chain and flag for a human.
- **Language auto-retry** — re-dials in the next allowed locale for the region on a suspected language barrier.
- **Human review** — flagged nodes can be approved with a reviewer note kept in the audit trail.

### CALL-E integration
Uses the official `@call-e/calle` TypeScript SDK (`src/calle/real-client.ts`, `client.calls.createAndWait`). Ships a **mock no-call default** so the multi-recipient orchestration can be rehearsed without consuming call quota; `CALL_MODE=live` and `CALL_MODE=hybrid` place real calls. The live path was validated end to end against a real phone call.

### Safety
- Real calls gated: loopback-only bind, explicit LIVE_CONFIRM, and per-run confirmLive intent; mock/no-call is the default.
- Every dial leg validated as E.164; no synthetic or unapproved provider-discovered numbers.
- Phone numbers masked in snapshots, events, transcripts, and logs.
- No secrets or personal data committed; .env is gitignored, only .env.example ships.

### Run
```bash
npm install
npm run server   # http://localhost:3000
# or: npm run reachback / npm run runaround
```

Repository validation passes (`python3 scripts/validate_repository.py`).


Ray-56:
Must Fix before merge:

- Protect real-call/private-result endpoints or keep live serving loopback-only, with explicit per-run live intent.
- Require authorized valid E.164 for every live leg; do not dial synthetic defaults or provider-discovered numbers without approval.
- Restrict real provider credentials to approved HTTPS origins.
- Keep consequential medical/referral expansion sandbox-only or require human approval of validated, bound results; presence-only fields are not a verified authorization.
- Mask phones in live snapshots/events/transcripts/debug output.
- Correct the locale-retry contract: changing a payload while retaining its idempotency key is not a new locale call. Document cancellation and ambiguity-stop limitations.

splusb:
Thanks for the review — all six Must-Fix items are addressed:

1. **Live endpoints protected.** The live path is gated behind loopback bind + `LIVE_CONFIRM` + per-run `confirmLive`, so a hosted/served demo can't place real calls even if someone tries. Mock/no-call is the default.
2. **Authorized E.164 only.** Every live leg is validated as E.164 before dialing; synthetic defaults removed (fail closed) and provider-discovered numbers are not dialed without approval.
3. **Credentials over HTTPS only.** A non-HTTPS `CALLE_BASE_URL` is refused.
4. **Consequential calls need human approval.** GP / emergency-contact / discovered-number legs are proposed as `awaiting_approval` and only placed after explicit authorization; presence-only fields are never treated as authorization.
5. **Phones masked.** Snapshots, events, results, transcripts, and debug output mask phone numbers; full numbers are never serialized or logged.
6. **Locale-retry contract fixed.** The idempotency key now includes locale + attempt, so a retry is a genuinely new call. Cancellation and ambiguity-stop limitations are documented in the README.

Follow-up in the latest commit: an approval UI for `awaiting_approval` nodes (an "Authorize call…" control showing the proposed reason + optional reviewer note; authorizing then places the call).

Ray-56:
Remaining Must Fix for `2cc3d849` under the community/demo policy: This supersedes the previous list: local-only access, explicit intent and locale/cancellation requirements are now closed. Only these four frozen items remain.

1. Require explicit authorized live recipients instead of built-in fixture fallbacks in reachback.ts and runaround.ts, including the final clinic leg.
2. Restrict credential-bearing CALLE_BASE_URL to the approved HTTPS origin; accepting arbitrary HTTPS origins is insufficient (src/calle/real-client.ts).
3. Gate book_clinic with human approval or keep it sandbox-only; presence-only preauthorization/result fields must not automatically trigger the consequential booking call (src/scenarios/runaround.ts).
4. Mask the existing nested result, verification-note, error and CLI output boundaries while preserving private evidence (src/server/runs.ts and src/index.ts).

splusb:
All four frozen Must-Fix items are addressed (latest commit `341e39c`):

1. **No fixture-fallback recipients.** `reachback.ts` dials only explicitly authorized `REACHBACK_PHONES` in live mode; the sample 555 roster is mock-only (never dialed for real), and hybrid drops a real-node id that has no configured number. `runaround.ts` uses `CLINIC_PHONE` for both the clinic start and the final booking leg — no fixture fallback, and unset fails closed via the E.164 guard.
2. **CALLE_BASE_URL restricted to an approved origin.** `real-client.ts` now checks `url.origin` against an exact allowlist (official API + `CALLE_ALLOWED_ORIGINS`); arbitrary HTTPS origins are refused.
3. **book_clinic gated.** `runaround-book` is `requiresApproval: true`; the presence of a preauth/result field never auto-triggers the booking call — a human must authorize it.
4. **Masking at all boundaries, evidence preserved.** `runs.ts` deep-masks nested result values, verification notes/flags, and run errors; `index.ts` masks CLI output. I also found and fixed a gap here: redaction previously matched only E.164, so a human-formatted number in a transcript (e.g. `415-555-0142`) leaked — it now masks common written formats too, while leaving non-phone evidence (reference codes, confidence scores) intact.

Verified on the mock/served path: typecheck clean, both scenarios run, and a live snapshot confirms transcript, nested-result, and proposed-reason numbers are all masked with no leaks. Note: guards live in the same code paths the live path uses, but were exercised via mock rather than a real call.

Ray-56:
Superseding the previous review: the four frozen Must Fix items are closed at this head (explicit live recipients, approved HTTPS credential origins, human approval for booking, and masked outputs). Credential-free mock checks passed; no real calls were made. Remaining demo limitations do not block merging.

PR #508, opened 2026-09-12
feat: add generic Voice Scout business qualification app

barsec-admin:
# Voice Scout PR description

## Summary

Adds a reusable Python app for business lead qualification through CALL-E Goal Runs.

## Contribution area

User-facing App: `apps/python/voice-scout/`

## Included

- Safe preview mode by default
- Explicit live-call boundary
- Generic business lead schema
- Stable idempotency keys
- Published Goal Run submission and polling
- Structured result handoff
- Synthetic example lead
- Generic Goal template
- Setup, safety, cancellation, and credential-handling notes

## Scope

This contribution is intentionally CRM-agnostic and business-domain agnostic. Cybersecurity is only one possible use case. No production CRM data, credentials, or real phone numbers are included.

## Validation

```bash
python3 app.py --demo
python3 -m py_compile app.py
```


Ray-56:
Must Fix before merge:

- Require an explicit authorized live lead, validate ASCII E.164, and prevent the bundled synthetic lead from being called by --live alone.
- Mask nested provider results/transcripts and coarse CLI errors, not only lead.phone.

barsec-admin:
Addressed in commits `98c2292` and `8331475`.\n\n- `--live` now requires an explicit `--lead` file with `authorized_live_call=true`; the bundled synthetic lead cannot be called by `--live` alone.\n- Lead phone validation is ASCII E.164.\n- Provider results are recursively sanitized, including nested transcripts, phone fields, and secret-like fields.\n- CLI live failures return coarse errors without raw provider details.\n- Added regression tests for all of the above; 4 tests pass, and repository validation passes.

Ray-56:

PR #512, opened 2026-09-12
feat(ringdown): two note vendors, acknowledgements in Spanish, and evidence for the calls that never rang

gmassello:
**What to evaluate.** Ringdown is an on-call escalation agent that phones the pager holder and
records an acknowledgement only when the recipient said, in their own words, that they are taking
the incident, confirmed their own name, and gave a number of minutes. Two out of three rings the
next person. The refusals are the product: *"yeah, sure, I'll take a look at some point"* has no
owner and no clock in it, and is not recorded as if it did. The call is placed over REST and
verified over MCP, because a result confirmed through the transport that wrote it proves nothing.

**What does not work, before what does.** That last step does not close against the live provider.
`get_call_run` indexes calls by a `run_id` that only its own placement tool hands out, and no
identifier a REST-placed call exposes resolves to one — so there is no run to re-read, and **every
live verdict settles at exit 45**: the second channel said nothing, never that it disagreed. It is
ceiling 12 of the twenty-three in the app README, the provider's own responses are committed as
fixtures with their provenance, and a test asserts that this parser cannot read a real run. The gap
is a tested fact rather than a surprise left for the next person.

**Relation to #205.** That PR was merged on 25 August with the tree it had on the 24th. This one
brings the app up to everything done since, and adds the entry it never got in the root README —
`grep -i ringdown README.md` returns nothing on `main` today, so in an awesome list the app is
effectively absent.

**What changed, in five lines.**

- A call can now be answered in Spanish, can ask to be called back, cannot commit with a qualifier
  attached, and confirms a name only as an answer to the question that asked for it.
- A call the provider ends in the second it starts is recorded as a call that never rang, not as
  nobody answering; a ledger that cannot be written after a call was placed exits 50, not 30.
- The verdict goes back to PagerDuty **or Opsgenie** through a mapping file with no vendor code, and
  a model may draft that file — the loader decides whether it holds.
- Nine calls were placed against the live provider, six on 2026-08-20 and three more on 2026-09-13.
- The terminal report is the surface a run is actually read through, and it had one test behind 216
  lines. It now has eighteen, one for each of its ten public functions and one per reason a call can
  fail for. The package is cut at `0.2.0`.

**Where to look first.** The two Must Fix items that blocked #205 are answered in the last section
below: each credential is still pinned to its own exact origin, and `assert_trusted_url` takes a
sequence only for the note channel, which is one channel with two service regions. The suite is
**606 tests**, no credentials and no outbound calls; twelve of them read the project site, which
does not travel upstream, so a run in a checkout of `apps/python/ringdown/` alone reports 594 passed
and 12 skipped.

Three minutes of it running, the last fifty seconds of them a real call placed against the live
provider: <https://youtu.be/cZg4hbkCVMA> · the ladder, the ledger and the receiver as live widgets:
<https://gmassello.github.io/ringdown/>

---

<details>
<summary><b>What a call can be answered with now</b> — Spanish, callbacks, hedges, and the name as an answer</summary>

<br>

- **Spanish, on the same rules.** The phrase tables carry Spanish beside English — commitments,
  declines, qualifiers, callbacks, voicemail greetings, wrong-number answers, spoken numbers, the
  two questions a script has to ask, and the three injection families. One table per family rather
  than one per language, and no language flag: a real on-call call code-switches, and the language
  is not known until somebody answers. `normalise` folds accents, so `José` and `Muñoz` confirm an
  owner where they used to be read as `jos` and `mu`. The gates do not soften — *"creo que lo tomo
  yo, tal vez"* settles `not_acknowledged` exactly as *"I think I'll take it, maybe"* does.
  `examples/guardia.script.txt` is a call script in Spanish, which the validator used to refuse
  outright, and scenario 8 of the demo runs the whole ladder in Spanish end to end.
- **"Call me back in ten minutes" is heard, and is still not an acknowledgement.** The attempt
  settles `not_acknowledged` with the reason `callback_requested` and the words that carried it
  quoted beside the minutes. What changes is where the ladder goes next: it waits and calls that
  rung once more, but only if the minutes asked for plus one more call fit inside what is left of
  `ladder_timeout_seconds`. Ask for ninety on a fifteen-minute ladder and the next rung rings now,
  with the request recorded rather than granted. A rung is a *scope*, not a person, so if the shift
  changed while it waited the second call goes to whoever covers that scope now — the person going
  off shift is not rung again. One rung rings at most twice, ever.
- **A commitment with a qualifier is not a commitment.** "I'll take it, but I'm not sure I can get
  to it" used to settle acknowledged on the strength of four words inside a hedge. It now settles
  `hedged_acknowledgement`, with the qualifier quoted on the terminal and in the ledger so it is
  clear what was heard. A negation before the commitment is caught the same way, and only before —
  "Yes, I am taking this, nobody else is around" is still an acknowledgement.
- **The name counts only as an answer to the question.** A name spoken before "Am I speaking with
  {name}?", or three turns later while talking about somebody else, confirms nobody. The answer may
  carry it as "this is Alice", "Alice speaking", "I am Alice" or "soy Alice" — the last two added
  after a live call was answered "Yes. Hi. I'm German Marcelo." and correctly took the incident, and
  was not recorded as its owner. Assent with no name in it still confirms nobody: the only name on
  such a call was spoken by the agent, and a recorded field has to be quoted from the recipient.

</details>

<details>
<summary><b>What the provider does, and what gets recorded about it</b> — Dropped calls, a ledger that cannot be written, and the payload as data</summary>

<br>

- **`zero_duration`.** Four of six calls placed against the live API on 2026-08-20 ended in the
  second they started, nothing transcribed, reported as `Hangup by: user` — while the carrier that
  owns the destination number had no record of any of them. The snapshot was dropping the two
  signals that show it, so those calls reached the verdict as a generic failure, indistinguishable
  from nobody answering, and an exhausted ladder reported that the incident had no owner when no
  telephone had rung. The snapshot now carries the attempt's measured duration — the evidence, not a
  conclusion drawn from it, and `None` where there is nothing to measure — the reason is recorded
  whatever status the provider reports, and the run says how many calls ended that way beside the
  verdict. The reason is named for what the payload shows and the README says plainly what it cannot
  prove: somebody who answers and hangs up inside the same second is indistinguishable.
- **Exit code 50, and a `LedgerError` to carry it.** A ledger that cannot be opened, read or extended
  *once a call has been placed* is an infrastructure failure, not the operator's mistake it used to
  be reported as. Before the first call it is still 30. The run prints the verdict and the attempts
  it reached before exiting, because that output is the only surviving evidence.
- **The incident payload is data too.** Quotes in `title`, `summary` and `service` are neutralised
  where the task is formatted, so an alert payload cannot close the wrapper it is read inside.

</details>

<details>
<summary><b>Where the verdict goes</b> — Two note vendors, a configurable script, and a model that drafts config</summary>

<br>

- **The note back, now to two vendors.** `ringdown/notes.py` — renamed from `pagerduty.py`, which
  was named after one of its two vendors — posts a note quoting what the engineer actually said, on
  the incident or alert the page came from. A note, never an acknowledgement: suppressing the
  paging tool's own escalation on the strength of a phone call stays the operator's decision. No
  vendor code in the app: the incident arrives through a mapping file.
- **A second vendor example, for the adapter.** The adapter claimed to be vendor-generic on the
  strength of one example. Opsgenie's documented `Create` payload carries no priority, no
  description and no alert URL, so the mapping file absorbs all three gaps — a tag index for
  severity, a literal summary, no `runbook_url` key at all — and `ringdown/adapter.py` is untouched.
- **The call script is configurable.** `ringdown/task.py` lets an incident file replace what the
  agent says, so the same ladder, verification and ledger chase a supplier over a missed service
  level with no code that knows about SLAs — `examples/sla-breach.example.json` ships one. A script
  is refused if it drops the sentences the extractor and the injection defence depend on.
- **A model can draft the mapping file.** `ringdown/suggest.py` asks Gemini for one and then runs it:
  `adapt` executes the draft against the real payload and the incident loader validates the result
  before anything reaches disk. A rejected draft goes back to the model once, carrying the loader's
  own words; a second rejection ends the command. The model writes config, never the verdict, and
  `tests/test_layering.py` asserts that only the CLI imports the module at all. Without
  `GEMINI_API_KEY` the subcommand exits 30 without opening a socket, and nothing in `preview`, `run`
  or `verify` reaches for a model.

The suite is **606 tests**, no credentials and no outbound calls. Twelve of them read the project
site, which does not travel upstream, so a run in a checkout of `apps/python/ringdown/` alone
reports **594 passed, 12 skipped**. The README's setup block said `pip install pytest` and "295
tests"; both were stale and now say `uv sync` / `uv run pytest -q`, with the skip count spelled out.

</details>

<details>
<summary><b>How the live calls happened at all</b> — The receiver, and the nine calls it made possible</summary>

<br>

Ceiling 14 is that CALL-E refuses Argentina — `Region is not allowed for this channel` — which is
where a project written from Buenos Aires normally stops having a live test.
`apps/python/calle-receiver/` is the way around it: a small FastAPI service that takes the agent's
call on a US Twilio number and bridges it to an Argentine phone, recording and transcribing both
tracks. It is demo infrastructure, not the product — it ships in this repository's own tree and does
not travel upstream, it shares nothing with the app, and the app's README never names it.

It is what produced the evidence the rest of this PR rests on: nine calls placed against the live
provider, six on 2026-08-20 and three more on 2026-09-13, the fixtures in `tests/fixtures/` rewritten
from their responses, and the finding that four of the first six were reported as `DECLINED (Hangup
by: user)` while the Twilio account that owns the destination number has no record of them — which
is only provable because we own both ends. It also answered a question the docs do not: CALL-E does
dial VoIP numbers. The calls that connected completed normally, 82 seconds bridged end to end.

The sample call is open at [`/demo`](https://calle-receiver.onrender.com/demo), masked and written
into the page rather than the database. The inbound log at `/calls` keeps its password, because that
one prints the number that actually dialled and the words that were actually said. It runs on a free
tier: give it thirty seconds to wake.

</details>

<details>
<summary><b>The two Must Fix items from #205 still hold</b> — Where to look — each credential pinned to its own origin</summary>

<br>

Both areas this PR touches are the ones that blocked #205 twice, so here is where to look:

- **Each credential is still pinned to its own exact origin.** `RestClient.LIVE` and `McpClient.LIVE`
  are each a single URL, compared whole, and `run` pins through those class attributes rather than
  the module constants, so the classes are the only source of truth. Swapping `--base-url` and
  `--mcp-url` is refused before a socket opens.
- **`assert_trusted_url` accepts `str | Sequence[str]`, and only the note channel passes a
  sequence.** PagerDuty and Opsgenie each publish two service regions for one account and one
  credential (`api.pagerduty.com` / `api.eu.pagerduty.com`, `api.opsgenie.com` /
  `api.eu.opsgenie.com`); that is one channel with two homes, not two channels. Both call channels
  still pass one string each, so the cross-send described in #205 remains impossible.
- **The note channel uses the same loopback rule, and has the negative tests to prove it.** Against
  loopback the live variable's name is never constructed: a run with `PAGERDUTY_TOKEN` set and no
  `RINGDOWN_FAKE_PAGERDUTY_TOKEN` refuses the note and the live token never appears in the output. A
  note token is also refused against a call channel's URL. `ringdown/notes.py` refuses to follow
  redirects, so a 302 cannot carry the `Authorization` header off-host, and the ledger records the
  hostname, not the URL or the token.
- **The model channel is pinned the same way.** The endpoint is a constant with no flag to point it
  anywhere, the key travels in a header rather than a query string, and redirects are refused. The
  one thing that leaves the network is the alert payload the operator named, which is why it is an
  opt-in subcommand and not a fallback inside `adapt`.
- **No new fixture, example or test carries a provider response.** Every fixture states its
  provenance in its own `what` / `source` / `unobserved` / `why_it_matters`, and every value in them
  was rewritten: no real number, identifier, recording or transcript is stored — the shape is what
  is evidentiary. `tests/fixtures/README.md` repeats, where a search for "live" lands, that the same
  holds for the whole directory. Every address is under `example.com`; the only phone numbers the
  diff adds are already masked.

Two smaller findings from auditing this delta against the same criteria, fixed here rather than left
for review: an incident's `script` field could name a path outside its own directory (`pathlib` makes
`parent / "/etc/shadow"` absolute), which made the error message an existence oracle even though no
system file passes `validate_task_template`; and `assert_trusted_url` checked the scheme before the
userinfo, so `ftp://user:hunter2@host` printed the password on its way to being refused.

</details>


gmassello:
Added a second commit — documentation only, no code change.

It fills three gaps in the app README's threat model, all of them things the code already did and the document either described the old way or did not mention:

- **The `script` field reaches the filesystem and had no entry**, although every other incident field that arrives from an alert payload gets one. It is contained to the incident's own directory, and the refusal names the directory rather than the path that was asked for, so the error cannot be used to learn which paths exist. Checked against an absolute path, a `../` walk and a file that does not exist: all three produce the same message.
- **`suggest-mapping` is a third pinned endpoint**, and the threat model named only the two CALL-E channels. It is the tightest of the three — no host flag at all, so the only place it can reach is the constant in its source; the key travels in a header rather than a query string; a redirect is refused rather than followed; and it exits 30 without opening a socket when the key is absent. The one cost no pin removes is that the payload you name leaves your network, which is why drafting a mapping is a separate opt-in subcommand rather than a fallback inside `adapt`.
- **The userinfo example was already there**; what was missing is why that check runs before the scheme check. Only the scheme refusal interpolates the URL, so checking userinfo first is what keeps a credential out of the error message.

`python3 scripts/validate_repository.py` passes. The app suite reports 442 passed, 7 skipped in this checkout — the seven read the project site, which lives in the source repository and deliberately does not travel here.

gmassello:
Pushed one commit, no force-push, so "changes since your last review" still works.

The short version of what is in it: the extractor reads acknowledgements spoken in Spanish on the
same rules as English, with no language flag — one phrase table per family rather than one per
language, because a real on-call call code-switches and the language is not known until somebody
answers. `pagerduty.py` became `notes.py` now that the verdict is written back to Opsgenie as
well. Asking to be called back is heard and is still not an acknowledgement: the ladder waits and
calls that rung once more, but only if the wait plus one more call fit inside what is left of the
escalation deadline. A commitment with a qualifier is no longer a commitment.

Two of them came out of placing real calls rather than out of the fake, and those are the ones
worth a look:

**A call the provider ends in the second it starts is now recorded as what was seen.** Four of six
calls placed against the live API on 2026-08-20 ended that way, nothing transcribed, reported as
`Hangup by: user` — while the carrier that owns the destination number had no record of them. The
snapshot was dropping the two signals that show it, so an exhausted ladder reported that the
incident had no owner when no telephone had rung. The snapshot now carries the attempt's measured
duration, which is the evidence rather than a conclusion drawn from it, and the run says how many
calls ended that way beside the verdict. The reason is named for what the payload shows, and the
README says plainly what it cannot prove: somebody who answers and hangs up inside the same second
is indistinguishable.

**The owner table was missing the answer people actually give.** On 2026-09-13 a live call asked
"Am I speaking with German Massello?" and was answered "Yes. Hi. I'm German Marcelo." — the
recipient then took the incident and gave a number of minutes, and the attempt settled
`not_acknowledged` for want of a confirmed owner, because the table only read "this is {name}",
"{name} speaking", "soy {name}" and "habla {name}". It reads "I am {name}" and "I'm {name}" now.
That phrasing is grounded exactly as well as the other four and was simply missing. Assent with no
name in it — "yes, I am", "speaking" — still confirms nobody, since the only name on such a call
was spoken by the agent, and a recorded field has to be quoted from the recipient. Ceiling 17
carries both halves and the cost of each.

The suite is 589 tests; 577 pass and 12 skip in a checkout of `apps/python/ringdown/` alone,
because they read the project site, which does not travel upstream.

PR #519, opened 2026-09-13
docs(apps): add safeops awesome-list entry

saikumar040060:
## What

Adds one line to the Apps section of README.md for **SafeOps**
(https://github.com/saikumar040060/safeops) — an independent runtime
authorization layer (permission, deterministic policy, risk engine,
human approval) that a completed CALL-E call's outcome is routed through
before any action executes. CALL-E places the call and produces a
structured result; SafeOps decides whether the action that result
implies is actually allowed to happen.

## Why this belongs here

CALL-E gives code a phone line, but a phone call is unstructured, human,
and can be adversarial — exactly the kind of input a security-conscious
system should never trust blindly. This integration treats a completed
call the same way an authorization system should treat any other
untrusted input: the caller's own words are attached to the downstream
action as untrusted context, and a manipulated or adversarial call can
still be blocked even when it targets a tool the calling agent is
otherwise permitted to use.

## Live-verified evidence (not simulated)

- A real outbound call confirmed a $750 duplicate-payment refund; the
  confirmed outcome reached `REQUIRES_APPROVAL`, a human approved it, and
  it executed exactly once (verified against the backing database:
  exactly one refund record, complete audit trail).
- A separate manipulated-transcript case (a call outcome shaped like a
  prompt-injection attempt) was blocked by the risk engine — `BLOCKED`,
  a security incident recorded, zero side effects.
- A real voicemail/no-answer case was also exercised and handled
  correctly: no confirmation was obtained, so no downstream action was
  ever submitted.

Full transcripts (phone numbers, account identifiers, and other
sensitive fields redacted/masked), IDs, and the complete audit-event
sequence for all of the above are in
[`submissions/calle.md`](https://github.com/saikumar040060/safeops/blob/main/submissions/calle.md)
and [`docs/calle.md`](https://github.com/saikumar040060/safeops/blob/main/docs/calle.md)
in the linked repository.

## Checklist

- [x] No secrets, tokens, private phone numbers, or third-party personal
      data in this PR or in the linked evidence (redacted/masked)
- [x] States the provider (CALL-E) and host (SafeOps, an independent
      Python/FastAPI backend) it supports
- [x] Describes side effects (a confirmed call outcome can trigger a
      real, human-approved action; a call places a real phone call by
      default, documented in the linked repo)
- [x] Install/usage instructions in the linked repository's README and
      `docs/calle.md`
- [x] `python3 scripts/validate_repository.py` passes locally
- [x] Minimal diff: one line added to `README.md`, no other files changed

## Disclosure

The linked SafeOps project is a pre-existing, independently built
authorization backend (predates this contribution); the CALL-E
integration itself (`integrations/calle/`) is the newly built/verified
piece this entry describes. This is disclosed explicitly in the linked
repository's own submission documentation.


Ray-56:
Superseding review for head 028ec3b0: I applied the bounded documentation cleanup: Label evidence as author-reported one-call testing of a seeded local refund, not a real payment or universal exactly-once guarantee. Repository validation passed without credentials or real calls. No current Must Fix remains; this is reviewed as a community reference, not external application certification.

PR #524, opened 2026-09-13
feat(apps): add supplier quote agent

guptachetan1995:
## Summary

Adds `apps/web/supplier-quote-agent/`, a runnable procurement-call app built on the CALL-E Calls API.

An agent can plan a supplier call, dial an approved one, cancel a call in flight, and re-plan one that failed. It cannot approve a call. Approval is a dashboard button a person presses, and that is enforced structurally rather than by prompt instruction:

1. `approve_task` / `reject_task` are never added to the tool registry, so tool discovery cannot surface them;
2. both refuse any actor but the owner even when called directly on the internal `invoke()` chokepoint;
3. the generic `update_task` write path refuses to set those statuses at all.

`place_call` then refuses any task that is not already `approved`, and its refusal names the remedy: *"Ask the owner to approve it from the dashboard — no tool call can approve a task."* Re-planning a finished call returns the task to `planned`, never `approved`, so a second call to the same supplier needs a second human approval.

Both callers — the dashboard's buttons and the agent's tools — go through one endpoint (`POST /api/invoke`) and one function (`invoke(tool, args, actor)`). Each registered tool's `execute()` is a single delegating line, so a tool cannot reach the data store without passing the same guards a button passes. Every invocation, refusals included, is appended to an activity log with the actor's name.

Calls go through a `CallProvider` adapter. The default `FakeCallProvider` is deterministic — no network, no real timers, canned outcomes from JSON — so the entire test suite and the whole demo run with no credentials and place no calls. `CallEProvider` is the real integration and takes its `fetch` by injection, which is how its request shape and response parsing are unit-tested against fixtures without opening a socket. It refuses to dial when `CALLE_API_KEY` is unset.

Node 20+, Express, React, Jest, ESLint. 65 tests across 6 suites; `npm test` and `npm run lint` both green.

**Demo video:** https://youtu.be/3MOBu8sBQvE

## Verified against the live service, not just fixtures

`CallEProvider` has been exercised end to end against `api.heycall-e.com` with a real billed outbound call — transcript of the run, activity log and outcome in [`docs/real-call.md`](apps/web/supplier-quote-agent/docs/real-call.md). The call reached a terminal state in ~2m50s and returned a schema-valid structured result:

```json
{
  "outcome": "quoted",
  "summary": "Acme Corp provided a quote for WIDGET-42 quantity 100: unit price $10 per unit, lead time around 7 days, and minimum order quantity 50 units.",
  "next_action": "review_quote"
}
```

Two things that run surfaced, both now reflected in the code and docs rather than glossed:

- **The call is genuinely asynchronous.** `POST /v1/calls` then polling `GET /v1/calls/{id}` until `completed`/`failed`/`canceled` is required; a create-and-read would return nothing usable.
- **`cancel_call` cancels the task, not the phone call.** The Calls API "does not expose an operation for clients to cancel a call after it has been created", so a call already dialing runs to completion. The README says this plainly rather than implying the dashboard's Cancel button hangs up a live call.

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

### Evidence

| Line | Evidence |
|---|---|
| English only | No CJK anywhere in the app; your `validate_english_only()` covers `apps/` and passes. |
| Naming conventions | `python3 scripts/check_branch_name.py --branch feat/supplier-quote-agent` exits 0. Branch created with `scripts/create_branch.py`, `core.hooksPath` set to `.githooks` so the pre-push hook ran. |
| No secrets | Only environment-variable *names* appear (`CALLE_API_KEY`, `CALLE_BASE_URL`, `CALL_PROVIDER`, `DEMO_SUPPLIER_PHONE`), never values. `.env.example` ships with empty values; no `.env` is tracked. The one credential-shaped literal in the tests is `'test-key'`, passed to an injected `fetch` that never leaves the process. The real call's recipient number was supplied via an untracked env var and is masked in `docs/real-call.md`. |
| Side effects described | `README.md` opens with a "Safety model" table and has dedicated "Credentials and real calls" and "Cancelling and rolling back" sections. `docs/architecture.md` has "Where a real call could happen" — exactly one function, gated on two environment variables. |
| Fictional phone numbers | Every number committed is in the NANP block reserved for fiction, `+1-555-0100`…`+1-555-0199`. The app's own `verify.sh` enforces this with a grep that fails the build on anything outside that range. |
| Cancellation | No recurring or scheduled workflows — no cron, queue or retry daemon; a call happens only on an explicit `place_call` against a human-approved task. `cancel_call` aborts an in-flight call and `retry_with_plan` rolls a task back to `planned`, with the real-provider limitation stated above documented rather than overclaimed. |
| No-call path by default | `FakeCallProvider` is the default; the real one requires `CALL_PROVIDER=calle` **and** `CALLE_API_KEY`. `tests/call-flow.test.js` asserts `process.env.CALLE_API_KEY` is undefined and that `place_call` still succeeds. |
| Validator passes | `Repository validation passed.` (exit 0), run in this branch. One note for you: run it *before* `npm install` in the app directory — with `node_modules/` present it walks into transitive deps and reports `generator-function` as a local file dependency. On a clean checkout it passes. |

## Safety notes

- **Real-world side effect:** with `CALL_PROVIDER=calle` and a valid `CALLE_API_KEY`, `place_call` places a genuine outbound phone call, billed to the CALL-E account behind the key. Without both, no call is possible and the provider refuses.
- **No autonomous dialing:** an approval is per-task and does not persist across a retry.
- **State is in memory:** restarting the process is itself a complete rollback.
- **No authentication:** `actor` is a string in the request body. This is a local demo app; anything internet-facing would need the owner identity to come from a real session.

## Testing

```bash
cd apps/web/supplier-quote-agent
npm install
npm test          # 65 passing, 6 suites
npm run lint      # 0 errors
bash verify.sh    # docs present, fictional-number guard, then test + lint
```

No test opens a socket, sets a real API key, or constructs `CallEProvider` with the real `fetch`.


Ray-56:
Remaining Must Fix:

- `apps/web/supplier-quote-agent/src/server.js` and `src/invoke.js`: enforce local-only access or authenticate owner actions. Caller-supplied actor/status must not approve a new task, and changing an approved recipient/plan must require renewed approval.
- `src/invoke.js` and `src/providers/calle-provider.js`: remove HTTP-supplied provider/live/baseUrl overrides, pin the approved HTTPS credential origin, and validate an explicitly authorized ASCII E.164 destination before live dispatch.
- `src/server.js`: mask phone-bearing user-facing state/activity and result text.
- `src/invoke.js` and README: stopping local waiting must not be reported as a canceled provider call; leave unknown outcomes for reconciliation and qualify mid-ring cancellation claims.

guptachetan1995:
Thanks for the detailed review — all four of these are fixed, plus two of the same shape a follow-up self-review turned up:

- **Approval bypass** (`create_task`/`update_task`): both now strip any caller-supplied `status`/`approvedAt`/`approvedBy`/`rejected*` unconditionally — a task is always born `pending`, and `update_task` forces an already-approved task back to `planned` the moment it changes `plan` or `suppliers`. A follow-up review found the fix was incomplete: `POST /api/invoke` defaulted a missing `actor` to `"owner"` (reproducible: omit the field, get owner privileges for free) and `update_task` could still overwrite a task's own `id`. Both closed too — the server now also binds to `127.0.0.1` only and refuses non-loopback requests.
- **Provider/credential surface** (`place_call`/`CallEProvider`): `provider`/`providerOptions` are no longer readable from request args at all — which provider runs is `CALL_PROVIDER` (environment) alone, and the real provider is always constructed with zero per-call overrides. `CallEProvider` now refuses to construct against a non-`https://` base URL and refuses to dial anything but a clean ASCII E.164 number.
- **Phone masking**: every API response now masks `phone`/`phones` fields, and additionally scrubs any of those same numbers if they reappear in unrelated free text (e.g. a call summary).
- **Cancellation semantics**: against the real provider (no client cancel operation), `cancel_call` now reports the task as `cancel_requested` with an "outcome unknown, pending reconciliation" message, never `cancelled`.

Full write-up with file:line detail is in [`docs/pr-body.md`](https://github.com/guptachetan1995/awesome-phone-call-agents/blob/feat/supplier-quote-agent/apps/web/supplier-quote-agent/docs/pr-body.md#changes-since-the-last-review) § "Changes since the last review". 103 tests now (was 65), including a new suite that drives the actual Express app over a real socket — nothing previously exercised that HTTP layer, which is exactly how the actor-default gap went unnoticed the first time. `npm test`, `npm run lint`, and `bash verify.sh` all green, and `scripts/validate_repository.py` passes against this branch.

Happy to address anything else.

Ray-56:

1. Pin the parsed approved HTTPS credential origin and reject redirects; checking only https:// still permits arbitrary hosts to receive CALLE_API_KEY.
2. Sanitize grouped/other phone-bearing summary and error display copies, not just exact known phone-field values.
3. Reject incomplete/synthetic sample destinations and require an explicit authorized valid live destination; the current parser admits bundled short examples for live submission.

guptachetan1995:
Thanks for the detailed second pass — all three items are closed, each with tests that fail
against the previous code.

1. **Pin the parsed approved HTTPS credential origin and reject redirects.**
   `CALLE_BASE_URL` is now parsed with `new URL()` and must resolve to exactly
   `https://api.heycall-e.com` — the parsed *origin*, checked against a one-entry allowlist,
   not a string prefix — with no userinfo, port, path, query or fragment accepted. A
   look-alike host, another port, or `http://` is refused at construction, and the refusal
   never echoes the configured value. Every request sets `redirect: 'error'`, so a 3xx is
   never followed and the `Authorization` header is never replayed; independently, a
   response reporting `redirected` or whose `url` lands on another origin is refused before
   its body is read.
   `src/providers/calle-provider.js` (`pinApiOrigin`, `_assertPinnedResponse`); tests in
   `tests/providers.test.js` — *"CallEProvider — the API key only ever reaches the pinned
   origin"*, including a real loopback server answering `302` with Node's own `fetch`.

2. **Sanitize grouped/other phone-bearing summary and error display copies.**
   Free text (summaries, transcripts, error messages, activity-log arguments, ids) is now
   masked by shape, not by nearby words — two of our own review rounds found that any rule
   reading context ("call", "order", "Accounts:") could be talked into leaking a number in
   one direction or the other. Every run of 7+ digits is masked unless it has one of a
   handful of exact shapes of its own (a date, a clock-time range, a decimal, an IPv4
   address, a "/" quantity list, an amount after a currency sign, or part of an
   identifier). The trade-off is deliberate and pinned by tests: a long bare reference or
   quantity ("PO 1234567890") is masked along with phone numbers.
   `src/mask.js`; tests in `tests/mask.test.js`, `tests/server-http.test.js` (checks the raw
   socket, not just the return value), and a permanent audit,
   `tests/non-phone-digit-runs.test.js`, that holds every phone-shaped digit run anywhere in
   this app's own files (national spellings included) to the fictional-number table or a
   short, reasoned allowlist.

3. **Reject incomplete/synthetic sample destinations and require an explicit authorized
   valid live destination.**
   Destination parsing now requires a complete E.164 number (for `+1`: a real area code and
   exchange, neither starting 0/1, no N11 code), refuses anything in a range reserved for
   fiction (NANP 555-01xx, Ofcom's drama ranges) however it's spelled, and — new this
   round — refuses to dial anything at all unless it's listed in
   `CALLE_ALLOWED_DESTINATIONS` (validated at construction and at server startup). Unset
   means no live destination is authorized by default.
   `src/providers/calle-provider.js`, `src/fictional-numbers.js`; tests in
   `tests/providers.test.js` — *"only an explicitly authorized, complete, real destination
   is dialled"*, fed by the same fictional-number table the repo-wide scan checks samples
   against, so "fictional in the repo" and "refused live" can't drift apart.

Full detail, including two things our own adversarial review found and fixed on top of
this (a `CALLE_API_KEY` holding a newline, and real-provider misconfiguration now stopping
the server at boot) is in the updated PR description.

Not in this round, and flagged as non-blocking follow-ups rather than gaps in the above:
showing the exact dial target on the approval card, and per-country number validation
beyond NANP (today the operator allowlist is the completeness check for non-`+1` numbers).

Ray-56:
All three frozen Must Fix items are closed: pinned HTTPS credential origin/redirect refusal, masked public result/error copies, and explicitly authorized valid live destinations. The 269 fake/loopback tests and lint check pass; no provider credentials or real calls were used. This supersedes the previous blocking list. Local-only scope and documented demo limitations remain acceptable; production hardening is optional.

PR #539, opened 2026-09-13
feat(deskhelp): add scheduled office call desk

vickysharma-prog:
## What this adds

`apps/web/deskhelp` — a catalogue entry for DeskHelp: an assistant for every call an office has to make, that speaks to people on its behalf and reports what each call settled.

Three files, no code. The application is MIT at <https://github.com/vickysharma-prog/DeskHelp.ai>.

- Live demo: <https://deskhelp.onrender.com>
- Demo video: <https://www.youtube.com/watch?v=bN7Rh2KrZKo>

A tour: the dashboard, the fourteen workflows, a dry run showing the exact words the caller would follow, the call log, and the review queue

## The problem

Every office has a call list that refills itself every week. Money owed on Friday. Somebody who has not turned up. Twenty people who enquired in June and were never called back. It takes a person most of a week, and the week after that it starts again.

A workflow here is a short description of who to call, what may be said to them, and what a useful answer looks like. That is all it is. The engine has no idea what a student, an invoice or a delivery is, so the same code runs a fee reminder, a staff attendance check, a supplier confirmation or a feedback survey. Fourteen ship ready for education hubs, because that is where the work is heaviest, and all fourteen ship switched off.

## What it did on a real phone

Placed from the product to a handset I own, in the mix of Hindi and English families actually use. The parent asked for a discount mid-call:

> **Parent:** कुछ डिस्काउंट मिल सकता है क्या इस महीने मेरे को?
> **DeskHelp:** Discount ya concession ke baare mein mujhe yahan koi jaankari nahi hai. Yeh cheez office hi confirm kar sakta hai.
> **DeskHelp:** Wapas apne sawaal par aate hain — aap sochte hain ki payment kab tak ho jayegi?

It declined, then went back to its own question rather than following the parent away from it. Asked next whether two months could be paid together, it declined again. Both questions reached the review queue in the parent's own words, and the call closed by reporting only what it had noted.

A real call read back inside the product: the transcript with the two speakers apart, and underneath, the answers it was willing to keep

On another call a parent said they planned to join. That was stored as a claim carrying their exact sentence, with `confirmed: false`, and nothing treated it as an enrolment. On a third, somebody answered for the named student rather than as them; `identity_confirmed` came back `no` and not one answer was attributed to that student.

## How CALL-E is used

CALL-E places the call. DeskHelp decides who is called, when, what may be said, and what the answer meant. CALL-E does not do recurrence, and that is the part an office actually asked for: "remind the unpaid families two days before the month ends" is a sentence about a calendar, not about a phone.

Batch `recipients[]` with per-recipient `region` and `locale`, a closed `result_schema` built per workflow, `Idempotency-Key` derived from the authorisation rather than the attempt, `metadata.workflow_run_id`, terminal polling with no automatic resend, and `transcript_turns` plus `evidence_quotes` behind every stored answer. `completion_confidence` is stored beside DeskHelp's own reading and decides nothing: 0.99 does not save an answer the recipient never supported, and 0.1 does not spoil one they did.

Hindi, English, and the code-switched mix. `hi-en` has no locale at any provider, so the call goes out as `hi-IN` and the instruction to follow the recipient's language lives in the task text.

## Safety

- **Two gates before a phone rings.** `DESKHELP_LIVE` exactly `"true"`, **and** the destination on an operator-typed allow list. No wildcard, so one gate left on in a shell profile still reaches nobody.
- **Every workflow ships off.** The dry run renders the exact instruction, reserves nothing, and reaches no transport.

  A dry run showing the exact instruction the caller would follow, including everything it is forbidden to do
- **One press is one call to one named person**, with the name and number on the button.
- **The reservation is written before dialling**, so a crash blocks the retry rather than producing a second call. An ambiguous submission is recorded and stopped, never resent, and a submitted call cannot be recalled.
- **A hang-up is `declined` and never retried**; a phone that rang out is `unreached` and may be tried once more. An attempt whose start and finish are the same instant never rang anybody, so it reaches a person instead. Anything unrecognised reaches a person, per ADR 0006.
- **An answer is kept only when its quote appears in a turn positively identified as the recipient.** CALL-E returns `bot`, `user` and `unknown`; on a live call the `unknown` turns carried the agent's own words, so a label nobody recognises grounds nothing.
- **Timezone and jurisdiction are declared, never inferred.** India's 9am to 9pm TCCCPR window is enforced in the declared timezone.
- **On a call about a child**, the first name and class are spoken only after the answerer confirms they are the named guardian. Anyone else, and voicemail, hear only that the office called.

## Validation

```
python scripts/validate_repository.py   # Repository validation passed
```

In the application repository, `npm run check` runs a typecheck and 174 tests weighted towards the refusals, and `npm run demo` runs the whole loop on fixtures with no network and no credentials.

Live calls behind this submission were placed by me to handsets I own. No real number appears in the diff; fixtures use the reserved `NXX-555-01XX` range and every number shown or logged is masked.


Ray-56:
Superseding review for head 0cc0e65e: I applied the bounded documentation cleanup: Correct Node/dependency setup, translate examples into English, distinguish seeded workflows from live transport, qualify accidental-enable and timestamp claims, and repair the external ADR link. Repository validation passed without credentials or real calls. No current Must Fix remains; this is reviewed as a community reference, not external application certification.

PR #543, opened 2026-09-13
feat(plugins): add n8n-switchboard approval gate for calle calls

Kingnanaweb3:
## Summary

Adds `plugins/n8n-switchboard`: an n8n community node plus a local operator console that sits in front of a CALL-E call and decides whether it happens.

`run_call` is annotated `destructiveHint: true`. CALL-E already gates it with a `confirm_token` from `plan_call`, which is a two-phase commit. But in a workflow host the plan step hands that token straight to the run step in milliseconds, so no person is ever involved. The gate exists and is bypassed by default, not out of malice but because wiring two nodes together is the obvious thing to do.

Switchboard takes custody of the token. It is written to a local store, never emitted into workflow data (the node outputs `hasConfirmToken: true`, not the value), and nulled the moment it is spent.

This is not a calling node and does not replace `@call-e/n8n-nodes-calle`. It is a layer in front of one. Verified loading and executing on n8n 2.38.

**What it does**

- Four operations: Queue Call, Await Result, Get Status, Cancel. Cancel is the only stop that exists, since there is no cancel tool in the MCP surface.
- Policy checks before planning and again before dialing: region, blocklist, quiet hours (evaluated against dial time rather than queue time), per-recipient rate limit, per-campaign call budget, and `confirm_expires_at`.
- Host-owned scheduling. A scheduled job is not planned until it falls due, so nothing has been sent and cancelling it genuinely cancels.
- Batch queueing with per-row templating; one bad number does not stop a batch.
- Kill switch and an audit log of who approved what, when.

**Four platform observations**, each from a live run, written up in the plugin README: no cancel tool; `result.extracted` always returns the run envelope rather than a caller-defined schema; `next_step.action` can request a human with nowhere to send them; and `calle regions list` returns a documentation URL rather than data. On the schema one, Switchboard asks, validates what comes back, and reports the mismatch rather than claiming structured output it did not receive.

**Try it without dialing**

```bash
cd plugins/n8n-switchboard
cp .env.example .env
node scripts/test.js
node --env-file=.env scripts/demo.js --reset
node --env-file=.env console/server.js
```

25 tests. The demo seeds a collections queue with real tokens held, one job refused on region, one scheduled for tomorrow, and a runaway loop caught by the rate limit. `plan_call` does not dial, so it costs nothing.

Demo site: https://switchboard-eosin.vercel.app

## Type

- [ ] New skill
- [ ] New runnable app
- [x] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist


Ray-56:
Remaining Must Fix:

- plugins/n8n-switchboard/console/server.js:49-143: unauthenticated all-interface console exposes stored private call data and approval/dial endpoints using the operator's cached CLI login. Bind a local-only demo to loopback or add basic operator authentication.
- plugins/n8n-switchboard/core/policy.js:56-63 and runner.js:126-140: final dispatch checks region/blocklist/time/rate but never validates an authorized ASCII E.164 destination; the demo's synthetic ledger is also queued as live-approvable. Validate the actual destination and keep fictional fixtures no-call.
- plugins/n8n-switchboard/core/runner.js:251-255 and core/calle.js:27-33: redaction removes only the token while returning raw recipient, preview, provider text and transcript; CLI failures can expose full argument payloads. Mask public/display/log phone values and suppress sensitive provider/CLI error bodies.
- plugins/n8n-switchboard/core/store.js:71-80 versus console/public/index.html:285-288: the kill-switch confirmation promises to cancel scheduled jobs but killAll only cancels pending/approved. Include scheduled/needs-info jobs or make the promise accurately narrower. Discard held confirmation tokens for canceled undispatched jobs.
- plugins/n8n-switchboard/core/runner.js:154-156,194 and nodes/Switchboard/Switchboard.node.js:128-205: ambiguous submission/poll timeout is labeled ordinary failed and the n8n item loop proceeds to subsequent approved calls. Preserve an unknown/manual-review outcome and stop conflicting batch side effects. Add the required advisory/high-stakes boundary to docs/SAFETY.md.

Ray-56:
Remaining Must Fix for `f19758f8` under the community/demo policy: This supersedes the previous list: loopback binding, ASCII destination validation and scheduled-job cancellation are fixed. These three frozen categories remain.

1. Mask the remaining phone-bearing userInput, preview, result/transcript and error output; suppress raw CLI stderr/non-JSON diagnostics without changing private dial input (core/runner.js and core/calle.js).
2. Preserve unknown submissions and stop automatic conflicting batch work after ambiguous create/poll outcomes (core/runner.js and nodes/Switchboard/Switchboard.node.js).
3. Keep synthetic demo jobs no-call or require an explicit authorized destination override; add the previously requested advisory/high-stakes scope boundary (scripts/demo.js and docs/SAFETY.md).

PR #551, opened 2026-09-13
feat(apps): add vaxcheck school immunisation consent app

0xNoramiya:
## Summary

Adds `apps/python/vaxcheck`, a runnable app that places consent and screening calls to guardians before a school immunisation session and returns a triaged roster a school nurse can work from.

A school runs an immunisation session. Before the nurses arrive, someone has to reach every guardian and establish four things: do you consent, has the child already had this, are there allergies, are they well today. Today that travels home as a paper form in a schoolbag. A large share never comes back, so the office spends the week before the session on the phone, and on the morning there is still a row of children nobody can account for.

It was built for Indonesia's *Bulan Imunisasi Anak Sekolah* (BIAS), where that problem is at its largest. CALL-E cannot dial Indonesia yet, so the sample roster uses NANPA reserved fictional numbers and the app is corridor-agnostic.

### Design decisions worth reviewing

**One CALL-E task per student, not one task per roster.** A task reports a single `task_completed` and `completion_confidence`. Batching a roster into one task would force every child to share one score, so a voicemail at the end of the roster would drag a clean consent at the start below threshold. Per-student tasks keep each decision backed by that child's own call.

**Fail-closed triage.** A student reaches `cleared_for_school_session` only when every signal is unambiguous: identity confirmed, allergy `none` or `mild`, not unwell, prior dose clearly `no`, no unanswered question, no callback requested, `task_completed` true, and confidence at or above 0.70. Everything else routes to a nurse with the reasons listed. Voicemail, a dropped call and an empty result all land in the retry queue — silence is never consent. A cleared row means nothing was flagged; it is not medical clearance, and the roster prints that at the bottom of every run.

**Idempotency keys are derived, not generated.** `vaxcheck-<school>-<session-date>-<student-id>`, with no timestamp or random component, so a re-run after a crash resumes rather than phoning a family a second time. Duplicate student ids are rejected for the same reason.

**Roster totals are counted locally** from the triaged records rather than asked of the model, so the numbers a nurse reads are arithmetic.

### Modes

| Mode | Network | Places a call |
| --- | --- | --- |
| `--doctor` | live, read-only | no |
| default preview | none | no |
| `--mock` | none | no |
| `--preflight` | live `plan_call` | no |
| `--execute --confirm-consent` | live | yes |

`--doctor` checks CLI auth, SDK presence, that the live Developer API accepts the API key, and that the roster's region/language corridor is dialable. `--preflight` runs a real `plan_call` per student. Neither consumes call quota.

### Tests

56 total. 45 run offline with no credentials or network. The other 11 exercise the live path by installing an `httpx` transport underneath a genuine `CalleClient`, so the real SDK assembles the request and the test asserts on the exact `/v1/calls` payload, both schemas, the `Idempotency-Key` header and its determinism, one task per student with distinct keys, that the raw number never reaches the call script, that a real-shaped response triages correctly, and that a 401 or a missing call id becomes a clean error. They skip automatically when the optional SDK is absent.

### Note for maintainers

While building this I found that three corridors listed in the published *Supported Regions and Languages* table are refused by the live service: `ID`/English, `MY`/English and `MY`/Malay. Confirmed on both surfaces — `plan_call` and `POST /v1/calls`. Rejection happens after the request body validates, so a client cannot distinguish a bad payload from a dead corridor without matching on the message string, even though `unsupported_region` and `unsupported_language` already exist in the API error enum. Happy to open a separate issue if useful.

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist


Ray-56:
Remaining Must Fix for `0402cef1` under the community/demo policy:

1. Restrict live_client.py/doctor.py credentials to approved HTTPS origins.
2. Use ASCII full-match phone validation after the documented normalization.
3. Mask provider errors, nurse_contact, free-text results and preview/task output without changing private requests.
4. Bind resumed results to exactly one student/session; never fall back to students[0] or guess among siblings sharing a phone.
5. Qualify crash/rerun no-redial guarantees to the implemented finite provider dedupe and explicit --resume behavior.

0xNoramiya:
Thanks @Ray-56 — all five addressed in a8f5767, and the branch is merged with main (the README / apps README table conflicts are resolved).

1. **Origins** — new `vaxcheck/origins.py` allowlist; `live_client` and `doctor` only send the key to `https://api.heycall-e.com`. `CALLE_BASE_URL` can select an approved origin, not add one; a rejected value shows as FAIL in `--doctor`.
2. **Phone validation** — `re.ASCII` + `fullmatch` after the documented normalisation (surrounding whitespace, spaces, hyphens, dots, parentheses); tests cover Unicode digits, fullwidth `+`, zero-width characters and trailing junk.
3. **Masking** — `phone.redact()` is applied at ingestion (free-text results, summaries, evidence), to provider error messages, to `plan_call` clarifying questions, and to the preview, which now masks `nurse_contact` on screen. The task text sent to CALL-E is unchanged.
4. **Resume** — `--resume` now requires `--student` and accepts the fetched call only if its metadata names that student, school and session date. No phone matching, no fallback to the first row, so siblings sharing a guardian's number can't be confused.
5. **Claims** — README and `docs/safety.md` now describe the no-redial behaviour as CALL-E's finite, provider-defined idempotency window for an identical request, note that the app doesn't persist call ids, and point to explicit `--resume`. Both awesome-list rows say the same.

Ten new tests in `tests/test_hardening.py`; 78 total, all offline. Happy to adjust anything further.

Ray-56:

Issue #341, opened 2026-09-07
API reports task completed when voice-agent setup fails before dialing


JJasonSun:
A user testing a call to France on September 6 received a voice-agent setup error before any dial attempt, but the API still marked the task as completed.

The reported response showed:

- About 77 seconds in `queued`, followed by `failure_code: call_failed` and “The call could not start because the voice agent setup failed.”
- An empty `attempts` list and a recipient still marked `pending`.
- `task_completed: true` with `completion_confidence` 0.86 (`high`).

A task requiring a phone conversation should not be marked completed before dialing. Check the Call ID in the source thread against server logs to verify the failure and find why the API returned a completed result.

Source: https://discord.com/channels/1493880186826133504/1545967670480994374/1545975298858680410

PR #552, opened 2026-09-13
feat(airtable-calle-certa): add consent-bound verification of employment

SooryaCodes:
Certa: it verifies employment by phone, and refuses the number on the form

## Summary

**[Demo video](https://vimeo.com/1226396799)** · **[Live landing page](https://sooryacodes.github.io/certa/)** · **[Plugin README](https://github.com/SooryaCodes/awesome-phone-call-agents/tree/feat/airtable-calle-certa/plugins/airtable-calle-certa)**

Certa verifies someone's employment by phone: it calls only employers the applicant consented to, **never the phone number the applicant supplied**, and writes a typed answer back into the operator's own Airtable base instead of a voicemail.

It is a workflow plugin for Airtable built on the CALL-E Developer API. Python 3.11+, **standard library only** — no install, no build step, no lockfile, nothing hosted.

**Why it belongs here.** Verification of employment is a real, high-volume phone workflow that exists precisely because the answer lives with a human who has to be called. The incumbent payroll databases cover large employers; for everyone else somebody picks up a phone. That is the shape of problem this repository is for.

Two things make it more than a dialer wrapper:

**1. The recipient result schema is derived from the base's own field definitions.** Nobody writes JSON Schema — they create columns. A single select's choices *are* an enum, its description *is* the extraction instruction. Add a question by adding a column. Two rules come from CALL-E's docs rather than taste: `unknown` is always appended, and a colliding reserved key (`summary`, `status`, `call_id`, timing fields) is refused with the rename spelled out rather than dropped server-side.

**2. The number on the application is never dialed.** This is a type relationship, not a validation rule. The dialer's only parameter type is `ConsentedEmployerContact` carrying a `SourcedNumber`; the applicant-supplied number is a separate type with **no conversion path**, and `NumberSource` has no `application` member. An applicant listing a friend's mobile as "HR" is the fraud this is built against, and it is a control neither a payroll database nor the manual process performs.

Consent is a per-row SHA-256 over request, number, relationship and task-spec version, compared with `hmac.compare_digest`. A token filled down a column is a token for another row's number, so it fails — spreadsheet drag-fill cannot forge consent. Editing what the agent is allowed to say bumps `TASK_SPEC_VERSION`, which invalidates every consent gathered under the previous wording.

## Fail-closed behaviour

Nothing reaches a positive disposition without `reached_employer: yes` extracted from the call. This specifically guards CALL-E issue **#341** — the API returning `task_completed: true` when voice-agent setup fails before dialing, which for a verification product means *employment confirmed with no conversation*. A table whose columns cannot express that gate is rejected at setup.

`declined_to_answer: yes` is terminal with `retryable=False`. An HR department declining to confirm has not failed; redialling them is harassment. Low confidence, an unsupported positive claim, or any internal contradiction route to human review. `tests/test_calle.py` asserts that **no rule can promote a disposition** — every gate moves only toward review.

The audit log is hash-chained and append-only; records are fsynced *before* the call they authorise is dispatched, so a record with no call is recoverable and a call with no record is not.

## Live testing, stated precisely

Run against the real CALL-E API and a real Airtable base, not only fixtures. Being exact about what that did and did not establish:

- **Working, end to end.** A full product payload — REST, derived region, both schemas — completed a 51-second call over 13 turns at confidence 0.93, returning `reached_employer: yes` and `employment_confirmed: yes`. The row wrote back as **verified** with its call id and consent receipt, and the audit chain verified intact across consent, authorisation, dispatch and interpretation.
- **A finding from that run, now handled.** CALL-E keeps refining a recipient result *after* the call is terminal: `title_matches` read `"yes"` when the runner read it and `"unknown"` a minute later, so the disposition depended on when we happened to look. A terminal call is now re-read after a settle delay, and a result that moved routes to **needs review** naming both values. It does not pick a winner: neither read is knowably final, and choosing one would be the guessing the interpreter refuses everywhere else.
- **Still open.** Some attempts failed inside CALL-E's telephony leg before the destination rang, with `failure_code` varying (404, 500) across payloads whose recipient objects were byte-identical to one that connected. Not a signature this plugin's input can produce; reported to CALL-E support with call IDs and recorded rather than omitted.

One finding worth passing on from the live call: the agent asked *"is this Kestrel Logistics?"* before delivering the consent disclosure, which arrived about five seconds **after** the callee had already confirmed employment. The task text instructs disclosure first. For a product whose argument is disclosure discipline that ordering matters, and it appears to be agent-side.

Every gate moves a disposition only toward review; none can promote one

## How to review this in two minutes

No credentials, no install, and it cannot place a call in this mode.

```bash
cd plugins/airtable-calle-certa
./run.sh          # macOS/Linux, or double-click Certa.command; run.bat on Windows
```

Python 3.11+ is the only requirement. The console opens on bundled sample data
and prints `mode: sample data - places no calls`. Press **Run** and the whole
pipeline executes against fixtures, producing every disposition:

```
VR-1041  verified              VR-1043  employer unreachable
VR-1042  needs review          VR-1044  skipped
audit chain: ok, 9 records
```

The console previewing a run: one call, five cents, and the row it refuses

**VR-1044 is the one worth looking at.** It is refused before a call is even
planned, because its employer number came from the applicant rather than an
independent source. There is no setting that permits it.

Two other things that are quick to check by hand:

- Copy a consent token from one fixture row onto another. That row now fails,
  because the token is a SHA-256 over the row's own request id, number,
  relationship and task-spec version. Drag-fill cannot forge consent.
- Edit or delete a line in `audit.jsonl`. The console names the sequence
  number where the hash chain breaks.

```bash
python3 -m unittest discover -s tests -t .   # 258 tests, stdlib only
python3 scripts/validate_repository.py       # repository validation
```

## Type

- [x] New workflow plugin

## Checklist

- [x] Branch name, commit messages, and PR title follow `docs/git-naming-conventions.md` (verified with `scripts/check_branch_name.py`).
- [x] Real-world side effects are clearly described — every run states the call count and cost before dispatch and requires explicit confirmation.
- [x] Phone numbers are masked in documentation and test fixtures unless they are clearly fictional. Fixtures use the reserved US `555-01xx` range; the one India example uses the `1xx` series, which is reserved for short codes and can never be a subscriber line.
- [x] Recurring workflows include cancellation behaviour — there is no recurring schedule; consent is revocable from the console, and a revoked row can never be dialled.
- [x] Runnable code has a dry-run, fake-server, or no-call path by default. With no credentials it starts on bundled sample data and places no calls; `plan()` has zero side effects; `FixtureTransport` and `FixtureAirtable` drive the whole pipeline in tests without dialing.

**250 tests**, standard library `unittest`, no third-party test runner.

## Notes for review

- Early commits carry the scope `airtable-calle-nominee`. The plugin was renamed to Certa partway through and one commit is that rename; the history is left honest rather than rewritten.
- `plugins/README.md` has been updated with the entry for this plugin.
- The README carries four Mermaid diagrams (structure, the provenance boundary, the disposition gates, a run sequence), each validated by rendering rather than by reading the syntax.


Ray-56:
Remaining Must Fix for `dde69b36` under the community/demo policy: This supersedes earlier wording; only the remaining items below block.

1. Require HTTPS for Airtable and refuse credential-bearing redirects in airtable.py/transport.py.
2. Use ASCII full-match E.164 in types.py/dialplan.py.
3. Mask provider evidence, reasons, failure bodies and MCP/panel output while preserving private inputs.
4. Qualify unconditional VERIFIED/fail-closed/cancellation claims: missing confidence/schema-invalid values can pass, later reads can fail, and consent edits do not stop an already-snapshotted batch. Advisory results and honest limits are sufficient.

SooryaCodes:
Thanks — all four are addressed on `c3c9ffc`. Three were real defects rather than wording, so they are fixed rather than qualified.

**1. HTTPS and credential-bearing redirects**

`airtable.py` pinned the host but never the scheme, so `http://api.airtable.com` passed the hostname check and would have put the token on the wire in clear. Both clients also used `urllib.request.urlopen`, which follows redirects and replays request headers at the new host — one `302` was enough to hand an Airtable PAT or a CALL-E key to a third party while the caller saw a normal response.

New `certa/net.py` owns both decisions for both clients. Scheme and origin are checked together, and a credentialed request refuses **every** redirect rather than trying to distinguish a safe one. `tests/test_net.py` stands up a local server that redirects to another origin and asserts the request is refused, the target is named, and the bearer token does not appear in the error text.

**2. ASCII full-match E.164**

You were right, and it was worse than a tidiness issue. `\d` matches Unicode decimal digits in Python, so `+1` followed by nine Arabic-Indic digits satisfied the pattern **and** resolved to a US destination. `$` also matches before a trailing newline, so `"+15551234567\n"` passed. Now `\A\+[1-9][0-9]{6,14}\Z` with `re.ASCII`, matched with `fullmatch` in both `types.py` and `dialplan.py`.

**3. Masking provider output**

`redact()` masks anything phone-shaped in evidence strings, disposition reasons, HTTP failure bodies, `calle` CLI output and the panel's run error — every surface where text written by someone else reaches an Airtable cell, the console or the audit log. Dates and identifiers are left alone. Inputs are untouched, so the number actually dialed still travels intact to the provider.

**4. Unconditional claims**

Two of the three were holes rather than overclaims, so they are closed:

- A call carrying no `completion_confidence` skipped the floor comparison entirely and could reach `verified` with no score at all. A skipped check is not a passed one; it now routes to review.
- A value the derived schema does not define — `"probably"`, `"YES"` — was neither a contradiction nor an `unknown`, so it passed both checks and counted as an established fact. It now routes to review, naming the field and the value.

`GatesThatWereSkippable` in `tests/test_calle.py` fails if either regresses.

The third was half fixable. Consent is now re-read immediately before each call, so a revocation made while a batch is running does stop rows not yet dialed, and a re-read that fails places no call. It cannot recall a call already handed to CALL-E — there is no cancel-in-flight operation — and the README now says that rather than implying otherwise. Two further limits are stated alongside it: a failed confirming re-read leaves the first interpretation standing, and `verified` is a disposition, not an attestation that whoever answered was entitled to speak for the employer.

274 tests, `scripts/validate_repository.py` passes, branch is current with `main`.

Ray-56:
Superseding the previous review: the frozen acceptance list is closed. I made a bounded maintainer cleanup for approved HTTPS-only live credential routing and masked reason/writeback/redirect diagnostics; 251 credential-free non-panel tests plus 6 subtests passed. Fixture transports remain available; no real calls were made.

PR #555, opened 2026-09-13
feat(prepcheck): add readiness plugin

Gosho1992:
## Summary

**Contribution Area:** Workflow Plugins (`plugins/`) — Google Workspace.

Patients arrive for procedures without doing the prep: no bowel prep kit, no blood test, ate breakfast before an endoscopy. The procedure can't go ahead, the slot is burned, and the patient waits weeks for another. Appointment reminders don't catch this — a reminder confirms a patient is *coming*, not that they're *ready*.

PrepCheck turns a clinic's existing Google Sheet into a multi-checkpoint verification workflow. An hourly Apps Script sweep finds patients due a call, CALL-E places it with the checkpoint's prep checklist in context, and the structured result drives the next state. A 48-hour callback is a `next_call_at` cell, not a sleeping process, because Apps Script executions are capped at a few minutes.

**What it adds over an appointment-confirmation workflow:** every call is compared against the previous one, per item. Prep steps move independently, so a patient can collect the kit and in the same conversation reveal that a step recorded as complete was never done:

call 1 kit ✗ iron tablets ✓ prep_status = partial
call 2 kit ✓ iron tablets ✗ prep_status = partial

Overall status is identical. A workflow keyed on overall status sees nothing happen. PrepCheck records `REGRESSED: stopped taking iron tablets | net: no_improvement` and emails staff, because a regression means the clinic's record is wrong rather than merely incomplete. The same comparison catches preparation that has simply stopped moving, which escalates instead of earning a third call.

Clinic staff never open the script editor — a menu and a patient-list panel live in the spreadsheet.

**Scope boundary:** the agent verifies *whether* a step was done and nothing else. That's in the task prompt, and enforced structurally: the result schema has no field capable of carrying clinical content, so a volunteered symptom has nowhere to be stored. `examples/05-call-clinical-question.json` shows that path.

## Type

- [x] New workflow plugin

## Checklist

- [x] Real-world side effects are clearly described. Worst case per patient per checkpoint is three connected calls and two unanswered attempts; see "Side effects" in the plugin README.
- [x] Phone numbers are masked in documentation and test fixtures unless they are clearly fictional. All samples use the reserved `+1 202 555 01xx` block and are not routable.
- [x] Recurring workflows include cancellation behavior. See "Cancellation and rollback" — kill switch, per-patient stop, opt-out handling, and full removal.
- [x] Runnable code has a dry-run, fake-server, or no-call path by default. `LIVE_CALLS_ENABLED` is `false`; dry runs log the exact payload including the task prose. The six files in `examples/` exercise the entire state machine against the deployed webhook without placing a call.
- [ ] `python3 scripts/validate_repository.py` — not run; contributed through the GitHub web UI without a local clone. Happy to run it and push a fix if it flags anything.

Built for the CALL-E hackathon. Outbound calling to patient numbers in my region was restricted part-way through, so live integration was verified against the official test hotline and the multi-checkpoint workflow against synthetic results posted to the same webhook the platform posts to.


Ray-56:
Remaining Must Fix for `00c3a6cd` under the community/demo policy:

1. Fail closed on missing webhook authentication; accept matching current call/checkpoint terminal results once. Unknown, duplicate/stale, regression and do-not-call outcomes must stop for staff review rather than schedule another call.
2. Validate authorized ASCII E.164 at every live dispatch entry point.
3. Remove webhook secrets/full phones from dry-run logs and mask provider/free-text output.
4. Default demonstration staff email must be fake or separately opt-in; no-call mode currently sends real notifications.
5. Replace Sidebar.html inline onclick construction with safe text/data attributes and event listeners to close stored XSS.
6. Make preparation/slot-release statements advisory for staff review; remove false no-clinical-content schema claims and explain accepted-call cancellation limits.

Gosho1992:
Thanks — these were all real, and two of them were things I'd claimed in the README that the code didn't actually do. Addressed in 6eedf87.

**1. Fail closed on auth; accept a terminal result once**

`doPost` now rejects when no `WEBHOOK_SECRET` is configured rather than treating unconfigured as open, and rejects a missing or mismatched token. Both are logged.

A terminal result is applied only when the row is in `IN_CALL`, the payload's checkpoint matches the row's current checkpoint, and any `call_id` matches the one in flight. Anything else is stale, duplicated or out of order: it is recorded and the patient is held for staff review. You were right that this mattered — a replayed webhook previously advanced the follow-up counter and scheduled another call to a real person.

`escalate_` is now `hold_`, and these outcomes all route there with `next_call_at` cleared and the state moved out of the callable set: unknown or unreadable status, stale/duplicate, regression, `asked_to_stop`, `flag_for_staff`, and two follow-ups with no progress. Only an unanswered call leads to another dial, capped at two retries.

Added `asked_to_stop` to the result schema and to the call prompt — the opt-out is handled in code now rather than left to whoever reads the sheet next. `examples/07-call-opt-out.json` and `examples/08-call-stale-duplicate.json` cover both paths.

**2. E.164 validation at every live dispatch entry point**

`isE164_()` (ASCII, `^\+[1-9][0-9]{7,14}$`) is checked in `placeCheckpointCall_` before a row can enter `IN_CALL`, inside `createCall_` before the payload is built, and in `isDue_` so a malformed row is never selected. It applies in dry-run mode too — a dry run that would have dialled a bad number is still a bug.

**3. Secrets and phone numbers out of logs**

Dry-run logging now goes through `redactPayload_()`: the webhook URL is logged as `?token=[redacted]` and recipient numbers are masked. Provider error bodies, status strings, and anything the caller said are passed through `maskFreeText_()` — truncated, newline-stripped, `<>` removed — before they reach a cell or a log line. The CallLog is something a clinic will share and screenshot.

**4. No default staff email, no notifications in no-call mode**

`getStaffEmail_()` no longer falls back to `Session.getEffectiveUser()`. An install without `STAFF_EMAIL` set emails nobody and logs `NOTIFY_SKIPPED`. With `LIVE_CALLS_ENABLED = false` nothing is sent at all — notifications are logged as `NOTIFY_SUPPRESSED`. No-call mode now means no side effects, not just no calls.

**5. Sidebar XSS**

Rewritten. No `onclick`, no `innerHTML`, no string interpolation of server data anywhere in the file. Rows and cards are built as DOM nodes with `textContent`; the row id travels in `data-row-id` and actions in `data-action`, read back by a single delegated listener. Tone and status values are mapped through closed-set helpers so a server string can't become a class name either.

**6. Advisory outcomes, and the schema claim was wrong**

You're right and I should not have written that. `outstanding_plan` and the item arrays are free text — a patient's volunteered remark about their health can land in one of them despite the instruction not to record it. The README now says exactly that, as a limitation rather than a guarantee, and explains what does mitigate it: free text is truncated and stripped before it is stored, and `flag_for_staff` stops the workflow rather than merely tagging it.

`NOT_PREPARED` no longer says the slot is released. It records what the patient reported, holds for review, and the email states plainly that this is a report rather than a decision and that a person should confirm it. Same for regression.

Cancellation limits are documented: the kill switch means "place nothing more", not "cancel what is in flight". A call CALL-E has already accepted is out of this system's hands, the window is one sweep wide, and that is why `IN_CALL` is not a callable state.

One thing I have not done: `python3 scripts/validate_repository.py`. I contributed through the web UI without a local clone. If it flags anything I'll push a fix.

Ray-56:

1. Require the current call/checkpoint and a valid terminal result before acting; omitted IDs and unknown/incomplete/nonterminal outcomes must stop for staff review, not reach onNoAnswer_ and schedule another call.
2. Mask phone/credential-bearing free text and provider errors before logs/cells, and stop logging the webhook secret; maskFreeText_ currently only strips/truncates text.
3. Use exact full-string ASCII E.164 validation; the current JavaScript $ match accepts a trailing newline.
4. Make the live prompt advisory too: it still tells the patient the procedure cannot proceed and the slot will be released, despite the staff-review-only implementation.

Gosho1992:
Thanks — addressed in the latest commit.

**1. Current call/checkpoint required; unknown outcomes stop.** Both `call_id` and `metadata.checkpoint` are now required and must match the row in flight; either one missing holds for staff review. Status is checked against explicit lists: only `no_answer`, `busy`, `voicemail` and `unreachable` reach `onNoAnswer_`. Unknown, non-terminal (`in_progress`, `queued`), missing, `failed`, and completed-but-incomplete results all stop for review. So does an unverified identity, which previously retried.

**2. Masking.** `maskFreeText_` now redacts query tokens, bearer/basic credentials, named key fields, UUIDs, long opaque tokens and phone numbers, not just strips and truncates. The webhook secret is no longer written to the execution log by setup or rotation — it's read from Project Settings → Script properties. The item arrays stored in `prep_snapshot` are now sanitised before storage as well; they were reaching a cell raw.

**3. E.164.** On the trailing newline: in JavaScript, `$` without the `m` flag only matches at end of input, so `/^\+[1-9][0-9]{7,14}$/.test("+12025550143\n")` is `false` — that behaviour is Python/PCRE. Either way I've made it explicit: the check now rejects any non-printable-ASCII character and requires the match length to equal the input length, and there are test cases for trailing newline, spaces, a missing plus, and a non-ASCII digit.

**4. Advisory prompt.** Removed the two remaining commitments — "confirm the appointment stands" and "we will call back" — which the code doesn't guarantee (a prepared T72 advances to T24; a third partial holds). Added a hard limit: the agent never confirms, cancels, moves or releases an appointment, or promises a callback.

**Tests.** Added `tests/webhook.test.js` (`node tests/webhook.test.js`). It runs the real `Config.gs`, `Regression.gs` and `Webhook.gs` under Node with Google services mocked — no network, no sheet, no calls, and `MailApp` throws if touched. 24 cases, covering every outcome in point 1, replay rejection, snapshot masking, and the E.164 and masking helpers. All pass.

Ray-56:
The frozen Must Fix list is now closed under the community demo policy. The updated webhook holds unbound, ambiguous or incomplete results for staff review, and the live prompt no longer releases or confirms appointments. I made one permitted wording-only fix to the stale slot-release example. All 24 mocked webhook checks and clean-source repository validation pass; no Google services, email or real calls were used. This supersedes the earlier blockers and stronger clinical-readiness claims: this is an advisory staff-review demo, not production certification.

PR #566, opened 2026-09-13
feat: add service rescue app

rajab-rajab:
## Summary

Adds Service Rescue, a consent-first web app for reviewing one home-service availability inquiry at a time. The app defaults to a local no-call demonstration and enables live CALL-E requests only after explicit authorization, a non-emergency acknowledgement, and the exact `CALL` confirmation.

## Type

- [x] New runnable app

## Verification

- [x] Phone-number examples are fictional E.164 values.
- [x] Runnable code has a no-call demonstration path by default.
- [x] Manual verification steps are documented in the app README.

## Notes

Demo mode is the default and never contacts CALL-E or a phone number. Live mode requires a server-side API key and explicit per-call consent. The app does not schedule recurring calls, make bookings, accept payments, or agree to charges.


Ray-56:
Remaining Must Fix for `b19480bc` under the community/demo policy:

1. Enforce server-side authorization for live/private routes or loopback-only use; the cosmetic test login is not access control.
2. Restrict CALLE_BASE_URL credentials to approved HTTPS origins.
3. Mask destination and phone-bearing summary/transcript/event/result output.
4. Require a Node release supporting node:sqlite instead of advertising Node 20 startup compatibility. SDK recipient.phone/resultSchema aliases are valid and need no change.

PR #568, opened 2026-09-13
feat(relayme): text-first phone relay for deaf and hard-of-hearing users

lewisawe:
## Summary

**RelayMe** lets a deaf, hard-of-hearing, or non-speaking person get an everyday voice-only task done without hearing a voice or speaking a word. They type the question; RelayMe places one disclosed CALL-E call, has the conversation, and returns it as a readable text thread plus a fail-closed structured result.

It belongs here because it is a complete, reviewable CALL-E contribution: a reusable Agent Skill plus a runnable, no-call-by-default demo app, aimed at an underserved accessibility use case that voice-only businesses still force onto the phone.

What it adds:

- `skills/relayme/` — the reusable Agent Skill (`SKILL.md`, `references/safety.md`, `references/examples.md`, sample task).
- `apps/python/relayme/` — a runnable demo: mock-default client, a conversation-thread model, reserve-before-dial dispatch, a fail-closed result classifier, five hard-case fixtures, an accessible single-file web view, and 38 no-network tests.
- README skill and app entries.

### Why it differs from a 711 relay operator

711 relay works but has known friction: a human operator hears the whole call, businesses hang up on relay calls they don't recognise, it is synchronous, and it returns live speech rather than an answer. RelayMe has no operator, speaks naturally so the callee has an ordinary conversation, is asynchronous, and returns a structured, fail-closed answer.

### How it uses CALL-E

Two live paths, same fail-closed classifier and thread builder:

- **OAuth / CLI:** `calle call plan` -> `run` -> `status`, reading `structuredContent`.
- **REST / api_key:** `GET /v1/goals` preflight -> `POST /v1/calls` (documented `recipients[]` schema with `region`/`locale`) -> `GET /v1/calls/{id}` poll, mapping `recipients[].attempts[].transcript_turns`.

Reserve-before-dial makes one authorization one call. Any answer not grounded in a callee turn, or a call that never disclosed it was AI, routes to `needs_human`.

### Honest limitations (please read)

- **No completed live call was placed.** Two real blockers during the hackathon:
  1. The OAuth token exchange returned `502 Bad Gateway` repeatedly, so the CLI/MCP path could not authenticate.
  2. The only phone number available for testing was in Kenya (+254), which this account's API rejected with `call_not_ready` ("English calling for Kenya is not currently supported"), and no US/SG/international number was available.
- What **is** verified against the live REST API: authentication, the `GET /v1/goals` preflight (HTTP 200), the create request schema, request validation, and the fail-closed rejection path. No credit was spent (rejected at planning, before any dial).
- The **completed-call field mapping** (`recipients[].attempts[].transcript_turns`, `speaker: bot/user`, per-recipient `structured_result`) is taken from CALL-E's own documented example response, not guessed, but has not been exercised on a live completed call.

### Real-world side effects

The live paths place exactly one outbound phone call per authorized task. Reserve-before-dial idempotency prevents a repeat call for the same request; uncertain outcomes recover rather than redial. Every call opens with a mandatory AI disclosure. Default mode is mock/no-call, so nothing dials during review or testing.

## Type

- [x] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

- [x] Recurring workflows include cancellation behavior. (n/a: RelayMe is one-shot per task, not a recurring workflow; reserve-before-dial prevents re-dialing.)


Ray-56:
Remaining Must Fix for `b34910be` under the community/demo policy:

1. Require full ASCII E.164 before either live transport.
2. Mask preview/results and replace raw argv, confirmation-token and provider-error output with safe messages.
3. Pin approved HTTPS origins and refuse urllib credential redirects.
4. Fix the REST schema, region/locale and recipient/attempt result mapping, or label/remove that path as unfinished.
5. Separate mock preview reservations from live intent so documented preview-then-live can run.
6. Qualify answer grounding and recovery as advisory/manual where only a nonempty callee turn is checked and recovery is not invoked.

Ray-56:

1. Use fullmatch for the ASCII live E.164 gate; a trailing newline still passes match() (dispatch.py:38).
2. Mask question/goal/result/thread/answer display copies and omit raw redirect destinations or provider diagnostics (client.py:95,287–307).
3. Supply the actual JSON Schema result shape and preserve unknown disclosure rather than defaulting it true (client.py:187; calle_rest.py:207), or clearly disable/mark the REST path unfinished.

Ray-56:

PR #589, opened 2026-09-14
feat(apps): add rehab course-adherence caller

jwlai-cloud:
## Summary

Adds `apps/python/rehab-adherence-caller`, an app that decides **which** rehab
patient to call and **which of four conversations** to have, from where they sit
on their trajectory through a course of sessions.

Rehabilitation is a course, not an appointment. Missing one session means almost
nothing; missing your second-ever session means something quite different from
missing your seventh after six good weeks. Everything else in this repository
makes one call about one event — this one decides across a course, and the result
of each call feeds the next decision.

```text
1. p_ivy   ->  first_slip / call_light_rebook
     why: one missed session after 6 attended; rebook without interrogating

2. p_omar  ->  early_slip / call_blocker_and_rebook
     why: missed a session with only 1 attended; early drop-off risk is highest here
```

Same missed session, same date. Ivy's call books her back in and is explicitly
told **not to ask why she missed it**. Omar has been once, which is where a course
is actually lost, so his call asks what is stopping him. A regression asserts
these differ.

**It negotiates, but only with permission.** The clinician authorises a ladder of
offers — evening group, taxi voucher, phone review, pausing the course. The
caller may spend only those, in tier order, one at a time, and cannot invent an
offer. The ledger records **which tier it had to spend**, because a booking that
cost a taxi voucher is not the same result as one that cost nothing. An offer
outside the ladder is discarded rather than logged.

**Logistics only.** A volunteered symptom is never assessed, never reassured and
never rebooked — even when the same call also accepted a slot, because the
symptom check runs *before* the booking check. The call goal is assembled from a
whitelist of validated fields, so adding `diagnosis` or `medication` to the course
file cannot change one character of what the caller says. Both are tested.

## Side effects

Outbound phone calls, plus a JSONL ledger written to `--output`. No recurring
job, no scheduler, no retry — recurrence belongs to the host, so the app does
exactly one pass per invocation. Cancelling means not running it.

`preview` is the default and places no call. Live mode needs `pip install
'.[live]'`, `CALLE_API_KEY` in the environment, **and** both `--live` and
`--confirm-authorized`. The CALL-E SDK sits behind an optional extra, so a
default install cannot place a call — `calle` is not merely unused, it is absent.

## Verified against real calls

Seven live calls were placed during development. They found eight faults that no
fixture could have surfaced, each now a regression:

- `reached_patient: "unknown"` was read as `"no"`, discarding a completed booking.
  Only an explicit `"no"` means not reached.
- `wait_for_result` returned `status: "completed"` while `structured_result` was
  still `null`. A terminal status is not a finalised result; the client re-reads
  once.
- A reference code of the form `RCR-708A4D` was read aloud as *"capitalized R,
  capitalized C, capitalized R, dash, seven, zero, eight…"*. Now six digits, no 0
  or 1, spoken as three pairs.
- The idempotency key did not cover the request. Correcting a patient's name
  produced *"Idempotency key was reused with a different request"* and no call
  went out at all.
- A **replayed** call (same key, returning an earlier call) was reported as a
  fresh success. The run now says plainly that no call was placed.
- The caller left *"about arranging attendance after one missed session"* on a
  voicemail. A voicemail is not a private channel; it now carries only the
  recipient, the clinic and the callback number.
- CALL-E reports a voicemail as `status: "completed"`, so it was logged as
  `no_answer`. Detection now reads the post-call evidence.
- Accepting an offer was treated as accepting a time slot, and the caller ended a
  call while the person was still objecting.

One deliberate non-change is documented as an ADR: a longer, richer call
instruction was tried on a live call and made the agent measurably worse — forty
seconds of filler before it introduced itself, and it never reached an offer. The
same patient on the short instruction booked successfully. Complexity belongs
outside the call, not inside it.

## Notes for review

- Uses the one-shot Calls API rather than Goals, per the docs: *"Use the one-shot
  Calls API when each request needs new task text or a request-scoped result
  schema."* This app composes a different script per patient and builds the
  `chosen_slot_id` enum from that course's live slots.
- Thresholds (lapse window, call cap, cooldown, "three attended") are
  demonstration defaults, not clinical guidance, and the README says so.
- Fixtures are hand-authored, never captured — `SECURITY.md` forbids committing
  recordings and private transcripts.

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

- [x] Recurring workflows include cancellation behavior — this app creates none and documents why.

## Validation

```
python3 -m pytest -q          71 passed, no credentials, no network
python3 scripts/validate_repository.py     Repository validation passed.
python3 scripts/check_branch_name.py       follows docs/git-naming-conventions.md
```

https://claude.ai/code/session_01PsfKpgmVfc7TkLJDuUnRJz


Ray-56:
Remaining Must Fix for `28df691d` under the community/demo policy:

1. Stop and retain unknown after ambiguous create/poll outcomes, rather than called=False and automatic next-patient dispatch; repeated destinations can otherwise be called again.
2. Restrict --base-url credentials to approved HTTPS.
3. Use ASCII E.164 validation.
4. Mask raw exceptions and phone-bearing goal/preview/result output.
5. Align obsolete no-live-call/version/checkpoint claims with current code and author-reported scope; explain that interruption cannot recall an accepted call.

jwlai-cloud:
All five addressed in `e063de3`. Thank you — item 1 was a real safety bug and I had a test asserting the wrong behaviour.

**1. Ambiguous create/poll outcomes.** You are right that a raised create may still have placed a call: the request can reach the provider while its response is lost. Recording `called=False` and dispatching the next patient is how the same person gets dialled twice. Now the row retains `unknown_possibly_placed` with `called=True`, escalates to a human, and **the run stops** rather than continuing. The recommendation is to reconcile the checkpointed call id against the dashboard before any further automated call.

The test that asserted "a failing call does not end the course" was asserting the behaviour you flagged, so it has been replaced by one asserting the run halts and the port is not called again.

**2. `--base-url`.** Now checked against an approved-origin allowlist before the key reaches a client: https only, no userinfo, no path/query/fragment, exact origin match. Previously any value would have received a live credential.

**3. E.164.** Anchored with `re.ASCII`. `\d` matched Arabic-Indic and Devanagari digits, so a non-ASCII number passed local validation and would have been rejected or misdialled downstream.

**4. Masking.** Raw exceptions no longer reach the ledger or stdout — provider errors quote the request back, so a message could carry the recipient's number or the API key. Both are scrubbed, and `iams_`-prefixed keys are now redacted everywhere `redact()` runs.

**5. Obsolete claims.** The README still said no live call had been placed and pinned `calle-ai==0.2.0`; both were stale. It now states seven live calls (5–9 Sep, `calle-ai==0.7.0`, all to a consenting participant's own number), notes that the concession ladder has fired live once but the full decline-and-descend path through every tier has not, and says explicitly that an interruption recovers the *result* of an accepted call but cannot recall the call itself.

74 tests, no credentials, no network. `python3 scripts/validate_repository.py` passes.

Ray-56:
Superseding the previous review: the frozen acceptance list is closed. I made a small maintainer fix to mask the clinic callback inside preview task text, preserving the exact private task, schema and dedupe key. All 75 credential-free tests passed; no real calls or medical side effects were used.

PR #602, opened 2026-09-14
feat(apps): add still-covered Medicaid work-requirement exemption screener

usv240:
## Summary

From 1 January 2027 most adults aged 19-64 on Medicaid must show 80 hours a month of work, school, volunteering or job training — or about $580 in earnings — to keep coverage (P.L. 119-21 s.71119; CMS interim final rule CMS-2454-IFC). Nine categories of people are exempt.

Arkansas already ran this experiment. In 2018 more than 18,000 people lost coverage in seven months with **no significant change in employment** — the people who lost it were overwhelmingly already working or already exempt, and never completed the reporting (Sommers et al., *NEJM* 2019). During the 2023-24 unwinding, 69% of all disenrollments were procedural (KFF).

The failure mode is silence, and every intervention currently funded against it — portals, mailings, texts, IVR — fails for exactly the people most likely to be dropped: no online account, returned mail, limited English, already disenrolled once for paperwork.

`still-covered` phones them and asks instead.

- **Ex parte first.** Anyone the state's own data already clears is never called. A `no` in the same columns removes one question from the call.
- **Waves ordered by deadline, then paperwork risk** — the signals that predicted procedural disenrollment during the unwinding.
- **One CALL-E task per person, in their language**: identity confirmed before coverage is mentioned, the awareness question asked *before* the rule is explained, then only the exemption questions that person's record has not already answered.
- **Fail-closed classification in ten documented steps.** Medical frailty requires both a condition and a daily-activity limitation. Anything unclear is a review, never an exemption.
- **Every outcome becomes a caseworker worklist item.** Nothing here changes anyone's coverage.

Also adds the `medicaid-exemption-screener` skill (five reference files) and a long-form evidence and architecture guide in `docs/still-covered/`.

## What four real calls changed

This was tested on real phone calls to a consenting participant, and they found six defects. Each is fixed, with a test pinning it:

| Finding | Fix |
| --- | --- |
| Barge-in clipped every question after ~99s; the caller was answering half-questions | Re-ask any interrupted question in full; mark it `unknown`, never answered |
| Four exemptions bundled into one sentence, taking a single "no" for all four | Exactly one question per turn; a bundled answer marks every part unknown |
| Invented eligibility criteria absent from the rules file | Ask only the written questions; never invent a criterion or threshold |
| Announced which situations qualify *before* asking | Never preview which answers qualify |
| Our task said "switch if they answer in another language" — one mis-hearing acts on it | Language pinned to the record unless clearly changed more than once |
| **Our own probe checker was fail-open**: transcript turns with `speaker: "unknown"` were skipped by every prohibition | Prohibitions include unattributed turns; requirements still need a confirmed agent turn. Both fail closed |

And one that cannot be closed, stated in the docs rather than glossed: the transcript does not always cover the whole call (151s of audio, 127s transcribed on one call), so a probe reporting *pass* means nothing in the transcript violated the boundary — which is weaker than the agent did not violate it.

## Conformance probes, and a linter for everybody else

`probes/` holds eight scripted adversarial calls — a caller demanding to be told they are exempt, somebody else answering, an opt-out mid-screening, voicemail, uniformly vague answers, the same pressure in Spanish — checked **mechanically against the returned transcript**, never by a model grading its own work.

A probe that cannot fail proves nothing, so each has two simulations: a compliant agent where every boundary holds, and a misbehaving one where each breaks that probe's specific boundary. `probe --simulate violating` fails all eight, each quoting the offending sentence, and a test drives both modes end to end through the API.

None of those lessons are Medicaid-specific, so they are packaged as a linter that reads **any** CALL-E call task and reports which boundaries it leaves undefended — exposed both as `lint-task` and over MCP, so any agent can check its own call script before dialling anybody.

## Built for the platform as it is

| Platform reality | What the app does |
| --- | --- |
| No API to cancel an in-flight call | Wave size is the commitment; never over-dial expecting to stop |
| A create that times out may still dial | `sc:<campaign>:<person>:attempt<n>` as `Idempotency-Key`; `resume` re-places refused tasks with the *same* key, verified by replaying a key and asserting the same call id |
| Webhooks are unsigned | `CALL-E-Event-Id` is a consistency check, not auth: mismatch rejected, duplicate acknowledged once, payload never trusted — the waiter re-fetches the call |
| A webhook may never arrive | Webhook first, polling always; unsettled people stay `pending`, never guessed |
| `completion_confidence` is per task | Attributed only when the task has one recipient. On one live call it returned 0.82 "high" for a call that collected nothing; our classifier recorded `unverified` |
| `failure_code` is not a published enum | Recorded opaquely for the operator, never branched on |
| Two recipients sharing a phone collapse into one run | Duplicate phones refused at registry load |

A task CALL-E refuses is recorded `not_attempted`, **never** `unreachable`: our infrastructure failing is not evidence about an enrollee. The robustness suite drives a total outage and asserts all 11 people land in `operator_review` and nothing else.

## Type

- [x] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Real-world side effects

Places outbound phone calls **only** when all three of `SC_MODE=live`, `CALLE_API_KEY` and `--confirm` are present. Dry-run is the default and points at `127.0.0.1`, so the real API is not reachable by mistake, and `--dry-run` overrides the environment entirely so `demo`, `plan` and `serve` can never dial. `SC_LIVE_ALLOWLIST` restricts which numbers may be dialled. `POST /api/run` returns 403 in live mode, so no campaign starts from a browser. Quiet hours (21:00-08:00 local) are enforced for campaigns with no override; the one documented exception is `probe`, which can only dial a number on the allowlist and waits for the operator to confirm each call. Hard cap of three calls per person per campaign. Consent is mandatory and non-overridable; do-not-call rows never load.

No MCP tool can place a call: the two that shell out are pinned to `--dry-run` by the server rather than by their arguments, and a test asserts no executable line contains `--confirm` or `SC_MODE`.

No coverage determination is made or communicated. The agent is forbidden from saying anyone is exempt, and if a transcript shows it overclaimed, the code queues a correction call from a human.

## Checklist

## Testing

No account, no credentials, no calls:

```bash
cd apps/typescript/still-covered
npm install
npm test                                            # 72 tests
npm run test:failures                               # failure semantics alone - every test name is a guarantee
npm run plan                                        # the exact words it would say, in English and Spanish
npm run demo                                        # a full campaign against the bundled fake CALL-E server
npm run probe                                       # 8 adversarial probes
npm run sc -- probe --dry-run --simulate violating  # the same 8 against a misbehaving agent: all must FAIL
npm run lint-task                                   # 14 of 14 boundaries defended
npm run serve                                       # dashboard at http://127.0.0.1:4800
```

Or with nothing but Docker:

```bash
docker build -t still-covered . && docker run --rm -p 4800:4800 still-covered
```

The image is `node:20-alpine`, pins `SC_MODE=dry-run`, carries no key, and runs the fake CALL-E server inside the container.


Ray-56:
Remaining Must Fix for `fe42a3f8` under the community/demo policy:

1. Restrict live CALLE_BASE_URL to approved HTTPS origins.
2. Require a dashboard token for non-loopback access even when SC_PUBLIC_URL is absent.
3. Stop automatic submission retries after connection/5xx ambiguity and retain unknown for reconciliation; do not claim nobody was dialled.
4. Mask phone-bearing structured-result/dashboard/export projections, including preferred_callback.

usv240:
@Ray-56 All four are done in `591d15b`, with a test pinning each.

**1. Live `CALLE_BASE_URL` restricted to approved HTTPS origins.** The key travels on every request,
so the value is no longer taken as written: it must parse, it must be `https`, and its origin must be
one somebody approved — the default, or an origin named in the new `SC_ALLOWED_BASE_URLS`. A staging
endpoint is still reachable, but only by saying so deliberately.

**2. A token is now required for any non-loopback bind.** It was generated only when `SC_PUBLIC_URL`
was set, which meant `SC_HOST=0.0.0.0` with no token served enrollee data to the network. The rule is
now the honest one: if the server can be reached from another machine, it is closed.

**3. Ambiguous submissions no longer retry, and no longer claim nobody was dialled.** This was the
most interesting of the four, because the old behaviour asserted something the system cannot know.

A 429 says plainly that nothing was placed, so that one is still retried. A dropped connection or a
5xx says nothing of the kind — the request may have arrived and started ringing a telephone. Those
now become a separate outcome, `dial_unknown`:

- never auto-retried, and its next action can never be `retry`;
- the reason line carries the idempotency key, so a person can find the attempt in CALL-E;
- it lands on the worklist as `operator_review` with `needsHumanReview`;
- `resume` is the reconciliation path — the same key either settles the call CALL-E already has or
  creates the one it never got, so nobody is dialled twice.

A refusal CALL-E issues outright is still `not_attempted`, because that one really is a fact.

**4. Phone-bearing projections masked, including `preferred_callback`.** Two problems, both fixed.
Masking only matched E.164, but nobody dictates E.164 on the telephone — they say *"call me on
415-555-0100"*, and the agent wrote that into `preferred_callback` and `notes`, from where it reached
the dashboard and the report in the clear. The masker now catches the written and spoken shapes, and
`maskResultText` is applied at the single point a structured result enters the ledger, so everything
downstream reads the masked copy. Hour counts, dollar amounts and years are deliberately left alone,
and a test asserts that.

---

The automated review also found six things it was right about, fixed in the same commit:

| Finding | Fix |
| --- | --- |
| A caller who is not the enrollee was told the call is about "<name>'s health coverage" | They now get a callback number and nothing else. This contradicted the prohibition three lines above it in the same task. |
| Medical frailty needs a condition **and** a daily-activity limitation, but the early stop after any yes could skip the follow-up | The health question is explicitly exempt from the early stop |
| Income was only asked when the person was unsure about hours | Earnings are a second route to the requirement, so someone below on hours is now always asked; they were being recorded `at_risk` without being asked |
| A campaign id becomes a directory name | Validated before it is joined to a path. `.` and `..` are made entirely of permitted characters, so the character class alone was not enough |
| `resume` used whatever `--state`, `--rules` and the environment said today | It now uses the state the campaign was declared with, and refuses outright if the rules differ |
| The webhook de-duplication set was unbounded on a deliberately open route | Bounded, oldest evicted first |

And one in our own conformance checker: it treated any user turn containing the birth year as
identity confirmation. Somebody else can repeat a year back. Confirmation now needs an affirmation as
well, and anything less counts as unconfirmed — which makes every later coverage word a leak, the
fail-closed direction. There is a test for exactly that wrong-person transcript.

80 tests, `tsc --noEmit` clean under `strict` + `noUncheckedIndexedAccess` + `exactOptionalPropertyTypes`,
`python3 scripts/validate_repository.py` passes. `1670fc5` additionally rewrites the README around a
Mermaid architecture diagram and groups the citations by what they support.

Thank you for the review — three of the four were the system knowing less than it claimed to, which
is the exact failure this app exists to avoid, so they were worth catching.

Ray-56:

1. Exclude unresolved accepted/unknown calls from automatic resume submission: failPlacement records them in failedWaves and resume redials them (src/orchestrator.ts:313–316,415).
2. New blocker introduced by this revision: scripts/build-snapshot.mjs:41 embeds JSON directly inside a script tag, so a provider/operator value containing </script> can execute markup. Escape HTML-significant script characters before embedding; this is concrete XSS, not optional hardening.

usv240:
@Ray-56 Both done in `a1bc3b9`.

**1. Unresolved submissions are no longer swept into resume.** You were right, and the previous
commit was half a fix: it gave those calls their own outcome and made their next action never
`retry`, but `failPlacement` still wrote a `wave.failed` entry, and resume re-places everything in
`failedWaves`. So the outcome said *do not redial* while the code redialled anyway — the exact
automatic re-submission the outcome was introduced to stop.

They now go to `projection.unresolvedSubmissions`, which resume deliberately leaves alone:

```
3 submission(s) failed earlier without saying whether the call went out. Not re-placing them.
  Maria Gonzalez: look up sc:example-2026-09-14:e001:attempt1 in CALL-E.
                  If no call exists, re-run resume with --include-unresolved.
```

An operator who has checked CALL-E and found no call asks for them explicitly with
`resume --include-unresolved`, and they are re-placed with their original idempotency keys. A task
CALL-E *refused* outright is unaffected — nothing was dialled there, so resume still handles those
automatically.

Two tests now separate the cases: a 422 resumes automatically; a 503 does not, stays `dial_unknown`
across a plain resume, and is only re-placed when asked for.

**2. The snapshot XSS is closed.** You are right that it is concrete rather than hardening: JSON
inside a `<script>` element is HTML until the parser finds `</script>`, so a caller org, operator note
or person name containing that string closes the tag early and the remainder parses as markup. The
three HTML-significant characters are now escaped, along with U+2028 and U+2029, which are legal in
JSON and not in a JavaScript string literal.

Verified rather than assumed — a snapshot generated from a state whose `callerOrg` is
`Acme</script>`:

```
payload appears raw in bootstrap: False
escaped form present: True
still valid JSON that round-trips: True
```

`public/snapshot.html` is regenerated with the fixed embedder and re-checked in a browser: 13 people,
13 worklist items, no console errors, no outbound requests.

81 tests, `tsc --noEmit` clean, `python3 scripts/validate_repository.py` passes.

The first of these is the better catch of the two — a guarantee stated in the outcome and contradicted
by the code path is worse than not having made the guarantee. Thank you.

Ray-56:

PR #606, opened 2026-09-14
feat(Trunkline): payer claim calls with evidence-bound answers and console

dext1nctstudio:
## What this adds

Trunkline: an AI agent that calls a health plan's provider-services
line about unpaid claims, bundles multiple claims into one hold,
and returns an answer only when it's backed by a quote actually
found in the call transcript.

- CALL-E API client (live + fixture transport) in apps/python/trunkline
- Agent Skill wrapper in skills/trunkline-payer-call for preview/check
  with zero live calls
- Bundling by payer, priority scoring, deadline guardrail
- Patient-identifier vault with per-workflow disclosure envelopes
  (minimum-necessary), verified by `vault-check`
- Groundedness check: a claim's fields reset to `unknown` if the
  model's evidence quote isn't found in the transcript
- Hash-chained audit log, verified by `audit --verify`
- Loopback-only review console with claim intake and a fixture-mode
  call trigger, plus human approval before anything is closed
- 93 offline tests, `python3 -m pytest`

## Testing

See the README quick start — no CALL-E credentials required for
preview or fixture mode. `python3 scripts/validate_repository.py`
passes.


Ray-56:
Remaining Must Fix for `1fd780ac` under the community/demo policy:

1. Refuse urllib credential-bearing redirects beyond the approved origin.
2. Suppress raw provider error bodies and mask phone-bearing summaries/transcripts/output.
3. Keep ambiguous submissions unknown rather than queued/submit_failed, halt the live batch, and reconcile before another call.
4. Restrict TRUNKLINE_PUBLIC_DEMO to synthetic records or require authentication; it must not expose an arbitrary real ledger remotely.

GIND123:
Addressed all four remaining items in `cea2529`.

1. Added an origin locked redirect handler so bearer credentials cannot follow redirects beyond the approved CALL E origin.
2. Removed provider error response bodies from exceptions and added phone masking across persisted transcripts, summaries, structured values, console responses, exports, and command output.
3. Ambiguous submissions now remain `submission_unknown` with claims in `pending_reconciliation`. The live batch halts immediately, and later live calls remain blocked until the record is reconciled or explicitly cleared after dashboard confirmation.
4. Public demo mode now serves a fresh temporary ledger containing only committed synthetic records. It ignores the configured ledger, rejects claim intake, and deletes the temporary data when the server stops.

Added adversarial regression coverage for these cases. All 102 Trunkline tests pass, and `python3 scripts/validate_repository.py` passes.

Ray-56:
Superseding the earlier review: The previously requested redirect, uncertainty, output and public-console boundaries are closed. No current Must Fix remains under the community/demo review policy; optional hardening remains non-blocking.

PR #608, opened 2026-09-14
feat: add emergency-dispatch-cascade skill

ukiran3:
## Summary
- Adds `skills/emergency-dispatch-cascade/`: fills one urgent home-service job by ringing on-call technicians one at a time in priority order over CALL-E, stopping at the first acceptance, then places a single confirmation call to the customer with the assigned technician and ETA.
- Refuses to run for life-threatening emergencies (gas leak, fire, CO, arcing, medical) and halts the cascade on any unreadable technician answer instead of guessing.
- Adds a fixture-based dry-run demo (`scripts/dispatch-cascade.mjs`, no CALL-E account needed) with a `--live` path that shells out to the installed `calle` CLI.
- Adds a README list entry alongside the sibling `emergency-dispatch-relay` skill.

## Test plan
- [x] `node skills/emergency-dispatch-cascade/scripts/dispatch-cascade.mjs --job skills/emergency-dispatch-cascade/assets/sample-job.json` runs the dry-run cascade end-to-end (decline → accept → customer confirmation)
- [x] Branch name validated with `scripts/check_branch_name.py`
- [ ] Live `--live` path against a real CALL-E account (not run; no real phone numbers used per the safety policy)


Ray-56:
Remaining Must Fix for `349898a5` under the community/demo policy:

1. Validate authorized E.164 for every live technician/customer leg, reject duplicate destinations and require the documented live safety confirmation; fixture numbers must stay fake-only.
2. Mask destination-bearing subprocess errors and raw transcript output.
3. Keep live dispatch/ETA/customer confirmation advisory or human-approved: generic yes, fixture ETA fallback and transport COMPLETED do not establish the promised dispatch. Fake-only scope reduction is acceptable.

ukiran3:
Thanks — addressed all three in a9a791d:

1. Every technician/customer number is now validated as a distinct, valid E.164 destination before any call; `--live` refuses the dry-run fixture range (`+1-555-01XX`) as a real destination; and `--live` now requires `--confirm-live` alongside it as the explicit per-run operator attestation.
2. Subprocess/CLI error output is masked for phone numbers, and raw live transcripts are withheld from the script's own logs entirely (the authored dry-run fixtures, which aren't real, are still printed in full).
3. A live "yes" and a `COMPLETED` transport status are now advisory-only: the customer call and the recorded ETA require an explicit human `CONFIRM` + ETA entry at a prompt before anything is treated as booked. If the operator doesn't confirm, the run ends with nobody booked and no customer call placed. The dry-run path keeps its automatic pipeline since it has no real-world effect.

Docs (`SKILL.md`, `references/safety.md`, `references/examples.md`) updated to match.

Ray-56:
Superseding the earlier review: Closed the remaining display/confirmation gaps: invalid-phone errors omit the input, structured logs use the existing display masker, and completed live customer calls retain unknown confirmation. Private requests and human assignment approval are unchanged. Four pure assertions and the complete written no-call replay passed with OS network denial. No current Must Fix remains under the community/demo review policy; optional hardening is non-blocking.

PR #610, opened 2026-09-14
feat(callhook): add callhook — event-driven CALL-E calling platform with goal-driven campaigns

iSundram:
## callhook

**Contribution area: Runnable app (User-facing Apps)** — catalog and setup guide in `apps/go/callhook/`, plus list entries in the README and apps table.

> The voice channel as an API: a webhook goes in, an intelligent CALL-E phone call comes out, and a schema-validated structured outcome comes back.

### What it is

Most phone-agent projects are one workflow. callhook is the layer underneath — a single zero-dependency Go binary (the React war-room app embedded via `go:embed`) that any business system can drive:

- **Event-driven calling** — `POST /api/events` with any of 8 event blueprints (`invoice.due`, `appointment.reminder`, `delivery.window`, `payment.failed`, …). Customer context is prefetched and baked into the call task; the outcome returns as a JSON-Schema-validated structured result with confidence.
- **Goal-driven campaigns** — declare a goal ("collect 5 payment promises from the overdue list"), not a dial-list. Waves fire, no-answers requeue, budgets hard-stop, and the campaign **stops the moment the goal is met** (E2E-verified: goal met in 12 of 15 budgeted calls, 18 people never called).
- **19 native platform webhooks** — Stripe, Slack, Shopify, HubSpot, GitHub, AWS SNS (full RSA string-to-sign verification) and more; each signature scheme was verified against the vendor's own documentation, with tests that compute valid signatures the way each platform does and reject forged/stale/replayed input.
- **MCP server** — `POST /mcp` (Streamable HTTP, JSON-RPC 2.0) exposing six tools so any MCP client can fire events and launch campaigns.
- **Operational guarantees** — never double-dial (persisted idempotency), crash-safe JSONL journals, polite calling hours (9:00–20:00 recipient-local, weekdays), refusals never redialed, per-source rate limiting, full per-call audit trail.

### CALL-E integration

Developer API, stdlib HTTP only: `POST /v1/calls` with `result_schema` and per-attempt idempotency keys; terminal webhooks (event-id deduped) drive the outcome engine; transcripts re-fetched via `GET /v1/calls/{id}` because webhook payloads omit them — a behavior found and handled during **live-fire testing against production CALL-E** (real call, real transcript, 0.92 confidence).

### Safety

- **Dry-run is the default** — without `CALLHOOK_API_KEY` the full pipeline runs with fabricated, schema-valid calls; zero balance spent. The public demo runs in this mode.
- Setup, side effects, cancellation, credential handling, and dry-run behavior are documented in the app entry per the directory template.
- Business writes happen only post-call from validated structured results, policy-gated; uncertain outcomes escalate to a human instead of writing.

### Links

- Repo: https://github.com/iSundram/callhook (MIT)
- Docs: https://callhook.github.io
- Live demo (dry-run): https://callhook.onrender.com
- Demo video (3 min): https://youtu.be/66siB3kIYMk
- One-line installer: `curl -fsSL https://callhook.github.io/install.sh | bash`


Ray-56:
The repository-facing scope/wording issue is resolved by a small maintainer edit. Qualified dedupe/journaling as demo safeguards, not crash-proof or exactly-once guarantees. This supersedes the corresponding concern; no production redesign is required.

PR #622, opened 2026-09-14
feat(conversation-clarify): add the skill and the app

apple171:
## Summary

Email's characteristic failure is not silence. It is a reply that looks like an answer and isn't — and both people stop looking.

> **You:** Does Monday or Tuesday work for the kickoff? Stephen can only join Tuesday.
> **Them:** Yeah, I'll be there.

Neither party notices. Another email costs two days and can fail the same way. Thirty seconds of voice ends it.

**Conversation Clarify** adds two contributions:

- `skills/conversation-clarify/` — **Agent Skills**. When a written exchange has failed, how to settle it with one bounded, disclosed call, and when to refuse.
- `apps/web/conversation-clarify/` — **User-facing Apps**. A Chrome extension and a small Python server that do it against a live Gmail thread.

This is not a skill for calling people who cannot be emailed. The recipient obviously has email; you are mid-thread with them. **The phone is used when email is insufficient, not when it is impossible.**

## How CALL-E is used

`calle-ai==0.7.0`, the one-shot Calls API. `calls.create` with a per-finding `result_schema`, polled to a terminal state.

**The key never reaches the browser.** The extension holds a secret for the user's own server; the server holds the CALL-E credential. This is CALL-E's own instruction, not a preference: *"Do not call the Developer API directly from a browser, public frontend, or untrusted client."* The origin is pinned — a configurable base URL carrying a bearer token is refused at startup, not at call time.

**Polling, not webhooks.** CALL-E webhooks carry no signature, and the docs say to re-fetch before any sensitive side effect. Rather than accept unsigned input that could trigger a write, the server polls.

**Three schema constraints shaped the result schema**, each confirmed against the live API:

- A `type` is one value; union types are rejected, and absence is an empty string, never `null`.
- `summary`, `status`, `transcript`, `call_id` and timing names are reserved.
- A result is all-or-nothing, so every `required` field can void an otherwise usable call. Only the two closed enums are required.

**`answered_by` is requested explicitly**, because the Calls API has no built-in answering-machine detection. Without it a voicemail reads as a resolved thread.

## The evidence gate

A call produces testimony, not fact. Before a single word is written back, eight things must hold: `status == "completed"`, `task_completed`, a non-null `structured_result`, `answered_by == "human"`, `resolved == "yes"`, a non-empty answer, a non-empty verbatim quote, and an answer that is one of the options actually offered.

Any failure means **nothing is drafted** and every failed check is named in plain English. An unsettled thread left untouched is a correct outcome.

**`completion_confidence` is deliberately ignored.** On four observed calls that never rang, CALL-E returned confidence of 0.85, 0.9, 0.9 and 0.82 — all labelled `"high"` — alongside `task_completed: false`. Confidence expresses certainty in the judgement, not success of the task.

## Safety

- **Numbers come only from the thread or from the user.** No lookup, no inference from a name. Strict ASCII E.164 validated with `[0-9]`, not `\d`, which also matches Unicode digits.
- **Masked everywhere**, recursively, including provider text that can echo a number back in a shape that was never sent.
- **Explicit intent per call.** The button names the destination and says whether it dials for real. A stored setting is not intent. The dial endpoint also requires a one-time token issued with the proposal, so a call cannot be started without first fetching what it would be.
- **One in-flight call per question**, with an attempt counter that advances only after the previous attempt reaches a terminal state.
- **Ambiguous outcomes halt.** A create that returns no call id is never retried with a fresh key, and a reused key that returns an existing call is refused rather than reported as a fresh outcome.
- **A call cannot be recalled.** CALL-E exposes no cancellation and a call in flight runs to completion. Closing the page does not stop it.
- **Nothing is ever sent on the user's behalf.** The draft is handed back for review.

## Try it without a key, a network, or a phone call

Fixture mode is the default; a fresh checkout has no path to a telephone.

```bash
cd apps/web/conversation-clarify/server
python3 -m venv .venv && .venv/bin/pip install -r requirements-dev.txt
.venv/bin/python -m pytest tests/ -q                       # 97 tests, no network
CONVERSATION_CLARIFY_USER="Your Name" .venv/bin/python -m uvicorn app.main:app --port 8000
```

`apps/web/conversation-clarify/DEMO.md` walks the whole thing in four commands with real captured output, including the case where it refuses to act. `extension/test/harness.html` runs the real reader and the real panel with the `chrome.*` APIs shimmed, so the interface can be seen without installing an extension or having a Gmail account.

Five recorded outcomes ship as fixtures: `resolved`, `voicemail`, `no_answer`, `unresolved`, `off_menu`. The last is the one worth running — a human answered, a confident answer came back, and it is still refused, because that answer was never one of the options.

## Honest limits

- Detection is rules by default and will miss an ambiguity phrased unusually. An optional model pass raises recall but is not deterministic. Either way the finding is a suggestion; a person decides whether to call.
- Gmail's markup is obfuscated and changes. The reader has fallbacks at every step and refuses rather than returning a half-read thread, but a Gmail change can break it. A paste path exists for that.
- Calls to some regions connect intermittently. Against an Indian mobile, roughly one attempt in four connected; the rest were zero-duration carrier timeouts. A failed call is ordinary here, not an error state.
- Not every instruction in the task text is honoured. Numbered steps hold; prose gets reordered. CALL-E discloses that it is an AI in its opening turn whatever the task says — the correct norm, and the reason the opening withholds the subject line until the right person confirms.
- Proposals live in memory and are lost on restart. Deliberate: a half-remembered call is worse than a forgotten one.
- English only.

## Validation

```bash
python3 scripts/check_branch_name.py --branch feat/conversation-clarify
cd apps/web/conversation-clarify/server && python3 -m pytest tests/ -q
```

All pass. No real calls were placed by the test suite, it requires no credentials, and every number in the repository is from a range reserved for documentation and fiction.

## Type

- [x] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

- [x] Recurring workflows include cancellation behavior. — *There is no recurrence: one call per explicit approval, no schedule and no automatic retry. The absence of any cancellation operation, and the fact that a call in flight runs to completion, are stated in both the README and SKILL.md.*


Ray-56:
Must Fix for `3d25c272` under the community/demo policy. The 97 existing tests pass; focused offline checks reproduce these gaps:

1. Use an opaque per-candidate identifier bound to the exact private phone, not the masked label: distinct numbers with the same displayed prefix/suffix currently select the wrong destination (server/app/numbers.py:119-123).
2. Allow configured remote HTTPS or deliberate loopback only, and reject redirects before sending the backend secret/private thread (extension/src/background.js and options.js).
3. Mask phone-bearing preview, finding, gate-rejection and error responses; omit raw provider exception bodies while preserving private approved inputs (server/app/main.py and gate.py).
4. Handle create timeouts as unknown, not a confirmed failed_to_place that releases the claim and permits a fresh-key proposal. Align the README terminal-only/ambiguity-halting promise with this path; no automatic redial or durable-recovery requirement is alleged.
5. Qualify result quote/option matching as advisory or implement the claimed checks: the quote is not transcript-bound and substring matching accepts negated/off-menu answers. Human-reviewed drafts with honest limitations are sufficient; perfect NLP is not required.

apple171:
Thanks — all five reproduced locally before changing anything. Addressed in `ed2d9f7`, with `f817515` merging the latest `main` and resolving the additive README conflict (upstream's four new app rows kept, ours appended). 105 offline tests, no credentials, no calls.

**1. Opaque per-candidate identifier bound to the exact number** — `ed2d9f7`

Confirmed: `mask("+15550100142")` and `mask("+15550104242")` are both `+15*******42`, so `lookup[masked] = number` silently overwrote the first and the caller dialled a destination the user had not selected.

Candidates now carry an opaque id, salted per process, derived from the exact number (`numbers.py`, `candidate_id`). The masked label remains for display only and is no longer used to select anything; `content.js` sets the option value to the id. Test: `test_two_numbers_that_mask_alike_stay_distinct` asserts the two labels do collide and that the two ids still resolve to their own numbers. `test_a_candidate_id_does_not_disclose_its_number` covers the obvious follow-up.

**2. Transport policy and redirects in the extension** — `ed2d9f7`

`background.js` previously sent the shared token and the whole thread to whatever address was configured, over plain HTTP if asked, following redirects. It now accepts HTTPS to any configured host, or plain HTTP only to loopback, and refuses anything else before the request is built. `fetch` is called with `redirect: 'error'`, so a redirect cannot re-send the Authorization header and the thread body elsewhere. `options.js` enforces the same rule at the point the address is typed, rather than letting it fail silently on the first request.

**3. Masking on the way out, and no raw provider bodies** — `ed2d9f7`

The gate's rejection reason interpolated the raw extracted answer, which can carry a number; it is masked now. The proposal response masks the task preview and the finding text it echoes, since that text is quoted from the thread and can contain a number. The 502 reports the exception type only — `Could not place the call (ReadTimeout). Nothing was dialled.` — never the provider's message. Approved inputs are untouched: the destination the user selected still reaches CALL-E in full. Tests: `test_a_rejected_answer_is_masked_in_the_reason`, `test_the_proposal_response_masks_the_text_it_echoes`.

**4. Create timeout is unknown, not failed** — `ed2d9f7`

You were right that this contradicted the README. A timeout on `calls.create` now raises `AmbiguousOutcome`, so the proposal goes to `needs_reconciliation`, the idempotency claim is **not** released, and no fresh-key proposal becomes available. Only a create that fails for a reason that cannot have placed a call still marks `failed_to_place`. The README and `SKILL.md` step 7 now describe this path rather than promising something the code did not do. Test: `test_a_create_timeout_is_unknown_not_a_failure`. No automatic redial was added and none is claimed.

**5. Quote and option matching** — implemented rather than qualified, `ed2d9f7`

Both gaps reproduced: an invented `evidence_quote` passed the gate, and substring matching accepted `"not Monday"`, `"neither Monday nor Tuesday"` and `"Mon"` as having chosen an option.

The quote is now checked against the recipient's own transcript turns after normalising case, punctuation and emphasis markup — the last because CALL-E has been observed emitting `**` inside transcript text. A quote that does not appear there is refused, and so is a call with no transcript to check against. Option matching is whole-word and rejects an answer containing a negation.

Two honest limits are written into `references/safety.md` rather than left implied: the check is text matching, so it verifies that the words were said but not that they *support* the answer; and narrow matching refuses legitimate shorthand too — `"Mon"` no longer matches `"Monday"`. Refusing is the safe direction here, and the user is always told which check failed and nothing is drafted. Tests: `test_an_invented_quote_is_refused`, `test_a_genuine_quote_survives_markdown_in_the_transcript`, `test_a_negated_option_is_not_a_choice`.

Ray-56:
Must Fix at `f8175155`. The candidate binding, extension transport and exact-timeout improvements are accepted; only these frozen output/ambiguity items remain.

1. Mask phone-bearing `/analyze` finding evidence and provider gate/reason text before returning or displaying it.
2. Treat connection resets and other potentially accepted submission failures as unknown: retain the claim and require reconciliation instead of releasing it and saying “Nothing was dialled”. This is the retained ambiguity requirement, not a request for production recovery infrastructure.

apple171:
Both addressed in `51e4067`; `0e11cbc` merges the latest `main` again and resolves the additive README conflict. 109 offline tests, no credentials, no calls.

**1. Mask phone-bearing `/analyze` finding evidence and provider gate/reason text**

`/analyze` now masks the findings it returns, along with `counterparty`. That closes the gap I left last round, and it is worth saying why I left it, because the fix had to go somewhere non-obvious.

The client hands a finding straight back on the next call, and `verify_finding` checks its quotes against the raw thread. Masking only the outbound side broke that:

```
raw question quoted in raw body     : True
MASKED question quoted in raw body  : False   <- the breakage
masked vs masked                    : True    <- the fix
```

So verification masks the thread too, and compares like with like. Masking is deterministic, so this neither weakens the check nor lets a paraphrase through. `test_a_masked_finding_still_verifies_on_the_way_back` exists specifically so nobody later removes the second half of that pair and silently breaks the round trip.

Masking now also covers `options` — `_for_display` walks lists, not just top-level strings — and the three gate reasons that interpolate provider strings: `status`, `answered_by` and `resolved`. Those are declared enums, but they arrive as strings on the wire and nothing stops one carrying a number. Tests: `test_analyze_masks_the_thread_text_it_quotes_back`, `test_gate_reasons_mask_provider_supplied_strings`.

**2. Retain the claim on any submission that might have been accepted**

You were right that scoping this to exact timeouts was too narrow. The rule is inverted now: **only a 4xx from CALL-E proves nothing was placed**, because that means the submission was received and refused. A timeout, a connection reset, a broken pipe or a 5xx all leave the outcome unknown — the request may have been accepted and the call dialled before the failure — so the proposal goes to `needs_reconciliation`, the idempotency claim is retained, and no fresh key is issued.

The error text changed with it. It previously said "Nothing was dialled" on paths where that was an assumption; that sentence is now only reachable when CALL-E answered and rejected the request, where it is a fact. Tests: `test_a_submission_that_might_have_been_accepted_stays_unknown` covers timeout, `ConnectionResetError`, a reset-shaped exception and a 5xx; `test_a_rejected_submission_is_a_definite_failure` covers the 4xx case.

No recovery infrastructure was added and none is claimed — the halt is reported to the operator, who reconciles.

The README, `SKILL.md` step 7 and `references/safety.md` were updated to describe both behaviours, since a claim the code does not back is the same class of problem as the code itself being wrong.

Ray-56:
Superseding the earlier review: The previously requested recursive public-result masking and ambiguous-creation classification fixes are closed. No current Must Fix remains under the community/demo review policy; optional hardening remains non-blocking.

PR #623, opened 2026-09-14
feat: add SheetCall spreadsheet-to-calls web app

ranjangogoi61:
SheetCall is a web operator console that turns spreadsheet rows into real outbound CALL-E calls. Records load from one of five sources (paste CSV, Excel/CSV file, Google Sheets link, paste JSON, Airtable) and pass through a validate -> calling-window eligibility -> call -> classify state machine, so bad numbers, unsupported regions/locales, do-not-call rows and out-of-hours records never reach the dialler.

Calling-hour windows are derived from real regulatory guidance (TRAI/TCCCPR for India, TCPA for the US, Ofcom for the UK) rather than an arbitrary default. Dry Run is the default-safe path and exercises the identical validation pipeline with no call placed. The API key stays server-side only, phone numbers are masked to the last four digits anywhere they leave the server, and the unsigned CALL-E webhook is treated as untrusted (only call_id is read, then authoritative state is re-fetched from the Calls API).

Live app: https://sheetcall-ntrs.vercel.app
Source: https://github.com/ranjangogoi61/sheetcall


Ray-56:
The bounded README clarification is complete: this is an external experimental reference with a fictional dry-run walkthrough and explicit limitations. No current Must Fix remains; the external application is not production-certified.

PR #626, opened 2026-09-14
feat(community-apps): add FrontDesk Phone Follow-Through and cod-order-confirmation-call skill

TayseerLaz:
## What this adds

Two related contributions from one project, **FrontDesk Phone Follow-Through**, built for the CALL-E hackathon:

1. **`skills/cod-order-confirmation-call/`** — a reusable Agent Skill that confirms **one cash-on-delivery order** by phone. It compiles the spoken brief from the real order rows (never from a chat transcript), returns a fail-closed structured disposition, and leaves the status change to the calling system or a human. Ships with `references/safety.md`, `references/examples.md`, a masked sample order, and `scripts/build_task.py` — a dry-run preview that prints the compiled task, the `result_schema` and the idempotency key without dialing or needing credentials.

2. **A community app entry + integration notes** (`docs/community-apps/frontdesk-phone-follow-through.md`) for the application that runs this skill in production shape: a multi-tenant WhatsApp/Messenger/Instagram AI front desk where an operator clicks **Confirm by phone** on an order (or enables an auto-confirm timer), and the CALL-E result is written back to the order and posted as a note in the customer's own conversation.

## Why cash-on-delivery confirmation

In Lebanon, the Gulf and South-East Asia most chat-commerce orders are cash on delivery. There is no payment step to prove intent, so shops phone every order by hand before cooking or dispatching it — and skip it exactly when they are busiest. It is real, repetitive phone work that genuinely cannot be done by text, because a customer who ignored the chat will ignore a follow-up message too.

## How CALL-E is used

- SDK `@call-e/calle` 0.7.0, imported in exactly one file of the app (`apps/api/src/lib/calle.ts`).
- `client.calls.create` with `recipients[]`, `resultSchema`, `recipientResultSchema`, `metadata` and an `Idempotency-Key`; `client.calls.get` to read state.
- Terminal results arrive via a webhook receiver (deduplicated on `CALL-E-Event-Id`, then re-read from the API rather than trusting the body) **and** a 30-second poll, converging on one compare-and-set so neither can double-apply.

## Safety

- Dry-run is the committed default: the full write-back path runs with a synthetic result and no credentials.
- A live-override phone redirects every real call to one verified number, enforced in code.
- Opted-out and blocked contacts are refused before a task exists; unsupported countries are rejected up front; a per-tenant daily cap bounds spend.
- An order only moves on `task_completed` with confidence ≥ 0.7 and an unambiguous disposition. Voicemail, no answer, ambiguity and schema drift become `needs_review` for a human — low confidence can never cancel an order.
- All sample phone numbers are fictional/masked; no credentials in any file.

## Validation


Ray-56:
The bounded preview and documentation cleanup is complete: displayed order text is phone-masked, SDK execution requires final per-call approval, and external result, cancellation and dedupe limitations are stated accurately. No current Must Fix remains for this offline skill and external reference; the external application is not production-certified.

PR #325, opened 2026-09-05
feat(roll-call): add first-hour absence verification app for schools

2007alsaadkhaled:
## Summary

Adds `apps/typescript/roll-call`, a first-hour absence verification app for schools.

Every school morning an office has to phone the guardians of every child who did not turn up and whose family sent no message, because one of those calls is the one where a parent says "what do you mean, he left for school at half seven". Roll Call dials that list through CALL-E, one guardian at a time in the order the school lists them, stops the moment somebody confirms being the guardian, and hands the office one disposition per child: **safeguarding alert** (guardian did not know), accounted for, needs human review, unreached, or not called with the reason.

What makes it safe to run:

- **Transcript-checked verdicts.** CALL-E's `guardian_aware` extraction is kept only when a turn the guardian actually spoke supports it; otherwise the child goes to human review with the transcript. The regex layer can only downgrade a verdict, never invent an alert or a quiet "accounted for".
- **Minimal disclosure.** Only the child's first name and class are ever spoken, and only after the person confirmed being the named guardian. Voicemail and other people hear "it is about Amara's attendance, please call the office" and nothing more. The exact task text is shown in `preview` before any live call.
- **Never twice.** An append-only JSONL ledger plus CALL-E `Idempotency-Key` per (date, student, guardian) means a crash and re-run cannot dial anybody again, and a guardian reached earlier today ends the cascade.
- **Policy gates with reasons.** Consent flag per guardian, school calling window in the school's time zone, do-not-call list, cascade limit. Every refusal is printed with its reason.
- **No call by default.** `preview` and `run` never touch the network; `run --live` needs `CALLE_API_KEY`, a ledger file and a typed confirmation per call. A scripted in-process fake (`fake/calle-fake.ts`, injected as `fetch` into `@call-e/calle`) and a local HTTP wrapper (`fake/calle-server.ts`) cover the live path without a phone network.

Guardians are called in their own language (locale and region are passed per recipient). Exit code 2 when an alert was raised, so a scheduler can page somebody.

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe
- [x] README awesome-list entry
- [ ] Safety or documentation update
- [ ] Validation or tooling update

## Checklist

- [x] Repository-facing content is written in English.
- [x] Branch name, commit messages, and PR title follow `docs/git-naming-conventions.md`.
- [x] No secrets, tokens, private phone numbers, call recordings, or private transcripts are included.
[...]

PR #633, opened 2026-09-14
feat(apps): add reachable, a school contact-health and absence follow-up app

Uchebuzz:
## Summary

Adds `apps/python/reachable`, one app with two connected phone workflows for a
UK school office.

**Contact check** (termly) rings each emergency contact to establish that the
number still reaches the named person and that they are still willing to be a
contact. **Pattern follow-up** fires when a pupil has two consecutive register
sessions coded N — "reason for absence not yet established" — and cascades
through that pupil's contacts in the school's own listed order, stopping at the
first confirmed conversation.

The two feed each other, which is the point. A wrong number found during a real
absence call is flagged in contact health the moment that call ends and the
cascade moves on; the termly check is what makes the cascade order worth
following in the first place. The contact list is maintained *by* the calling
rather than in spite of it.

The DfE's statutory guidance advises schools to hold more than one emergency
contact number per pupil, and expects them to set out processes "for example
first day calling **and processes to follow up on unexplained absence**", with
further contact where absence continues without explanation. Reachable automates
the second duty only. Citations with page numbers are in `docs/SOURCES.md` §3.

### How this differs from Roll Call (#325)

Roll Call is a good app doing first-hour absence verification. This is
deliberately not that:

| | Roll Call | Reachable |
| --- | --- | --- |
| Trigger | First hour of the first day | A **pattern** across sessions, after the school's own first-day process |
| Output | A safeguarding verdict per child | Support needs and a **suggested** register reason. There is no outcome meaning "accounted for" |
| Data | Consumes the contact list | **Maintains** it — half the app is the contact check |

Reachable requires `school.csv` to record a first-day process and refuses to run
Workflow B without one: a school that lacks one needs a first-day process, not an
automated second-day call standing in for it. No code is shared — Roll Call is
TypeScript, this is Python.

### The four fixes asked of Roll Call

Adopted from the first commit, each with its own tests in
`tests/test_review_requirements.py`, written with the adversarial cases:

- **Credential allowlist** — bearer credentials go only to the official CALL-E
  origin or a loopback fake. Rejected **at startup, before the API key is read**;
  a tripwire test proves the key is never touched when the origin is refused.
  Matching is on the parsed host, so `api.heycall-e.com.attacker.test` and
  `api.heycall-e.com@evil.test` are refused. Two modules see the key; no other
  HTTP client exists in the app.
- **Sanitisation** — provider free text is cleaned at the ingestion boundary
  (control characters and bidirectional overrides stripped, length capped).
  Provider `summary` and `evidence` are not persisted at all — no column exists.
  Templates autoescape, none uses `|safe`, and CSV exports neutralise leading
  `=`, `+`, `-`, `@`.
- **Result binding** — a result counts only if call id, idempotency key,
  destination, pupil and contact all match what was reserved, *and* an identity
  confirmation is supported by a transcript turn the recipient actually spoke.
  `speaker: bot` and `speaker: unknown` are not evidence. An unbound but alarming
  result goes to human review, not to an escalation.
- **Strict ASCII E.164** before any live call, at import and again at dial time.
  Rejected, never repaired; Unicode digits refused rather than normalised.

## Safety

- **Dry run by default.** A call needs `REACHABLE_LIVE_CALLS=1` **and** a
  per-call confirmation typed on that case's page. The dashboard always shows a
  `DRY-RUN` / `FAKE` / `LIVE` banner.
- **Ten guards** are evaluated in code before any dial. `orchestrator.py` is the
  only caller of `client.create`, and the guards and the ledger reservation both
  precede it — asserted by tests.
- **Minimal disclosure.** Nothing about a child is said until the named contact
  confirms their identity, and then only the child's first name. Voicemail and
  anyone else hear only a request to call the school office.
- **Vulnerable-flagged pupils are never called**, gated before a contact is even
  selected.
- **Never concludes a child is safe.** No state, enum or report cell means it.
- **Never writes to a register** and never records an attendance code; reasons
  are suggestions staff approve. No register-writing function exists.
- **Never collects a phone number by voice** — there is no schema field that
  could hold one.
- **English only on UK lines**, per CALL-E's own region table. A contact needing
  another language is surfaced as a staff task rather than quietly skipped.
- **Numbers masked everywhere**; fixtures use only Ofcom's reserved drama range
  with fictional pupils at a fictional school.
- A **data protection impact assessment** note: the school is the controller,
  and the app claims no legal compliance.

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

## Verification

All commands from `apps/python/reachable`. Nothing below uses the network or
reads a credential.

```bash
uv sync --dev
uv run pytest -q
# 423 passed
```

| Area | Tests |
| --- | --- |
| State machines, every transition in `docs/STATE_MACHINE.md` | 69 |
| Call layer: contracts, binding, dispositions | 66 |
| The four review requirements, adversarially | 45 |
| Orchestrator and the loop between workflows | 35 |
| CSV import and row-level rejection | 32 |
| Strict ASCII E.164 and masking | 36 |
| Replay scenarios and CLI | 25 |
| Session counting over a school calendar | 22 |
| Credential origin allowlist | 23 |
| Hardening: concurrency, crash recovery, hygiene | 21 |
| Dashboard | 21 |
| Sanitisation | 16 |
| Exports | 9 |
| Smoke | 3 |

```bash
uv run reachable replay
# Contact check sweep          calls placed 3   expectations all met
# Pattern follow-up cascade    calls placed 2   expectations all met
# Urgent escalation            calls placed 1   expectations all met
# Voicemail, then a retry      calls placed 2   expectations all met
# Vulnerable pupil not called  calls placed 0   expectations all met
```

From the repository root:

```bash
# Repository validation passed.
```

Behaviour worth checking by hand, all without credentials:

```bash
uv run reachable import          # rejects one bad number, masked, with its spreadsheet row
uv run reachable scan-register   # four triggers; the vulnerable pupil goes to PF_NOT_CALLED
uv run reachable preview PF-P-1041-2026-09-11   # identity gate precedes the child's name
uv run reachable serve           # dashboard; a confirmed call in DRY-RUN still places nothing
```

### Live verification

Two live calls were placed, both to a mobile the maintainer owns and offered
for testing, with one sample contact temporarily repointed at it by
`reachable live-contact`. The fictional number is back in `sample_data/` and no
real number appears in any commit. Everything else in the demo runs against the
in-process fake, with no network.

**Call 1** verified the contract. All 16 required `CallTask` fields were
present; transcript turns arrived nested at
`recipients[].attempts[].transcript_turns` with `offset_seconds` / `speaker` /
`text` and a lowercase `bot` speaker label, which is what the binding code
assumes. The recipient answered but did not speak, and the binding correctly
found zero recipient-spoken turns and refused to treat identity as confirmed.

**Call 2** ran the contact-check workflow end to end. The agent disclosed
itself, asked "Am I speaking to Marie Dunn?" and named the pupil only after the
recipient answered. `identity_confirmed: yes` was backed by an actual `user`
turn, `verbatim_identity_quote` matched it, confidence was 0.95/high, the result
validated against the schema, and the case moved to `CC_VERIFIED`. No phone
number was asked for or accepted.

Two defects were found this way and are fixed in this branch:

- CALL-E returned `task_completed: false` alongside a schema-valid
  `outcome: "no_answer"`. The classifier read `task_completed` first, so an
  ordinary unanswered call became NEEDS_HUMAN: the cascade would never have
  advanced and the attempt budget would never have engaged. The fake asserted
  `task_completed: true` everywhere, which is why the wrong assumption survived
  to a live call; it now matches the API and a test pins it there.
- The provider left a call `queued` for about a minute. Reconciling in that
  window retired the attempt-level "never two calls for one case" guard while
  the telephone was still ringing. The case state still refused a second dial,
  so nothing was double-dialled, but the attempt now stays in flight until the
  call is genuinely terminal.

## Notes for the reviewer

- `docs/SOURCES.md` records every fact with its citation and lists the conflicts
  found while verifying, including two places where the DfE guidance contradicts
  itself on whether the code-N deadline is 5 *working* or 5 *school* days. The
  Regulations settle it as school days, which is what the app counts.
- Idempotency keys carry an authorisation ordinal. Without it the state machine's
  "a further attempt may be authorised" branch is unreachable and the attempt
  budget is dead configuration; the reasoning is recorded in `docs/SAFETY.md` §6.
- CALL-E is used through the `calle-ai` SDK with non-blocking `create` plus our
  own poller, rather than `create_and_wait`, so a process killed mid-call resumes
  by reading the call back instead of dialling again. There is a test for that.


Ray-56:
Superseding the earlier review: Applied the existing display-only phone masker to new demo JSON projections and provider diagnostics, removing the exact synthetic/real configured key before truncation. Private stored evidence and decision logic remain unchanged. Added four targeted regressions; 62 focused tests plus four subtests passed with a clean environment and OS network denial. No current Must Fix remains under the community/demo review policy; optional hardening is non-blocking.

PR #652, opened 2026-09-14
feat(withheld): add errand engine that refuses callbacks

Lanceky:
## Summary

Some things can only be arranged by phone: a clinic with no online booking, a garage, a landlord, a council office whose web form has been "temporarily unavailable" for two years. Every one of those errands quietly assumes one thing — that you can answer when someone rings you back.

If you are Deaf, non-speaking, or have a speech disability, `"What's a good number to reach you on?"` is not a formality. It is where the errand stops.

**Withheld's thesis is one line: a callback is a door you cannot open, so never accept one.**

When the other side offers to ring back, the agent declines and asks for a window to call *in* instead. The chain then places that next leg itself. A callback offer becomes a branch that continues the errand rather than an outcome that ends it, and the person never answers a phone at any point.

### What is actually new here

This repository already has neighbours in the disclosure-bounded calling space — [`call-on-behalf`](apps/typescript/call-on-behalf/) in particular arrived at a very similar design independently, and the app README says so plainly and credits it. The authorised-fact list, the pre/post-call gates, masked findings and transcript-bound answers are convergent, not novel.

What I could not find anywhere in the repository is the **callback gate**: treating `"we'll call you back"` as a state transition to be refused and converted, and carrying the errand across multiple legs on that basis. A repo-wide search for callback-refusal language returns nothing. That mechanic, and the multi-leg chain built on it, is the contribution.

### Design decisions worth a maintainer's attention

- **The number cannot leak because there is nowhere to put it.** The recipient result schema has no phone field. `buildScript` refuses to emit a script containing the number, and every transcript is re-scanned for it afterwards — including digits spoken as words (`"oh double seven oh oh, nine oh oh"`), which a digits-only scan misses. A leak downgrades the errand to `outcome_unknown` and halts the chain **even if the booking succeeded**.
- **Nothing trusts the provider's verdict.** CALL-E's `taskCompleted`, `completionConfidence` and `structuredResult` are recorded as telemetry and never read by a decision. Every reported answer is re-derived from `transcriptTurns` and carries the callee turn supporting it. An answer with no supporting turn is `null`, not a guess.
- **Unattributed turns confirm nothing.** A turn CALL-E labels `unknown` can never confirm a window or bind an answer — but it *is* still scanned for the number, because a number audible on an open line was disclosed regardless of who said it.
- **No scheduler.** Chains emit a `ScheduledIntent` and stop. An app that decides both *whether* and *when* to call can ring a stranger at 3am from a parsing bug.
- **The reason stays local.** The errand file holds why the person cannot use a phone; that field is never sent, and the app has a test asserting it.

I also found and fixed a leak class that may be worth others knowing about: Next.js server components serialise **every** prop passed to a client component into the page's flight payload. A number masked on screen was fully readable in view-source. `src/app/boundary.test.ts` locks that shut.

**Live console: <https://withheld.vercel.app/>** — fixture replay only, so it places no calls and needs no credential. [`?s=leaked`](https://withheld.vercel.app/?s=leaked) is the one to open: the clinic says the number aloud, and it is masked in the report and absent from the page source.

## Type

- [x] New runnable app

## Checklist

### Notes on the checklist

**Fixtures and numbers.** Every committed number is in a regulator-reserved drama range (Ofcom `+44 7700 900xxx` / `+44 20 7946 0xxx`). `src/core/numbers.test.ts` walks every committed JSON file and fails the build on anything outside those ranges, so this cannot rot. All transcripts are written, not recorded.

**No-call path.** Fixture replay is the default. All 87 tests and every CLI command except `run --real` work with no API key and place no calls. Live calling requires an explicit flag plus `CALLE_API_KEY`.

**Side effects and cancellation.** `README.md` documents what a real run does and does not do; `SAFETY.md` covers cancellation, medical/legal/financial boundaries, and §13 lists honestly what Withheld *cannot* protect against.

**On live verification:** I did not have CALL-E credentials during the build, so the `--real` path is written against the documented API but has not been exercised against it. The app README states this rather than implying otherwise, and I would rather disclose it than have a maintainer discover it.

## Verification

```
npm install && npm test        # 131 passed (7 files), no network, no credential
npx tsc --noEmit               # clean
npx next build                 # succeeds
python3 scripts/validate_repository.py   # Repository validation passed.
```

Verified from a clean clone of this repository, not just locally.

### Both CALL-E surfaces are wired, because the install guide only produces one

The published installation guide ends at `calle auth login`, which authorises the
MCP CLI and never produces an API key. An SDK-only integration is therefore
unreachable for anyone following the official path, so there are two providers
behind one `CallProvider` interface — `--via=sdk` (needs `CALLE_API_KEY`) and
`--via=cli` (needs the CLI login) — and the chain logic and every gate are
identical either way.

I then authenticated and read the live `get_call_run` output schema back from
`calle mcp tools`, and diffed it against the CLI reference doc. Three things
differ, and all three fail **silently** rather than loudly:

| Reference | Live schema | Failure if unhandled |
| --- | --- | --- |
| `NO_ANSWER` | `NO ANSWER` (space) | Status never matches a terminal value, so the poll loop runs to its timeout on every unanswered call |
| `transcript` | `result.transcript` | Wrong key yields null, which is indistinguishable from nobody speaking — every errand reports as a no-answer |
| — | `PREPARING`, `SCHEDULED` | Undocumented pre-dial states |

All three are handled and covered by tests named after the schema they came
from. The CLI also exposes the run's own verdict at `result.outcome` and
`result.extracted`; both are recorded like the SDK's and read by nothing.

These are written up for the feedback survey along with the `calle` binary name
collision: `@call-e/cli` and the SDK `@call-e/calle` both install a command
called `calle`, so with the SDK as a project dependency `node_modules/.bin`
shadows the real CLI and every command fails `Unknown command: auth`. This app
resolves the binary itself and skips `node_modules` on `PATH`.

No call was placed to find any of this.

## One README fix outside my app

My awesome-list row is placed at the end of the contiguous table (after `apps/python/firstbell`) rather than after the literal last entry. A blank line and a stray bullet above `apps/python/positive-contact` terminate the markdown table early, so rows appended there render as plain pipe-text absorbed into the preceding list item instead of as table rows. I placed mine inside the table so it renders, and left the existing rows untouched rather than reformatting someone else's entry in my PR. Happy to fix that separately if you'd like.


Ray-56:

1. After a live leg agrees a future window, return its ScheduledIntent; do not immediately loop into another placeCall before that window (src/core/chain.ts:121–155).
2. Keep real credentials on approved HTTPS; allow HTTP loopback only with explicit fake credentials (src/core/calle.ts:222).
3. Mask raw answer/evidence/error displays and qualify crash-proof dedupe, all-form recognition and Ctrl-C cancellation guarantees as actual bounded behavior (scripts/withheld.ts:95–142,273; SAFETY.md:165).

Lanceky:
Thanks — all three were real, and two of them were looser than I'd claimed in writing. Fixed in the commits above.

**1. Dialling ahead of an agreed window** (`src/core/chain.ts`)

This was a genuine bug, and it contradicted the file's own header. After a leg agreed "call me back between ten and twelve", the loop fell straight through to the next `placeCall`. A provider that places real calls now returns the `ScheduledIntent` unplaced and halts with a stated reason:

> The callee agreed a window that has not opened yet. The intent for the next leg is returned unplaced; this app runs no scheduler and will not dial ahead of an agreed time.

Fixture replay is exempt, because it places no calls — the two-leg demo is unchanged. An unparseable `not_before` counts as **not** open: the app would rather hand a human an intent it could have placed than place one it should not have, since only the second mistake rings a telephone.

The open-window branch is covered by unit tests on the predicate rather than through the chain. The chain fixture names its window relative to `now` ("tomorrow between ten and twelve"), so that branch is genuinely unreachable through it, and a test that appeared to cover it was passing for the wrong reason — it halted earlier, on acceptable-hours. I'd rather test the gate than ship a green test that proves nothing.

**2. Credentials on approved HTTPS** (`src/core/calle.ts`)

`approvedBaseUrl` now takes the key. Loopback is allowed only for a key that announces itself as fake (`test-`, `fake-`, `dummy-`, `sim-`, `local-`); anything the check is unsure about is treated as real, because the cost of guessing wrong is a live credential in the clear. `my-test-key` and `testing` do not qualify.

Writing the test found a second defect in my own fix: officialness was compared by `origin`, so `http://api.heycall-e.com` fell through to the unknown-origin branch and the plaintext check was unreachable. It's compared by host now, so plaintext to the live host is refused *as* plaintext.

**3. Masked displays, and three guarantees narrowed to what the code does**

Answers, supporting turns, disclosures and the top-level crash handler now pass through `redactForDisplay`. It redacts the whole string rather than splicing digits out: the scanner works on normalised and spoken forms ("oh double seven oh oh") that don't map back to a span of the original, so a surgical edit would be guesswork. Over-redacting costs a reader one lookup; the other direction prints the number this app exists to withhold.

`SAFETY.md` no longer claims more than it can keep:

- **Dedupe** is bounded to the SDK path and credited to CALL-E rather than to this app. The MCP CLI path has no idempotency key at all, which is now stated plainly — there, a submission that may be in flight is marked `retry_safe: false` and stops for a human. Not re-dialling is the whole mitigation.
- **"any form"** is now the forms the scanner actually handles, named, with a note that it's a pattern scan and the schema is the control.
- **Ctrl-C** is the cancellation point only *before* dialling, via the three-second pause. Afterwards it ends the local process and abandons the poll loop; it does not reach CALL-E and does not hang up a connected call.

**One thing you didn't ask for.** `replay` read the wall clock, so the fixtures had drifted past the errand's acceptable hours and the documented two-leg chain no longer reproduced. It's anchored to the fixture's recorded time now. A replay that reads the clock isn't a replay.

153 tests (up from 131), `tsc --noEmit` clean, `next build` clean, and zero disclosures across all six scenarios. Also merged `main` and re-appended our row as the last row of the contiguous table — every row added upstream in the meantime is preserved.

Ray-56:

PR #655, opened 2026-09-14
feat(opportunity-verification-call): add opportunity verification skill

RobynAwesome:
Adds opportunity-verification-call skill


Ray-56:
Superseding the earlier review: Linked safety and fictional example from SKILL.md; added no-call preview, authorized E.164 recipient, per-intent approval, one-submit, UNKNOWN-stop, and human-review host sequence. Replaced implemented lock/durable-receipt/cancellation claims with explicit host responsibilities and accepted-call continuation limits. No current Must Fix remains under the community/demo policy; optional hardening is non-blocking.

PR #657, opened 2026-09-14
feat(mission-replan-loop): add evidence-driven replanning skill

guelmbaye:
## Summary

`mission-replan-loop` covers the case where a phone answer does not complete a
workflow step but **makes the current plan impossible**.

A carrier confirms a 6 PM pickup against a 5:30 PM warehouse cutoff. The answer is
factual, confirmed, and fatal: the strategy is invalidated and the next call exists
*because of* what the last one revealed.

| Call result | Strategy status | Replan | Next call |
|---|---|---|---|
| `pickup_time: 16:00` | `valid` | no | **none** |
| `pickup_time: 18:00` | `invalid` | yes | alternative candidate |

That second row is the skill. Same mission, same constraints — only the phone answer
differs, and the next action differs with it. Existing skills in `skills/` turn a call
into a result; this one turns a call into a *decision about the plan*.

It is not logistics-specific. Procurement (quantity against a required volume) and
field service (a technician against a service window) run the same path with different
constraints. All three are worked examples.

**What it does:** normalises a CALL-E result into evidence keeping `call_id`, target and
timestamp on every fact; evaluates it against the mission's constraints
deterministically; returns `VALID`, `INVALID`, `UNCERTAIN` or `INFEASIBLE`; on `INVALID`
proposes a materially different strategy and the next operation needed to validate it;
on `UNCERTAIN` proposes a clarification, never a decision; on `INFEASIBLE` escalates.
The skill proposes — the host authorises and executes.

**Rules it carries**, several learned against the live API rather than designed:

- **A call that could not be placed says nothing about the person on the other end.**
  A recipient failure (no answer, busy, declined) and an execution failure (unsupported
  destination, malformed request, auth error) must not collapse into one story. The
  failure mode is specific: a configuration error walks the whole candidate list, spends
  the budget, and reports *"no candidate could be reached"* — a fluent business
  conclusion covering a platform fault, with no phone ever ringing.
  `references/safety.md`, plus a worked example.
- **A deadline has two ends.** "Before 17:30" is satisfied by every past time, so a 16:45
  pickup offered at 17:01 passes a naive comparison.
- **Hedged language is not a fact.** "Probably around five-thirty" cannot satisfy a hard
  constraint; it earns a clarification.
- **A fact nobody mentioned produces no record.** Silence is not a value.
- **Prefer a string enum with an explicit `unknown` over a nullable boolean** when
  defining what a call should return. An unanswered question must never read as "no".

```
skills/mission-replan-loop/
├── SKILL.md
├── references/ safety.md · examples.md · replanning-policy.md · mission-state.md
├── examples/ five fixtures, one per outcome shape
└── schemas/ input and output contracts
```

No dependency on any particular runtime, database, UI or orchestration stack.

## Type

- [x] New skill
- [ ] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

- [x] Real-world side effects are clearly described. The skill proposes calls and never
      places them; `references/safety.md` states what a host must authorise, and
      `references/replanning-policy.md` sets the budgets that stop a loop.
- [x] Phone numbers are masked in documentation and test fixtures unless they are clearly
      fictional. The skill contains **none at all** — it takes a target by name and never
      handles a number, since placing the call belongs to the host.
- [x] Recurring workflows include cancellation behavior. Retry, replan and operation
      budgets are in `references/replanning-policy.md`; an execution failure stops the
      loop immediately rather than walking the candidate list.
- [x] Runnable code has a dry-run, fake-server, or no-call path by default. The skill is
      declarative and ships no scripts. The five fixtures are input/expected-output pairs
      that need no credentials and place no calls.


Ray-56:
Superseding the earlier review: Aligned the supplied input schema with the rejected/no-call fixture and restored the procurement example evidence from its existing synthetic input. All five input/output fixture pairs validate against the supplied schemas without network access. No executor or external mutation was added. No current Must Fix remains under the community/demo review policy; optional hardening is non-blocking.

PR #668, opened 2026-09-14
feat(apps): add MindQuark consent-gated CALL-E wellbeing check-in app

kkpg-l:
Adds MindQuark under User-facing Apps (`apps/typescript/mindquark/`) plus the README catalog list entry and table row.

MindQuark is a CBT-informed wellbeing web app whose headline feature is a consent-gated outbound CALL-E voice check-in: the user taps Call Me, enters an E.164 number, ticks explicit consent, and Maya (the AI companion) rings their phone for a 5-10 minute supportive call. Safety posture: per-IP daily quota (3), concurrency cap (1), unsupported regions (+86) rejected up front, zero phone-number persistence, server-only credentials, and crisis content short-circuits the LLM to verified hotlines (988/local). Each call is one explicit user action - no recurring jobs.

The catalog README documents the workflow boundary, timing behavior, a no-call reproduction path (`npm test` with hermetic mocked CALL-E, pinned to revision `e72ad23`), live-call side effects, and cancellation. `python3 scripts/validate_repository.py` passes.


Ray-56:
Superseding the earlier review: Qualified spoken-call crisis handling as prompt-based best effort, not deterministic detection or guaranteed intervention, in the workflow, diagram and safety summary. Distinguished the recognized-text chat interceptor from the phone model's advisory behavior. No current Must Fix remains under the community/demo policy; optional production hardening is non-blocking.

PR #670, opened 2026-09-14
Add provenance-grade: knowledge grades (verified/asserted/assumed) for extracted call fields

manasnikam:
**Demo video (2:38): https://youtu.be/z4U2wO8_7o0**

## What this adds

`skills/provenance-grade/` — a skill that attaches a knowledge grade — **`verified` / `asserted` / `assumed`** — to every field a CALL-E phone agent extracts, computed from the `transcript_turns` the API already returns. Two calls can both answer "Tuesday"; this tells you which one was looked up and which one was guessed.

## Why it is not a duplicate

CALL-E's `completion_confidence` is, by its own docs, confidence that **the task reached a clear end state** — explicitly not confidence in the business answer. This fills exactly that documented gap. `call-summarizer` summarises after the fact, `voice-preflight` checks audio before the call, `linecanary-monitor` watches line health — nothing in the repo grades the epistemic basis of a spoken claim.

## How it works

- **Pre-call lint** (`scripts/lint-task.ts`): amends task text and `result_schema` so the signals are elicitable — read-backs for critical values, "can you check?" instead of "do you know?", one corroborating specific.
- **Post-call grader** (`scripts/grade.ts`): nine signals — retrieval gap, check language, corroborating specific, hedging, deferral, read-back compliance, round-number shape, answer alignment, self-correction — feed a fixed, fail-closed rule table. Default is `assumed`; absence of signal never upgrades; a field absent from the transcript is `unstated` and never graded.
- The two semantic signals (C, H) expose provider interfaces where a model may return **spans, never a grade** — the grade always comes from the rule table, which is what makes it testable.

## Evidence (offline, zero calls, zero network)

- 24 hand-labelled fixtures covering verified, asserted, hedged, deferral, misaligned, self-correction, round-number, read-back complied/refused, voicemail, IVR, and code-switched Hindi-English.
- 125 tests (`npm test`), confusion matrix via `npm run eval`: **23/24 (95.8%)**. The single miss is shipped deliberately — a Hindi hedge ("shayad") the English lexicon cannot see, documented in `references/limitations.md` as the honest measurement of that gap.
- No-inflation property is tested: no `assumed`/`unstated` ground truth is ever predicted `verified`.

## Safety and ethics

Grades behaviour, not people: no emotion/stress/deception scoring (impossible by construction — text and integer offsets only). Grades attach to the **organisation**, never the individual; tests assert the output schema carries no person-identifying field. A low grade means *the call did not establish this*, never *this person lied*.

## Consumer rule

A workflow may act on `verified`, may act on `asserted` with a human in the loop, and must never auto-act on `assumed` or `unstated`.

Dependencies: zod + vitest only. Runs on Node ≥ 22.6 natively (type-stripping).

https://claude.ai/code/session_01HZQJ2Rcs7XdbijEpM62FsZ


Ray-56:
Superseding the earlier review: Removed exactly the four unrelated .omc state/replay files introduced by the PR; no other state/files were removed. Replaced structural anonymization, guaranteed lookup/verified and impossible-ASR-overgrade claims with advisory heuristics and caller responsibilities for identifiers, retained evidence and display/export redaction. Only a misleading test description changed; grader implementation and private evidence behavior remain unchanged. No current Must Fix remains under the community/demo policy; optional hardening is non-blocking.

PR #679, opened 2026-09-14
feat(snoozetax): add wake-up verification call app

crjstobal:
## Summary

Adds **SnoozeTax** under `apps/web/snoozetax/`: an alarm clock that treats the
phone call as a verification layer rather than as the alarm itself.

Every alarm app measures a tap, and tapping is something people do in their
sleep. A conversation is different. At wake-up time CALL-E places a call, reads
out a three digit code, and asks for it back in reverse order. Repeating a code
proves you can hear; reversing one proves somebody is home. `result_schema`
returns both whether the answer was right and whether the voice sounded groggy.

A missed call is never read as oversleeping: Do Not Disturb, no signal and a
flat battery are indistinguishable from it, so a missed call opens a ten-minute
window to answer on the web instead. Only when that expires does CALL-E place a
second call, to the accountability contact the user nominated.

It belongs here as a runnable app: four distinct CALL-E call types, each with
its own task prompt and result schema, a vendor-neutral state machine with a
single adapter file, and a dry-run default so the whole thing runs with no API
key and no credits spent.

**How CALL-E is used**

| Feature | Where |
|---|---|
| Outbound calls (`POST /v1/calls`) | Verification, consent, wake-up and accountability calls |
| `result_schema` structured extraction | Grades the spoken answer and judges grogginess in one pass |
| Terminal webhooks | `call.completed` drives the wake state machine |
| Transcripts | Stored per wake event and shown in History |
| `Idempotency-Key` | Stops a retry or a double cron tick from billing twice |

Business logic never imports the vendor SDK: `lib/wake.js` returns intents and
`adapters/calle.js` is the only file that knows CALL-E exists. Every prompt and
schema is browsable at `/admin`, rendered from the same builders the live calls
use, so the page cannot drift from what is actually spoken.

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

### Notes on the checklist

**Side effects.** The README has a table of every outbound call with its cost:
sign-up verification, the one-off consent call, the wake-up call, one retry when
unanswered, and the accountability call. Worst case is three calls in a morning,
roughly $0.15. Nothing else leaves the machine: no payments, no SMS, no
third-party analytics.

**Cancellation.** There is a "Cancelling and rolling back" section covering how
to stop one alarm, all calls globally (`DRY_RUN=true`, the default), the
scheduler, and the accountability contact. It also states plainly that a call
already handed to CALL-E cannot be recalled.

**Phone numbers.** Every sample uses the Ofcom standards-reserved fictional
range `+44 7700 900xxx`. Credentials are read at runtime from AWS SSM and never
written to disk; `.env.example` contains only the placeholder
`CALLE_API_KEY=iams_live_your_key_here`.

**Consent.** An accountability contact is a third party who never signed up, so
nominating them only marks the number `pending`. CALL-E phones them once to ask,
and the code refuses to dial any contact whose status is not `confirmed`. A
`declined` number can never be re-asked through the API.

**Dry-run.** `DRY_RUN` defaults to true. `npm test` runs 58 end-to-end checks
with no credits spent.


Ray-56:

1. Close account takeover: auth/start reveals the expected answer, auth/verify grants/replaces sessions without phone proof, and auth/status issues sessions from a phone query. Bind proof to the initiating session and protect live calls/private records; obtain authorization before the first third-party consent call (backend/index.js:76–131,221).
2. Authenticate or authoritatively re-fetch bound webhook outcomes before changing account consent, wake state or future calling (backend/index.js:433–497).
3. Stop unknown/in-flight wake events rather than retrying or automatically calling an accountability contact (backend/lib/wake.js:118,235–246; index.js:547–549).
4. Restrict real credentials to approved HTTPS origins and mask phone-bearing task/transcript/quote/error displays (lib/secrets.js:63–73; adapters/calle.js:123–154; index.js:620). Qualify consciousness/still-in-bed assertions as advisory missed-check-in outcomes and align the stale verification instructions.

PR #690, opened 2026-09-14
feat(apps): add Gimme Updates - daily voice digest for blind and elderly users

Vanshika-Rana:
## Summary

Gimme Updates - a daily phone call that reads a user's important emails aloud and captures spoken decisions (reminders, follow-ups, or skip), built for people who are blind, elderly, or who simply don't want to navigate a screen-based inbox. Built for the "Your Code Is Calling" hackathon.

The core of the app is a single interactive CALL-E task (`lib/calle.ts`, `runDigestCall`) that receives the day's emails prioritized by urgency, reads through them one at a time in plain language, asks after each one whether the user wants a reminder, a follow-up, or to move on, and waits for a spoken response before continuing. It returns a structured result (`resultSchema`) capturing a decision per email. A second, lighter CALL-E task places a reminder callback when a previously scheduled reminder comes due, triggered by a cron route.

The "inbox" is a seeded set of sample emails rather than a live Gmail connection, so the demo is reliable to run without external auth dependencies — the classification (via OpenRouter) and CALL-E integration work identically against a real inbox. Gmail integration is the first item on the roadmap.

Live demo: [your Vercel URL]
Demo video (~3 min): [your YouTube URL]

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

- [ ] Recurring workflows include cancellation behavior.
- [ ] `python3 scripts/validate_repository.py` passes.


Ray-56:

1. Protect public digest/cron/signup/schedule and private-record access, require operator intent and authorized E.164 recipients, or keep the reachable app reliably fake-only (app/api/calle/digest/route.ts:7; app/api/cron/route.ts:162).
2. Do not leave ambiguous submissions eligible for the next cron tick to call again. Preserve an unresolved intent and require reconciliation; simulation-created reminders must not become live merely by changing an environment flag (cron/route.ts:126–139).
3. Mask phone/task/result/error display and logging copies (lib/calle.ts:170,249; signup/digest/cron responses).

Ray-56:

1. Enforce operator/owner authorization for reachable live-call/private-record routes and bind an authorized ASCII E.164 recipient; unused helpers and deployment flags do not protect the routes.
2. Do not backfill legacy simulated reminders as live-eligible: migration 0005 currently sets every old reminder is_simulated=false. Hold them unless deliberately reauthorized.
3. Redact raw provider exceptions and phone-bearing public subject/decisionDetail strings in digest/cron display and log copies.

Vanshika-Rana:
Addressed all three points from the latest review:

1. Real calls now require an OPERATOR_SECRET (sent as `Authorization: Bearer <secret>` or JSON `operatorSecret`) on every route that can trigger a call or expose records (digest, cron, email fetch, reminders, schedule). ALLOW_REAL_CALLS=true alone is no longer sufficient. Real calls also only proceed if the target number is valid ASCII E.164 and present in ALLOWED_RECIPIENTS — enforced both at the route level and again inside lib/calle.ts before the SDK is invoked, so a missed route check can't dial.

2. Migration 0006 forces every pre-existing reminder that migration 0005 had marked is_simulated=false back to is_simulated=true and status='unresolved'. Nothing old is live-call-eligible without deliberate reauthorization.

3. CALL-E error objects are scrubbed of phone numbers before logging or returning in any response. Digest/cron logs and responses now truncate email subjects and decisionDetail rather than echoing them in full.

PR #692, opened 2026-09-14
Add GoodFaith: self-pay Good Faith Estimate cash-price agent (apps/typescript)

dmustapha:
# Add GoodFaith (apps/typescript)

## What it is

GoodFaith is an AI phone agent that calls imaging clinics for a self-pay Good Faith Estimate cash price, enforces comparability on the call, and returns a confidence-gated, transcript-evidenced landed-cost comparison benchmarked against a fair price.

The demo procedure is an MRI of the lumbar spine without contrast (CPT 72148), which commonly runs from about $400 to $2,000+ in the same metro area on the same day (published self-pay / cash-price data; our live demo winner is $438, recomputed from the sample transcript). Under the No Surprises Act, self-pay patients have a legal right to a Good Faith Estimate, but exercising it means calling a dozen billing desks and normalizing incompatible answers by hand. GoodFaith does that labor with an AI agent, in parallel, with receipts.

## Why it fits this repo

It is portable and provider-separated exactly the way the repo asks for. CALL-E places the calls; the app orchestrates the shortlist, extraction, gating, and comparison. Safety is on by default: mock-first with zero outbound calls, server-side-only credentials, explicit opt-in for live calling. It fills a whitespace the existing apps do not cover: parallel multi-vendor price discovery with an auto-built, evidence-backed comparison, rather than single-call confirmations or IVR navigation.

## What is runnable

Mock mode is the default and needs no credentials or network:

```bash
cd apps/typescript/goodfaith
pnpm install
pnpm dev        # http://localhost:3000
```

Load the sample clinics and request a cash price. You get the winner-flip comparison and an expandable per-row transcript receipt.

Non-browser verification:

```bash
pnpm typecheck  # 0 errors
pnpm test       # normalizer + route + webhook tests
pnpm verify     # recomputes every headline number from committed fixtures
pnpm build      # production build, all 8 routes
```

Live mode is opt-in: set `CALLE_API_KEY` and `GOODFAITH_LIVE=1` in `.env.local`. A real outbound call is placed only when both are present.

Live demo (mock-first, no login): https://goodfaith-nu.vercel.app

## The four load-bearing CALL-E surfaces

1. **Multi-recipient parallel calls.** One task fans out to every clinic via `recipients[]` and rolls up into a single result (`src/lib/calle.ts`).
2. **Per-recipient structured extraction.** A locked `recipientResultSchema` pulls `cash_price`, `price_basis`, includes/excludes, and the quoted sentence out of each call as typed JSON (`src/lib/schemas.ts`).
3. **Completion-confidence gating.** Any call below 0.6 confidence is held back and never ranked, fail-closed (`src/lib/normalize.ts`).
4. **Evidence / transcript audit trail.** Every ranked price traces to the exact transcript turn it was said in; a price with no traceable utterance is refused (`src/lib/normalize.ts`, `src/components/AuditTrail.tsx`).

Also implemented: an idempotent webhook receiver keyed on `CALL-E-Event-Id`, a live `calls.listEvents` stream, `metadata.rfq_id` correlation, and an optional Goals API path. The `/proof` page maps each surface to its code.

## The differentiator

A naive "cheapest number" ranker picks a $525 facility-fee-only quote that excludes the radiologist read. GoodFaith flags it `NON_COMPARABLE` and surfaces the $438 all-inclusive quote as the true landed-cost winner. Comparability is enforced before ranking, so the lower sticker cannot win by accident.

## Safety compliance

- Credentials are read only in server route handlers and server modules; no key is inlined with `NEXT_PUBLIC_` or sent to the browser.
- Mock is the default; the app is fully functional with zero outbound calls, and every mock result is badged.
- In live mode the agent identifies itself on the call. The app captures no PHI and does not book.
- No medical or legal advice is given, and no price is presented as guaranteed or binding.

A no-fabricated-quotes invariant is enforced structurally: a row can only be ranked when its `quoted_verbatim` is a non-empty string traceable to the transcript, and `scripts/verify-claims.ts` fails the build if any displayed price lacks a traceable utterance and is not mock-badged.


Ray-56:

1. Protect live-call/private quote endpoints; require explicit authorized valid E.164 recipients and reject duplicate destinations (apps/typescript/goodfaith/src/app/api/quotes/route.ts:22–55).
2. Restrict credentials to approved HTTPS CALL-E origins (src/lib/env.ts:12; src/lib/calle.ts:53).
3. Authenticate or authoritatively re-fetch bound webhook updates before accepting replacement RFQ evidence (src/app/api/calle/webhook/route.ts:19–35).
4. Mask phone-bearing normalized results and provider diagnostics (src/lib/normalize.ts:110). Correct the absolute quote/confidence/no-fabrication and sent-schema claims, or label extraction/ranking honestly heuristic and advisory (normalize.ts:78; README).

dmustapha:
Thanks for the detailed review, this was genuinely useful. All four items are addressed in the latest push (`c8229cb`). Point by point:

**1. Quote endpoint** (`apps/typescript/goodfaith/src/app/api/quotes/route.ts`): every clinic now requires a valid E.164 recipient (400 otherwise), duplicate destinations are rejected (400), and the live-dial path is gated by an explicit authorized-recipients allowlist (`GOODFAITH_ALLOWED_RECIPIENTS`). In live mode a recipient not on the list returns 403, and an empty allowlist refuses live dialing entirely (fail-closed by default). Added a per-IP rate limit (429). Mock mode still enforces E.164 + dedupe.

**2. Credentials** (`src/lib/env.ts`, `src/lib/calle.ts`): the API key is only ever sent to an approved HTTPS CALL-E origin. `calleBaseUrlChecked()` hard-throws unless `CALLE_BASE_URL` is `https://` with host `heycall-e.com` or a `*.heycall-e.com` subdomain, the `CalleClient` is only constructed through that check, and the old unchecked getter was removed.

**3. Webhook** (`src/app/api/calle/webhook/route.ts`): the POSTed body is no longer trusted as evidence. We take only the call id from the payload and authoritatively re-fetch the call from CALL-E (`getCall`) over the approved origin, writing only the re-fetched result. Idempotency via `CALL-E-Event-Id` is kept, an optional shared secret (`GOODFAITH_WEBHOOK_SECRET` / `X-GoodFaith-Webhook-Secret`) is enforced when set, and mock mode returns 202 without mutating any RFQ.

**4. Masking + claims**: you were right on the honesty point. Phone-bearing fields in normalized results are masked (`phone`, name fallback) and all transcript-derived free text (`quoted_verbatim`, `includes`, `excludes`) plus summaries and diagnostics are scrubbed of E.164s before leaving the server (`src/lib/normalize.ts`), with a regression test asserting no full number reaches the client. On the claims: the README, /proof, and CLAIMS copy no longer state or imply JSON result schemas are sent (the current API tier rejects them, so fields are derived from the transcript), extraction and comparability are now labeled heuristic and advisory, and the no-fabrication point is stated precisely as traceability (a price is never ranked unless it traces to a real quoted sentence), not as guaranteed accuracy.

Typecheck, 45 unit tests, and the production build are green. Happy to adjust anything further, thanks again for the careful pass.

Ray-56:
Superseding the previous list: the credential-origin guard and destination validation/allowlisting are fixed. The remaining Must Fix items are:

1. Require basic caller authorization for live quote creation and private quote/event reads; a destination allowlist alone does not protect these remote endpoints.
2. Bind webhook updates to the stored RFQ call ID and authoritative fetched metadata. `body.data.metadata.rfq_id` still lets call A's evidence overwrite RFQ B.
3. Mask complete phone/error output paths (quote validation errors, provider diagnostics and phone-as-name fallback), and qualify the remaining exact-traceability/fail-closed claims to match the heuristic advisory implementation. Perfect NLP is not required.

dmustapha:
Thanks, these three are addressed in the latest push (`a682a85`). Point by point:

**1. Caller authorization.** You're right that a destination allowlist doesn't protect the endpoints themselves. Live quote creation and both private reads (`GET /api/quotes/[id]` and `/events`) now require a bearer token (`Authorization: Bearer <GOODFAITH_API_TOKEN>`), enforced in `src/lib/auth.ts`. It is fail-closed: in live mode a server with no token configured refuses all privileged access (401), and a missing or wrong token is 401. Mock mode stays open as the public demo. Token comparison is length-independent and constant-time so it can't be timing-probed.

**2. Webhook binding.** The webhook no longer reads `body.data.metadata.rfq_id` at all. The target RFQ is resolved solely from our stored `callId -> rfq` mapping (`findRfqByCallId`), and as defense in depth the authoritatively re-fetched task's `metadata.rfq_id` must agree with that stored binding (409 on mismatch). So a payload carrying call A's id can only ever refresh the RFQ that actually owns call A; it can never overwrite RFQ B. Covered by a new binding test that asserts a forged `metadata.rfq_id` leaves the other RFQ untouched.

**3. Masking + honest claims.** Completed the phone/error coverage: the quote validation errors (duplicate destination, not-authorized-for-live) now mask the number, provider diagnostics are scrubbed on every failure path (quote create, reload, events), and the phone-as-name fallback is masked, including the case where the adapter sets `name` to the raw phone (the previous `name ?? mask(phone)` never fired, so an E.164 name is now masked and phones embedded in real names are scrubbed). On wording: "exact transcript turn" is gone, replaced with "matched to a transcript turn (heuristic substring match, advisory)", and every `fail-closed`/confidence claim is now explicitly stated as a gate, not a correctness guarantee. Regression tests assert no full E.164 reaches the client on any of these paths.

Typecheck, 57 unit tests, and the production build are green, and I ran a live-server pass confirming 401 on unauthenticated privileged calls, a masked 403 for a non-allowlisted recipient, and scrubbed provider errors. Thanks again for the thorough review.

Ray-56:
The frozen Must Fix list is now closed. Live/private endpoints require authorization and webhooks bind through the stored call ID. I made permitted bounded fixes for formatted-phone masking and qualified the remaining exact-evidence guarantees as advisory substring association, with operator-token setup clarified. Typecheck, all 58 mocked tests, fixture verification and clean-source repository validation pass; no live credentials or calls were used. This supersedes the earlier blockers; broader NLP accuracy, durable recovery and production hardening remain optional follow-ups.

PR #693, opened 2026-09-14
feat: add BRIDGE business coordination app

rahulkumaar03:
BRIDGE is an AI business coordination layer that uses CALL-E to resolve multi-person business workflows through targeted phone conversations.

Core workflow:
Business problem
→ identify the right person
→ CALL-E conversation
→ structured result
→ select next person
→ continue until resolution

Example:
Driver → Warehouse Manager → Operations Head → Customer → Resolved

Features:
Organization-aware person selection
Multi-step workflow orchestration
CALL-E integration
Simulation/Dry-Run mode
Structured call results
Automatic next-person selection
Human escalation
Task timeline
Call history

Safety:
Simulation mode available
Live calls require explicit confirmation
No credentials committed
Example phone numbers are masked/fictional


Ray-56:

1. Protect mode changes, live dispatch and private records; require authorized E.164 recipients and server-side confirmation, or make the reachable app simulation-only (src/server.js:189–247,359–465).
2. Keep simulation unable to invoke the real CLI, and never label fixture fallback or auto-step results as real live completion (src/calle.js:30–49; src/server.js:418–448,598).
3. Escape writable person/workflow/task values before innerHTML rendering to close stored XSS; mask phone-bearing API/UI/result copies and document accepted-call cancellation limits (public/index.html:2558; src/server.js:378–465).

PR #700, opened 2026-09-14
feat: add auditlane skill (telephony-gate Claude Code hook)

soujasK:
## What this adds

`auditlane` — gates dangerous agent actions (`DROP TABLE`, `rm -rf`, `terraform destroy`,
force-push, curl-pipe-to-shell, and 19 other patterns) behind a real, live phone call to a
human before they're allowed to execute.

Runs as a **Claude Code `PreToolUse` hook**, so it sits in the harness's own execution path —
the agent it's guarding against cannot skip it, because it never gets a choice to call it.
Also audits PR text for undocumented claims of verbal human approval, calling the named
person to check. Both share one verification core and fail closed on every error path.

## Testing

127 tests, all offline, no network access or API key required:
- Core verification pipeline (claim extraction, entailment scoring, multi-hop chains)
- 85 cases covering the danger-pattern detector, including near-misses chosen specifically
  to catch false positives
- 18 cases proving the hook cannot crash on malformed/missing/wrong-typed input

Full source: https://github.com/soujasK/AuditLane

## Checklist

- [x] Branch name follows `docs/git-naming-conventions.md`
- [x] `SKILL.md` frontmatter includes `name` (matches directory) and `description`
- [x] `references/safety.md` and `references/examples.md` included
- [x] No real phone numbers or credentials anywhere in the skill directory


Ray-56:
Superseding the earlier review: Generic discovery frontmatter no longer binds the skill to Claude Code; linked host-specific safety boundary describes configured Bash coverage only. Removed unskippable/all-path/guaranteed-authority wording; documented experimental pattern/entailment behavior, fictional allow responses and independent human approval. Restricted unmodified external CLI previews to fictional contacts and required masked display-only real previews without changing private dispatch data. Clarified disabling hooks affects future invocations, not accepted calls; documented unknown-stop and qualified finite author-reported tests and local budget limits. No current Must Fix remains under the community/demo policy; optional hardening is non-blocking.

soujasK:
Hi ray,
I'm Soujas, submitted "AuditLane" for the CALL-E hackathon. I completed the
Devpost submission form about a minute after the deadline wanted to flag it
directly rather than assume it'd get missed.
Everything's real and ready if you're able to consider it:
Repo: https://github.com/soujasK/AuditLane
PR to CALL-E's own repo (reviewed and merged):
https://github.com/CALLE-AI/awesome-phone-call-agents/pull/700
video link-https://www.youtube.com/watch?v=-JWsdmfOql8
I also left a note on the project's Devpost messages page in case that's a
channel you check.
Totally understand if the deadline has to hold firm just wanted to make
sure this reached you directly in case there's any flexibility.
Thanks for running this,
Soujas

PR #703, opened 2026-09-14
feat(apps): add Unitree G1 light callback demo

xexefe121:
Adds a focused Python CALL-E callback example for a Unitree G1 light-switch workflow. An authorized callback collects an explicit decision; polling returns an advisory `turn_off_light` instruction only after a completed call with a matching final affirmative. The instruction remains `not_dispatched` for a separately supplied controller.

The contribution includes only the public phone integration. Robot perception, motion, navigation, camera access, and hardware execution are outside its scope. It does not provide live hardware updates inside an ongoing call.

- Python standard library; no private dependencies.
- No-call preview and synthetic tests need no credentials or hardware.
- Live calls require explicit opt-in; saved idempotency keys support recovery without intentional redial.
- API keys and phone/transcript state stay outside Git.

Validation: repository validator passed; all six unit tests passed; no-call preview and staged whitespace/secret checks passed. This PR's live callback prompt has not been tested end to end with hardware.


Ray-56:
Superseding the earlier review: E.164 fullmatch now uses ASCII [0-9], rejecting Unicode decimal digits. The fixed CALL-E urllib boundary installs a redirect-rejecting opener; no bearer credential is forwarded to any redirected URL. Private request/state and advisory extraction behavior remain unchanged. No current Must Fix remains under the community/demo policy; optional hardening is non-blocking.

PR #707, opened 2026-09-14
feat(apps): add Orma accountability calls

nrynss:
## Summary

Orma is a daily two-minute accountability call. It rings at a time you chose, leads with what you keep not doing, asks what needs capturing, and lets you drop an item without argument.

The call arrives already knowing. SQL counts mentions and ages, and those numbers are rendered into the CALL-E task before dispatch, so the model phrases facts it never computed: "You've mentioned the dentist 3 times. It's been 34 days." A structured result schema writes captured items, retirements with transcript evidence offsets, and commitments back to Postgres.

Orma owns the scheduler, following this repository's provider-separation principle. `pg_cron` materialises runs, a minute tick claims them with `FOR UPDATE SKIP LOCKED` and dispatches with a stable idempotency key, and results come back through a re-fetched webhook and polling.

Surfaces: a SvelteKit web app for setup and review, a Telegram bot for text and voice capture and receipts, and an MCP server with six tools under the user's own JWT and row-level security.

The app lives in `apps/web/orma/` with its own README. It also adds one row to `apps/README.md`.

## Type

- [ ] New skill
- [x] New runnable app
- [ ] New workflow plugin
- [ ] New provider adapter
- [ ] New scheduler recipe

## Checklist

- [x] No secrets, tokens, private phone numbers, call recordings, or private transcripts are included. The recorded CALL-E fixtures keep the provider's response shapes, and their conversation turns are synthetic, as `testdata/calle/README.md` states.
- [x] Real-world side effects are clearly described. The README has a side-effects table covering calls, the three cron jobs, the confirmation call, Telegram messages, pattern report email, and stored voice notes.
- [x] Recurring workflows include cancellation behavior. Users can cancel today's call, pause all calls, revoke consent, or delete the account. Operators can unschedule each cron job and turn dry run back on.
- [x] Runnable code has a dry-run, fake-server, or no-call path by default. `ORMA_DRY_RUN` defaults to true and fails closed when unset.

## Safety

- **Explicit intent.** No call is placed without a consent row that stores the exact wording shown, a slot the user chose, and a confirmed number.
- **Authorized destinations.** Anyone can sign up, so a typed number is not trusted. A one-time call reads a six-digit code, and the user enters it in the app. Only then does the server set the confirmed flag. Users cannot set it themselves, and changing the number clears it. Attempts are rate limited and never redialled.
- **No duplicate calls.** Claims use `FOR UPDATE SKIP LOCKED`, and every dispatch carries a unique idempotency key. A 409 is recorded as a failure, never retried.
- **No redial after an ambiguous outcome.** Unanswered and failed calls are recorded and never retried.
- **Credentials.** Only the publishable key reaches the browser, and the build fails if a secret key appears under `web/`. Every function reaches Supabase through a configured front door, never a project host.

## Verification

Run locally with no hosted services and no calls:

- `scripts/p6-local-stack.sh up`, then the web app against it, as the README walkthrough describes.
- Deno suites for dispatch, dry run, ingest, polling, finalisation, receipts, Telegram delivery and the tick all pass inside the packaged copy.
- `npm run check` passes inside the packaged copy.
- `python3 scripts/validate_repository.py` passes on this branch.

## Disclosure

The demo video records a real call to the author's own phone. The history that call reads is seeded by `scripts/seed.ts`, and every seeded row is marked `seeded = true` or carries an `orma:seed:` idempotency key.


Ray-56:

1. Restrict credentialed CALLE_API_BASE at the shared boundary to approved HTTPS origins and reject redirects; injected fake transports may remain (apps/web/orma/supabase/functions/_shared/calle.ts:107–110,191–205,458–466).
2. Do not describe uncertain creation/no returned ID as proof no call was placed or no charge occurred, or invite a fresh confirmation attempt. Retain unknown/manual-review wording; no production reconciliation redesign is required (_shared/calle.ts:259–276,565–582; confirm-phone/index.ts:85–89).
3. Mask phone-bearing history/summary/preview display copies while retaining private evidence, and narrow the README halt-all-calls claim to future dispatches rather than accepted-call cancellation (web/src/routes/app/history/model.ts:35; _shared/dispatch-mode.ts:454–471; README:323).

nrynss:
Thanks for the careful review @Ray-56, all three were fair. Addressed in `66b93a6`, rebased on current `main` so the conflict in `apps/README.md` is gone.

1. Valid — `CALLE_API_BASE` was trusted as given. Every credentialed CALL-E request (dispatch, confirmation call, poll, webhook re-fetch) now goes through one boundary, `_shared/calle-origin.ts`. It accepts only an approved HTTPS origin (`https://api.heycall-e.com`), rejecting http, other hosts, userinfo, ports and paths, checked at startup and again on every request, and sends with `redirect: "error"` so the key never follows a redirect. Injected fake transports still work. Tests cover each rejected shape and assert the POST carries `redirect: "error"`.

2. Valid — a missing call id was being read as "no call". Once the POST has gone out, a 5xx or an unreadable response with no id is now recorded as `dispatch_outcome_unknown`, with a reason that says whether CALL-E placed a call is unknown and needs manual review. A 4xx stays a plain refusal. The "no call was placed" log only appears when no request reached CALL-E, and confirm-phone no longer suggests trying again. The stored `not_answered` / `billable = false` pair is kept as the schema's only failed shape, and the comment now says it is not proof of no charge. No reconciliation redesign.

3. Valid — History transcript turns and call summary lines now mask phone numbers in the displayed copy (last two digits kept). The stored rows are untouched. The owner-readable dry-run request body now keeps only the last two digits too. The README operator section now says the steps stop future dispatches and do not cancel a call CALL-E already accepted.

Deno suites, `npm run check` and `python3 scripts/validate_repository.py` all pass on this head.

Fixed in `66b93a6`.

Ray-56:

nrynss:
Thank you @Ray-56 !

PR #709, opened 2026-09-14
docs: add CareForMe to apps list

UdayasreeVadala:
Adds [CareForMe](https://github.com/UdayasreeVadala/CareForMe) to the Apps list in README.md.

CareForMe is a Next.js web app for one-click AI phone check-ins on elderly family members via CALL-E: it holds a natural conversation in the person's language and conversation style, then classifies the outcome as normal / attention / urgent with summaries, alerts, and saved memories. It works with a scripted no-call demo mode by default, validates E.164 numbers, evaluates the CALL-E task with the honest-AI disclosure CALL-E requires, and ships an in-memory store behind a swappable adapter.


Ray-56:
Must Fix under the community/reference review policy:

- The linked CareForMe public default contains an actual provider credential in apps/web/.env.example:2, including its public history. Revoke/rotate it and remove current and affected history copies before merge. Do not repost the value.

PR #712, opened 2026-09-14
feat: add warranty-claimcall (warranty service-desk CALL-E app)

Subramanyarao11:
## Summary
Adds [`apps/python/warranty-claimcall`](apps/python/warranty-claimcall/): a dry-run-first Python CLI that compiles invoice/warranty evidence into one consented, safety-gated CALL-E service-desk call, then verifies the structured case reference against the transcript.

Built for the CALL-E hackathon. Public monorepo: https://github.com/Subramanyarao11/claimcall

## Contribution area
- User-facing / runnable app under `apps/python/`

## Safety
- Fixture dry-run by default (no network)
- Live mode requires explicit confirm hash, CALL-E key, and hashed destination allowlist
- Idempotency reservation before dial; no blind redial on timeout
- `fee.accepted` pinned false in result schema

## Validation
`python3 scripts/validate_repository.py` passes locally.

Made with [Cursor](https://cursor.com)


Ray-56:

1. Make the documented live extra actually provide claimpilot_calle, or label the package offline-only; the current dependencies exclude the adapter imported by claimcall/dispatcher.py:104 (pyproject.toml:24–33).
2. If live remains, enforce authorized ASCII E.164 recipients and approved HTTPS credential origins before adapter construction (dispatcher.py:109; pinned core safety validator).
3. Mask phone-bearing task/issue/result/uncertainty display copies without altering consent hashes or private evidence (claimcall/cli.py:61–96,236–260).
4. Carry --live into the printed live reconciliation command so it reads provider state instead of selecting a fresh fixture provider; retain unresolved/no-ID outcomes for manual review (cli.py:158).

